PR #220, opened 2015-01-08
Handle repeated imports when loading protofiles with a root.

danverbraganza:
see: https://github.com/dcodeIO/ProtoBuf.js/issues/219


danverbraganza:
Btw, is it my responsibility to run npm scripts build, or do you do that as maintainer?

PR #223, opened 2015-01-15
Add streaming methods to parser and reflection API

murgatroid99:
Add the ability to parse streaming requests and responses and expose that information in the reflection API. This is one of the changes needed to support [Protocol Buffers v3](https://github.com/google/protobuf/releases/tag/v3.0.0-alpha-1).


murgatroid99:
I wrote the Node.js library for [grpc](http://grpc.io). I depend on this package from that one, and it would be very useful to me if you merge and publish this change.

PR #243, opened 2015-03-07
rm: index.js; add sources to .npmignore

KidsKilla:


dcodeIO:
I am not convinced that this is how it should be done. Pretty much all packages ship with tests and tooling included.

Issue #238, opened 2015-02-20
Message constructor sometimes crashes when passed a message


murgatroid99:
To replicate:

```js
var ProtoBuf = require("protobufjs");
var builder = ProtoBuf.newBuilder();
builder.define("Test");
builder.create([
  {
    name: "Point",
    fields: [
      {
       rule: "optional",
       type: "int32",
       name: "x",
       id: 1
     },
     {
       rule: "optional",
       type: "int32",
       name: "y",
       id: 2
     }
    ]
  }
]);
var Test = builder.build("Test");
var Point = Test.Point;
var p = {x:0, y:0};
var p_msg = new Point(p);
// this crashes
new Point(p_msg);
```

To me, this is unexpected behavior because `p` appears to be identical to `p_msg` in every relevant way.

vgatto:
I've also noticed this when using gRPC when I attempt to respond with a message that I've deserialized from backend storage. I forked the repo to attempt to fix the constructor, but I don't think it's possible since the constructor is ambiguous.

From Message.js:

```
var Message = function(values, var_args) {
    ProtoBuf.Builder.Message.call(this);
    // Create virtual oneof properties
    for (var i=0, k=oneofs.length; i<k; ++i)
        this[oneofs[i].name] = null;
    // Create fields and set default values
    for (i=0, k=fields.length; i<k; ++i) {
        var field = fields[i];
        this[field.name] = field.repeated ? [] : null;
        if (field.required && field.defaultValue !== null)
            this[field.name] = field.defaultValue;
    }

    if (arguments.length > 0) {
        var value;
        // Set field values from a values object
        if (arguments.length === 1 && typeof values === 'object' &&
            /* not another Message */ typeof values.encode !== 'function' &&
            /* not a repeated field */ !ProtoBuf.Util.isArray(values) &&
            /* not a ByteBuffer */ !(values instanceof ByteBuffer) &&
            /* not an ArrayBuffer */ !(values instanceof ArrayBuffer) &&
            /* not a Long */ !(ProtoBuf.Long && values instanceof ProtoBuf.Long)) {
            this.$set(values);
        } else // Set field values from arguments, in declaration order
            for (i=0, k=arguments.length; i<k; ++i)
                if (typeof (value = arguments[i]) !== 'undefined')
                    this.$set(fields[i].name, value); // May throw
    }
};
```

The constructor has two paths: work with a single object representing all fields, or work with positional arguments.

The offending line for this particular issue is:

```
[...]

PR #248, opened 2015-03-25
Change ProtoBuf.Builder.Message#toRaw to return bytes fields as buffers by default

murgatroid99:
Instead of silently dropping bytes fields, `Message#toRaw` should return them as byte sequence objects. `ByteBuffer#toBuffer` returns different buffer types depending on platform, so this will work on Node and browser. It also either returns a reference to the backing buffer or a slice of it, so this is both fast and memory-efficient.


murgatroid99:
Can you please merge this? This would enable a workaround to #238 in my library that my users are actually experiencing.

dcodeIO:
Merged a couple PRs now and did my best to resolve any conflicts. Could you give it a look and try, before I push this to npm?

murgatroid99:
It looks right to me.

PR #254, opened 2015-04-09
Proto3 support: .proto syntax, map fields, field semantics, JSON.

cfallin:
- Primitive (non-message) fields no longer have a notion of presence. They are always set to some valid value on message objects, and are emitted on the wire when their values are non-default.

- Map fields are first-class citizens: a new `Map` container is added (it follows the ES6 Map API, but I did not want to depend on ES6 or an external polyfill library here), and it serializes and parses in the format defined by proto3 (a repeated field of key/value submessages).

- Enum fields allow any integer value, not just known enum values. This is a more direct approach to forward/backward compatibility than rejecting unknown enum values or punting them to an unknown-field set.

- Proper support for JSON encode/decode with encodeJSON() and decodeJSON() methods on message objects and message classes, respectively. The JSON format is compatible with the standard proto3 format.

Note: the change factors out common code into `src/ProtoBuf/Element.js` from the field reflection object so that maps can use the logic for both keys and values. Hopefully this is not too confusing in the diff.


sunwenteng:
Is anyone trying to merge this pull request? It will be of great help if these features are applied in protobuf.js.

sunwenteng:
pbjs is not supported for .json or .js file from my test.

cfallin:
@sunwenteng , could you describe your problem in some more detail (i.e., what commands to run and what error you see)? What do you mean "pbjs is not supported for .json or .js file"? The `pbjs` utility reads a `.proto` file as input, no? And does this issue appear specifically with this PR merged, or in general?

sunwenteng:
@cfallin, pbjs can convert the `.proto` file into `.json` file or `.js` file. The new feature in #254 such as **'map support'** cannot be transformed to `js/json` file correctly.

i.e., a.proto is below:
```
message test{
    map<string, string> mapData = 1;
}
```

when you run: `./pbjs ../examples/test/a.proto -target=commonjs > ../examples/test/a.js`
you will get `a.js` file below: 
``` js
module.exports = require("protobufjs").newBuilder({})['import']({
    "package": null,
    "messages": [
        {
            "name": "test",
            "fields": [
                {
                    "rule": "optional",
                    "type": "string",
                    "name": "mapData",
                    "id": 1
                }
            ]
        }
    ]
}).build();
```
so, my test is below:
```js
var a = require('./a');
var instance = new a['test']({});
instance.set('a','b');
```
then, error occurred:
```
/home/bingo/node_modules/protobufjs/dist/ProtoBuf.js:1832
                                throw Error(this+"#"+keyOrObj+" is not a field
                                                             ^
Error: .test#a is not a field: undefined
```
In conclusion, obviously, the `a.js` file is not transformed correctly as you can see or is there any mistake I have made?

cfallin:
Ah, the JSON descriptor wasn't describing "map" field types! Actually, `pbjs` was not writing out `"repeated"` for repeated fields, either -- this was a little surprising. In any case, the JSON descriptors are now correct for both map fields and repeated fields.

By the way, in your example, I think you're trying to set a field named `a`, when you probably really meant to set the key `a` in the map field `mapData`, so something like `instance.mapData.set('a', 'b');` ? The actual exception you're getting indicates that the `a` field does not exist, which is not surprising since your `.proto` file does not define an `a` field (you are getting the  `set` method on the message class, not on the map container).

sunwenteng:
yeah, it is my mistake. i was trying to show you my example. I have checked your test in `suit.js`.
Also, `"repeated"` is an error, too.
Thanks a lot.

sunwenteng:
@cfallin, hey, sorry for bothering. I still meet up with the `pbjs` problem. Let's see the example.

i.e., a.proto is below:
```
message test{
    map<string, string> mapData = 1;
}
```

when you run: `./pbjs ../examples/test/a.proto -target=commonjs > ../examples/test/a.js`
you will get `a.js` file below: **(from your pull request, the rule changed)**
``` js
module.exports = require("protobufjs").newBuilder({})['import']({
    "package": null,
    "messages": [
        {
            "name": "test",
            "fields": [
                {
                    "rule": "map",
                    "type": "string",
                    "name": "mapData",
                    "id": 1
                }
            ]
        }
    ]
}).build();
```
but, when i try to require the `a.js`, like below
```js
var a = require('./a');

```
then, error occurred:
```
/home/bingo/WebstormProjects/three/node_modules/protobufjs/dist/ProtoBuf.js:4901
or("Illegal key type for map field in "+this.ptr.toString(true)+": "+this.ptr.
                                                                    ^
Error: Illegal key type for map field in Message.Field .test.mapData: [object Object]
    at Error (native)
    at ProtoBuf.Builder.BuilderPrototype.resolveAll (/home/bingo/WebstormProjects/three/node_modules/protobufjs/dist/ProtoBuf.js:4901:106)
    at ProtoBuf.Builder.BuilderPrototype.resolveAll (/home/bingo/WebstormProjects/three/node_modules/protobufjs/dist/ProtoBuf.js:4872:30)
    at ProtoBuf.Builder.BuilderPrototype.resolveAll (/home/bingo/WebstormProjects/three/node_modules/protobufjs/dist/ProtoBuf.js:4872:30)
    at ProtoBuf.Builder.BuilderPrototype.build (/home/bingo/WebstormProjects/three/node_modules/protobufjs/dist/ProtoBuf.js:4936:26)
    at Object.<anonymous> (/home/bingo/WebstormProjects/three/src/share/db.js:531:4)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
```
The `pbjs` still not works.

cfallin:
Really sorry about that! Now includes one more fix. I've verified with my own test too.

cfallin:
Friendly ping -- @dcodeIO, do you think this is OK to merge?

dcodeIO:
Merged a couple PRs now and did my best to resolve any conflicts. Could you give it a look and try, before I push this to npm?

cfallin:
Looks OK to me, and looks like tests are still passing too. Thanks!

delapavap:
@cfallin Hi, I'm having some problems trying to decode a message that uses a Map field.

```
message SignPartition {

    option (test_message).extends = "AbstractParametrizedAttributeTile";
    repeated SegmentAnchor segment_anchors = 1 [(test_field).pool = "signs" ];
    map<string, AnchorReferences> signs = 2 [(test_field).in_pool = "signs" ];
    repeated SpeedLimitAttribute speed_attrs = 10;
    repeated SignMetadata metadata = 11;  
    
    message AnchorReferences {
       repeated uint32 anchor_references = 1;
       string timestamp = 2;
    }
}
```

the error that I get is the following:

Illegal wire type for field Message.Field SignPartition.AnchorReferences.anchor_references: 2 (0 expected)

any Ideas?

cfallin:
Hi @delapavap, my best guess is that you are trying to decode corrupted data. But if you believe it to be a bug, I'd recommend opening a new issue; I'm not a maintainer of this repo and I don't work on protobufs any more, so I won't be able to investigate. Best of luck!

Issue #263, opened 2015-05-14
TypeError: Object .services.profile.Profile.resume has no method 'getChild'


mhahn:
Parsing the following structure causes an error.

```
$ tree tests
tests
└── services
    ├── profile.proto
    └── resume.proto
$ cat tests/services/resume.proto
package services.resume;

message Resume {
    repeated string experience = 1;
}

$ cat tests/services/profile.proto
package services.profile;

import "services/resume.proto";

message Profile {
    optional services.resume.Resume resume = 1;
}

$ pbjs tests/services/profile.proto -path=tests -target=json

 Processing 'tests/services/profile.proto' ...


/usr/local/lib/node_modules/protobufjs/dist/ProtoBuf.js:1546
                        child = ptr.getChild(part[i]);
                                    ^
TypeError: Object .services.profile.Profile.resume has no method 'getChild'
    at ProtoBuf.Reflect.NamespacePrototype.resolve (/usr/local/lib/node_modules/protobufjs/dist/ProtoBuf.js:1546:37)
    at ProtoBuf.Reflect.NamespacePrototype.qn (/usr/local/lib/node_modules/protobufjs/dist/ProtoBuf.js:1577:36)
    at buildMessageField (/usr/local/lib/node_modules/protobufjs/cli/pbjs/targets/json.js:180:51)
    at /usr/local/lib/node_modules/protobufjs/cli/pbjs/targets/json.js:159:21
    at Array.forEach (native)
    at buildMessage (/usr/local/lib/node_modules/protobufjs/cli/pbjs/targets/json.js:156:53)
    at /usr/local/lib/node_modules/protobufjs/cli/pbjs/targets/json.js:74:23
    at Array.forEach (native)
    at buildNamespace (/usr/local/lib/node_modules/protobufjs/cli/pbjs/targets/json.js:73:46)
    at /usr/local/lib/node_modules/protobufjs/cli/pbjs/targets/json.js:89:9
```
This has something to do with the field name being called "resume" and the package being "services.resume".

If I change the field name or the package name, it compiles correctly.

Issue #262, opened 2015-05-14
throw Error("Duplicate name in namespace "+this.toString(true)+": "+child.nam


jeckbjy:
files : "common.proto", "test1.proto", "test2.proto", "all.proto"
"test1.proto" use "common.proto"
"test2.proto" use "common.proto"
"all.proto" use "test1.proto" and "test2.proto"

it will throw error:
Error: Duplicate name in namespace Namespace : Person
    at Error (native)
    at ProtoBuf.Reflect.NamespacePrototype.addChild (C:\Users\duan\AppData\Roaming\npm\node_modules\
protobufjs\dist\ProtoBuf.js:1508:92)
    at ProtoBuf.Builder.BuilderPrototype.create (C:\Users\duan\AppData\Roaming\npm\node_modules\prot
obufjs\dist\ProtoBuf.js:3966:42)
    at ProtoBuf.Builder.BuilderPrototype.import (C:\Users\duan\AppData\Roaming\npm\node_modules\prot
obufjs\dist\ProtoBuf.js:4119:26)
    at ProtoBuf.Builder.BuilderPrototype.import (C:\Users\duan\AppData\Roaming\npm\node_modules\prot
obufjs\dist\ProtoBuf.js:4104:47)
    at ProtoBuf.Builder.BuilderPrototype.import (C:\Users\duan\AppData\Roaming\npm\node_modules\prot
obufjs\dist\ProtoBuf.js:4104:47)
    at Object.ProtoBuf.loadJson (C:\Users\duan\AppData\Roaming\npm\node_modules\protobufjs\dist\Prot
oBuf.js:4394:30)
    at Object.module.exports [as proto] (C:\Users\duan\AppData\Roaming\npm\node_modules\protobufjs\c
li\pbjs\sources\proto.js:34:14)
    at Object.pbjs.main (C:\Users\duan\AppData\Roaming\npm\node_modules\protobufjs\cli\pbjs.js:194:4
7)
    at Object.<anonymous> (C:\Users\duan\AppData\Roaming\npm\node_modules\protobufjs\bin\pbjs:2:51)

mhahn:
I'm running into the same issue above.

If the proto has already been imported into the Namespace, why can it not just be ignored?

mhahn:
this seems to be the same as: https://github.com/dcodeIO/ProtoBuf.js/commit/4269097114c43a5b946ed8dcc2c16240a1b531be

dcodeIO:
[...]

PR #265, opened 2015-05-16
Test cases and fixes for #262 and #263

mhahn:


dcodeIO:
Hmm, removing the errors thrown there doesn't quite fix this. The exceptions are there for a reason, that is, there shouldn't be duplicate namespace declarations at all. Removing the exceptions will result in a situation where you'll never know what exactly is in that namespace - it could be completely different structures.

A proper fix would require taking a look at why exactly a namespace is declared multiple times, and fixing the root cause instead.

mhahn:
Agreed. I was trying to do this in a hurry to get something working.

After taking a look at it, the root cause is this:

https://github.com/dcodeIO/ProtoBuf.js/blob/master/src/ProtoBuf/Builder.js#L465-L468

simply appending "importN" to the filename means that the same import in different files will have a different "filename" so the checks you have to protect against duplicate imports don't work.

I'm not sure what the best way to do this in javascript is, but doing something like the following ensures that we're not importing duplicate structures:

in src/ProtoBuf/Util.js:
```
    Util.hashCode = function(value) {
        var hash = 0, i, chr, len;
        if (value.length == 0) return hash;
        for (i = 0, len = value.length; i < len; i++) {
            chr   = value.charCodeAt(i);
            hash  = ((hash << 5) - hash) + chr;
            hash |= 0; // Convert to 32bit integer
        }
        return hash;
    };
```

in src/ProtoBuf/Builder.js at line 463:
```
    if (!filename) {
        this["import"](json['imports'][i]);
    } else {
        var hash = ProtoBuf.Util.hashCode(JSON.stringify(json['imports'][i]));
        this["import"](json['imports'][i], filename.split('::')[0] + '::' + hash);
    }

```

the filename itself isn't as clean as how you had it since it can contain a negative integer, but I don't see that you're using this other than in the duplicate import checks where it doesn't matter.

Let me know if you think that is in the right direction and I'll update the PR.

dcodeIO:
Indeed, renaming stuff like this doesn't seem to be sane. I am currently wondering why exactly this was even required, as duplicate imports are already skipped earlier in the code.

I assume this is somehow connected to the old proto2js command line utility, and might not be required anymore. The pbjs tool combines everything into a common namespace, effectively removing all imports from the generated outputs.

Have you already tried to replace that portion with just

```js
                } else // Import structure
                    this["import"](json['imports'][i], filename);
```

?

mhahn:
That doesn't work. It will skip parsing the imports because of these lines:
https://github.com/dcodeIO/ProtoBuf.js/blob/master/src/ProtoBuf/Builder.js#L403-L405

The JSON structure its working with looks like this:

```
{ package: 'main',
  messages:
   [ { name: 'MainMessage',
       fields:
        [ { rule: 'optional',
            options: {},
            type: 'main.Dep1',
            name: 'test1',
            id: 1 },
          { rule: 'optional',
            options: {},
            type: 'main.Dep2',
            name: 'test2',
            id: 2 },
          { rule: 'optional',
            options: {},
            type: 'main.Common',
            name: 'test3',
            id: 3 } ],
       enums: [],
       messages: [],
       options: {},
       oneofs: {} } ],
  enums: [],
  imports:
   [ { package: 'main',
       messages:
        [ { name: 'Dep1',
            fields: [Object],
            enums: [],
            messages: [],
            options: {},
            oneofs: {} } ],
       enums: [],
       imports:
        [ { package: 'main',
            messages: [Object],
            enums: [],
            imports: [],
            options: {},
            services: [] } ],
       options: {},
       services: [] },
     { package: 'main',
       messages:
        [ { name: 'Dep2',
            fields: [Object],
            enums: [],
            messages: [],
            options: {},
            oneofs: {} } ],
       enums: [],
       imports:
        [ { package: 'main',
            messages: [Object],
            enums: [],
            imports: [Object],
            options: {},
            services: [] } ],
       options: {},
       services: [] },
     { package: 'main',
       messages:
        [ { name: 'Common',
            fields: [Object],
            enums: [],
            messages: [],
            options: {},
            oneofs: {} } ],
       enums: [],
       imports: [],
       options: {},
       services: [] } ],
  options: {},
  services: [] }
```

The unique filenames allow it to parse through the nested imports as it stands today. What I don't really understand is why the "package" is maintained for the imports, but not the filename. 

If the imports maintained their actual filename, I think everything would work properly.

dcodeIO:
How exactly did you obtain that JSON definition? pbjs should actually collapse everything into one namespace to prevent these kind of issues.

And yes, without file names, there is no way to know whether an import is a duplicate or not - and it shouldn't generate something like that.

mhahn:
It's a result of the "dupimport" test case I submitted in the PR

dcodeIO:
So, this has been generated by pbjs? If so, is this still reproducible with the latest changes?

mhahn:
The test case passes the file name to pbjs. I'll check if it's still
reproducible.

mhahn:
cool, looks like this is fixed with the recent changes. this PR just contains the tests now.

Thanks!

PR #284, opened 2015-06-08
Convert cli to use yargs

### LINKED ISSUE #277: pbjs interface is non-standard
Pbjs is acting differently than what most people would expect a cli app to be like. Here are a few specific examples of what I mean:

Doesn't implement the standard help flags. Running `pbjs -h` or `pbjs --help` both generate an error instead of printing help output. I had to look at the source to figure out how to use the cli.

Implements non-standard flag format. Instead of using the normal short/long format (`-h`, and `--help`) everything is a single dash long name, which is unlike most other apps.

Would you be open to using something [`yargs`](https://www.npmjs.com/package/yargs) or [`commander`](https://www.npmjs.com/package/commander) to help normalize some of these unintuitive things and provide a bit more robust option parsing?

englercj:
Fixes #277.

The help output isn't quite *as* pretty as it was before, but it is pretty close:

```
$ ./bin/pbjs --help

 _ |_ . _
|_)|_)|_)           ProtoBuf.js v4.0.0-b3 https://github.com/dcodeIO/ProtoBuf.js
|     '

CLI utility to convert between .proto and JSON syntax / to generate classes.
Usage: pbjs <filename> [options] [> outFile]


Options:
  --help            Show help  [boolean]
  --version         Show version number  [boolean]
  --source, -s      Specifies the source format. Valid formats are:
                        json         Plain JSON descriptor
                        proto        Plain .proto descriptor
  --target, -t      Specifies the target format. Valid formats are:
                        amd          Runtime structures as AMD module
                        commonjs     Runtime structures as CommonJS module
                        js           Runtime structures
                        json         Plain JSON descriptor
                        proto        Plain .proto descriptor  [default: "json"]
  --using, -u       Specifies an option to apply to the volatile builder
                    loading the source, e.g. convertFieldsToCamelCase.
  --min, -m         Minifies the output.  [default: false]
  --path, -p        Adds a directory to the include path.
  --legacy, -l      Includes legacy descriptors from google/protobuf/ if
                    explicitly referenced.  [default: false]
  --quiet, -q       Suppresses any informatory output to stderr.  [default: false]
  --use, -i         Specifies an option to apply to the emitted builder
                    utilized by your program, e.g. populateAccessors.
  --exports, -e     Specifies the namespace to export. Defaults to export
                    the root namespace.
  --dependency, -d  Library dependency to use when generating classes.
                    Defaults to 'protobufjs' for CommonJS, 'ProtoBuf' for
                    AMD modules and 'dcodeIO.ProtoBuf' for classes.
```

(it is still colored)

### DIFF
--- a/cli/pbjs.js
+++ b/cli/pbjs.js
@@ -17,6 +17,7 @@ var ProtoBuf = require(__dirname+"/../dist/ProtoBuf.js"),
     fs       = require("fs"),
     path     = require("path"),
     cli      = require("ascli")("pbjs"),
+    yargs    = require("yargs"),
     util     = require("./pbjs/util.js"),
     pkg      = require("../package.json");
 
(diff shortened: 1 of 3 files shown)


dcodeIO:
Thanks!

PR #314, opened 2015-08-22
Allow calling service methods with plain objects

### LINKED ISSUE #312: Allow calling service methods with plain objects
Currently, the following is an error:

```js
someService.SomeMethod({ arg1: 'val1' }, callback);
```

ProtoBuf.js knows which type is expected, so it should be able to construct an instance of the appropriate message automatically.

seishun:
This fixes #312.


seishun:
Great, looking forward to the npm release with this.

PR #325, opened 2015-09-24
fix #225 ignore fs for browser builds

### LINKED ISSUE #225: Build a browser specific version instead of using require fs/path
Would it be possible to build a separate browser specific version instead of having this kind of code in the output?
```js
Util.IS_NODE = false;
try {
    // There is no reliable way to detect node.js as an environment, so our
    // best bet is to feature-detect what we actually need.
    Util.IS_NODE =
        typeof require === 'function' &&
        typeof require("fs").readFileSync === 'function' &&
        typeof require("path").resolve === 'function';
} catch (e) {}
```

I'm saying this because browserify will include the fs and path module in the output when we require ProtoBuf, but these modules are not needed for ProtoBuf to work.
You could dynamically inject a value for `Util.IS_NODE` during building/bundling.

What do you think?

watsoncj:
The browser builds (webpack, browserify) shouldn't attempt to require and inline the unreachable `require('fs')`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,9 @@
   "engines": {
     "node": ">=0.8"
   },
+  "browser": {
+   "fs": false
+  },
   "scripts": {
     "prepublish": "npm test",
     "test": "node bin/pbjs tests/complex.proto --target=json > tests/complex.json && node node_modules/testjs/bin/testjs tests/suite.js",


dcodeIO:
Seems to be exactly what is needed, thanks!

PR #341, opened 2015-10-21
Parse nested services

kjvalencik:
Resolves https://github.com/dcodeIO/ProtoBuf.js/issues/285#issuecomment-149718823


kjvalencik:
The failed test is because I did not check in compiled files. Should I?

PR #371, opened 2015-12-03
avoid so many fn.bind calls

jfromaniello:
Some evidence that we can improve a bit the performance by not calling bind so many times:


RenEvo:
+1 on this, anything to squeeze a bit more perf out of this library will make my life happy.

englercj:
:+1:  Bind is extremely expensive, and closure scoping is extremely cheap. This is a great addition.

PR #378, opened 2015-12-17
Performance optimizations

kjvalencik:
By far the largest optimization is removing the function / bind for `fail`. Benchmark results:

```
Old Protobuf.js: count:  978, mean: 60.215909270156000 µs
New Protobuf.js: count: 3364, mean: 15.561118733463415 µs
```

### DIFF
--- a/dist/protobuf-light.js
+++ b/dist/protobuf-light.js
@@ -865,7 +865,6 @@
          * @param {boolean=} unsigned Whether unsigned or not, defaults to reuse it from Long-like objects or to signed for
          *  strings and numbers
          * @returns {!Long}
-         * @throws {Error} If the value cannot be converted to a Long
          * @inner
          */
         function mkLong(value, unsigned) {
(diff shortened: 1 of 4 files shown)


dcodeIO:
Thank you for the PR. I did fix it by just removing the call to .bind. If this yields bad performance still, let me know.

kjvalencik:
I suspect that since this is such a hot loop, the closure will still be a significant cost. I'll re-run my tests and see if this is sufficient. Thanks!

kjvalencik:
I re-ran my tests with the three versions. Bind is the old version before your last commit, closure is your last commit, and prototype is `fail` moved to the prototype instead.

```text
Bind:      count: 1008, mean: 58.055880591439560 µs
Closure:   count: 2655, mean: 19.477010936329330 µs
Prototype: count: 3334, mean: 15.614463297582034 µs
```

Switching to a closure from bind improved things quite a bit, but there is still another 20%+ improvement available by removing the closure.

Issue #373, opened 2015-12-07
Accessors usage in your implementation


Braingines:
It seems that starting from this topic

https://github.com/dcodeIO/ProtoBuf.js/issues/179

Nothing is changed with accessors, right? 

The AoJ and dougrad users clearly said that current accessors implementation is just wrong and violates the ProtoBuf specification in case of setting default values for primitive fields and representing empty optional or repeated primitive fields as their defaults.

So the only motivation to use your current implementation is in not transfering the data scheme with every message, that's it. 

And, at the same time, the main reason why the ProtoBuf was just created (that is the proper way of accessors usage that leads to reducing dramatically the whole data parsing time) is still ignored.

Do you plan to fix this "JS feature" somehow? Do you have any proposals how to fix it or still not? If not, I want to propose you to implement "has" accessor and to fix other accessors in the way they implemented in C++ / Java.

[UPD]:

https://github.com/dcodeIO/protobuf.js/issues/200#issuecomment-98153230

This topic seems to me as a solution to decode message and get non-empty optional and repeated fields

dcodeIO:
Closing this for now.

[protobuf.js 6.0.0](https://github.com/dcodeIO/protobuf.js/releases/tag/6.0.0)

Feel free to send a pull request if this is still a requirement.

PR #380, opened 2015-12-31
Move default values to prototypes and ditch nulls

### LINKED ISSUE #319: proto2 default optional fields construction behaviour
Hi, 

Reading the [official language guide](https://developers.google.com/protocol-buffers/docs/proto#optional), I came across the following clause:

> If the default value is not specified for an optional element, a type-specific default value is used instead: for strings, the default value is the empty string. For bools, the default value is false. For numeric types, the default value is zero. For enums, the default value is the first value listed in the enum's type definition.

When using ProtoBuf.js, I construct via builder a message (`new MyMessage ()`). All optional fields are `null`, but I guess they should be set to its type's default value, as stated above i.e. integers as zero, booleans as false and strings as empty strings. Even after decoding, if optional fields were not in the wire data they will remain `null` instead of having their protocol default values. 

Does this means ProtoBuf.js construction is not compliant with the protocol?

**I've edited the description to better state the issue**

seishun:
This fixes https://github.com/dcodeIO/protobuf.js/issues/200 by implementing the API described in https://github.com/dcodeIO/protobuf.js/issues/200#issuecomment-98153230.
- Fields without an explicit default are treated as if the default were a type-specific value, as per the spec and the official implementations. For example, `optional bool foo = 0;` is now equivalent to `optional bool foo = 0 [default = false];`.
- For every non-repeated non-map field, there is an equivalently named property on the message prototype. Its value is the field's default value.
- When constructing a message object, its properties are no longer set for non-repeated non-map values. Consequently, attempting to read any of its fields results in the default value.
- When encoding, only own properties are taken into account, unless it's a required field (previously, required fields with explicit defaults were set on construction, so one could encode without explicitly setting them).
- When decoding, fields are no longer set to their defaults if they were missing on the wire, since the default values are already in the prototype. If a required field is missing on the wire, decoding throws even if the field has an explicit default.
- The `ProtoBuf.populateDefaults` option is removed.

The changes are breaking for some (unlikely) use cases, but I tried to be as conservative as possible. Also note that all the above only applies to `proto2`. The behavior of `proto3` should be exactly the same as before.

Also fixes #319, #373.

### DIFF
--- a/src/ProtoBuf/Builder/Message.js
+++ b/src/ProtoBuf/Builder/Message.js
@@ -23,10 +23,11 @@ var Message = function(values, var_args) {
     // Create fields and set default values
     for (i=0, k=fields.length; i<k; ++i) {
         var field = fields[i];
-        this[field.name] =
-            field.repeated ? [] :
-            (field.map ? new ProtoBuf.Map(field) : null);
-        if ((field.required || T.syntax === 'proto3') &&
+        if (field.repeated)
+            this[field.name] = [];
+        else if (field.map)
+            this[field.name] = new ProtoBuf.Map(field);
+        else if (T.syntax === 'proto3' &&
             field.defaultValue !== null)
             this[field.name] = field.defaultValue;
     }
@@ -128,7 +129,7 @@ MessagePrototype.set = function(keyOrObj, value, noAssert) {
         var currentField = this[field.oneof.name]; // Virtual field references currently set field
         if (value !== null) {
             if (currentField !== null && currentField !== field.name)
-                this[currentField] = null; // Clear currently set field
+                delete this[currentField]; // Clear currently set field
             this[field.oneof.name] = field.name; // Point virtual field at this field
         } else if (/* value === null && */currentField === keyOrObj)
             this[field.oneof.name] = null; // Clear virtual field (current field explicitly cleared)
--- a/src/ProtoBuf/Reflect/Message.js
+++ b/src/ProtoBuf/Reflect/Message.js
@@ -109,6 +109,13 @@ MessagePrototype.build = function(rebuild) {
             throw Error("Illegal reflect child of "+this.toString(true)+": "+this.children[i].toString(true));
     }
 
+    // Default values
+    for (i=0, k=this._fields.length; i<k; i++) {
+        var field = this._fields[i];
+        if (!field.repeated && !field.map && T.syntax !== 'proto3')
+            clazz.prototype[field.name] = field.defaultValue;
+    }
+
     return this.clazz = clazz;
 };
 
@@ -130,7 +137,7 @@ MessagePrototype.encode = function(message, buffer, noVerify) {
         if (field.required && val === null) {
             if (fieldMissing === null)
                 fieldMissing = field;
-        } else
+        } else if (field.required || message.hasOwnProperty(field.name))
             field.encode(noVerify ? val : field.verifyValue(val), buffer, message);
     }
     if (fieldMissing !== null) {
@@ -258,7 +265,7 @@ MessagePrototype.decode = function(buffer, length, expectedGroupEndId) {
             if (field.oneof) { // Field is part of an OneOf (not a virtual OneOf field)
                 var currentField = msg[field.oneof.name]; // Virtual field references currently set field
                 if (currentField !== null && currentField !== field.name)
-                    msg[currentField] = null; // Clear currently set field
+                    delete msg[currentField]; // Clear currently set field
                 msg[field.oneof.name] = field.name; // Point virtual field at this field
             }
         }
@@ -267,15 +274,14 @@ MessagePrototype.decode = function(buffer, length, expectedGroupEndId) {
     // Check if all required fields are present and set default values for optional fields that are not
     for (var i=0, k=this._fields.length; i<k; ++i) {
         field = this._fields[i];
-        if (msg[field.name] === null) {
+        if (!msg.hasOwnProperty(field.name)) {
             if (this.syntax === "proto3") { // Proto3 sets default values by specification
                 msg[field.name] = field.defaultValue;
             } else if (field.required) {
                 var err = Error("Missing at least one required field for " + this.toString(true) + ": " + field.name);
                 err["decoded"] = msg; // Still expose what we got
                 throw(err);
-            } else if (ProtoBuf.populateDefaults && field.defaultValue !== null)
-                msg[field.name] = field.defaultValue;
+            }
         }
     }
     return msg;
--- a/src/ProtoBuf/Reflect/Message/Field.js
+++ b/src/ProtoBuf/Reflect/Message/Field.js
@@ -145,14 +145,10 @@ FieldPrototype.build = function() {
     if (this.map)
         this.keyElement = new Element(this.keyType, undefined, true, this.syntax);
 
-    // In proto3, fields do not have field presence, and every field is set to
-    // its type's default value ("", 0, 0.0, or false).
-    if (this.syntax === 'proto3' && !this.repeated && !this.map)
-        this.defaultValue = Element.defaultFieldValue(this.type);
-
-    // Otherwise, default values are present when explicitly specified
-    else if (typeof this.options['default'] !== 'undefined')
+    if (typeof this.options['default'] !== 'undefined')
         this.defaultValue = this.verifyValue(this.options['default']);
+    else if (!this.repeated && !this.map)
+        this.defaultValue = Element.defaultFieldValue(this.type);
 };
 
 /**
--- a/src/protobuf.js
+++ b/src/protobuf.js
@@ -248,14 +248,6 @@ ProtoBuf.convertFieldsToCamelCase = false;
  */
 ProtoBuf.populateAccessors = true;
 
-/**
- * By default, messages are populated with default values if a field is not present on the wire. To disable
- *  this behavior, set this setting to `false`.
- * @type {boolean}
- * @expose
- */
-ProtoBuf.populateDefaults = true;
-
 //? include("ProtoBuf/Util.js");
 
 //? include("ProtoBuf/Lang.js");
--- a/tests/suite.js
+++ b/tests/suite.js
@@ -432,7 +432,7 @@
                 var builder = ProtoBuf.loadProto("message Test { optional bool ok = 1 [ default = false ]; }"),
                     Test = builder.build("Test"),
                     t =  new Test();
-                test.strictEqual(t.ok, null); // Not set as it is optional
+                test.strictEqual(t.ok, false); // = default when missing
                 t.setOk(true);
                 test.strictEqual(t.ok, true);
                 test.strictEqual(Test.decode(t.encode()).ok, true);
@@ -561,6 +561,7 @@
                 var Test = builder.build("Test");
                 var t = new Test(), bb = new ByteBuffer(2);
                 t.setA(1);
+                t.b = null;
                 try {
                     bb = t.encode(bb).flip();
                     test.ok(false);
@@ -577,7 +578,7 @@
                     // But still be able to access the rest
                     var t3 = e.decoded;
                     test.strictEqual(t3.a, 1);
-                    test.strictEqual(t3.b, null);
+                    test.strictEqual(t3.b, 0);
                 }
                 test.strictEqual(t2, undefined);
             } catch (e) {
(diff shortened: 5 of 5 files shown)


englercj:
I'm not sure I like this change because it will break Hidden Class optimizations since the object will change itself at runtime. This will also destroy the Inline Caches for methods that take in these objects as parameters. The object should not change the properties it has at runtime, at construction it should have all the properties it would ever have.

seishun:
@englercj Perhaps you can suggest a test script that we could use to measure the performance difference? Unless it's significant, I would argue that compliance to the spec and API simplicity is more important.

englercj:
@seishun The performance difference is heavily different because one way will generate optimized code, the other will be full compiled (unoptimized code). Depending on your application this can be unimportant or application breaking. Measuring the performance difference in raw numbers doesn't mean much except in the context of your application, but breaking well known engine optimizations and forcing these objects to unoptimizable (and therefore break all functions that use them as well) is not a good idea.

seishun:
>Measuring the performance difference in raw numbers of some test script doesn't mean much except in the context of your application

It means more than speculation.

>I would move to a different library for our applications if the parsing of messages caused it to return a different hidden class every call.

Surely you wouldn't move without seeing the numbers first? In that case, it would be beneficial for everyone involved if you could test this solution now.

>I'm suggesting that changing the properties of the object by adding or removing them at runtime is not the correct solution.

I'll be glad to see other suggestions if I see that this solution is unacceptable performance-wise.

dcodeIO:
I actually like the hasOwnProperty idea a lot, great replacement for has_ API-wise. Regarding hidden class optimizations: The issue here is `delete`, right?

englercj:
> It means more than speculation.

I'm not speculating, I'm trying to explain to you how V8 works. I even included a benchmark and an article that explains the compilation process I am describing.

> Google seems to disagree with your statement that "at construction [the object] should have all the properties it would ever have":

Huh? The prototype is part of the object's hidden class; by placing value properties on the prototype, you essentially let V8 do the initialization of them for you on `new` rather than having them explicitly in the constructor. That article in no way disagrees with what I am saying. That article is suggesting that instead of running js to initialize the vars, let v8 do it for you.

The optimization they are referring to is to reduce construction time, but has nothing to do with add/removing properties at runtime. Hidden classes are per-object (which includes `__proto__`). Where they are doesn't matter, not changing them is what matters.

-------

> The issue here is delete, right?

Pretty much, the properties that the message will hold need to be added at construction time then not deleted. Each delete causes a hidden class transition which is expensive, and having the multitude of hidden classes (depending on the diversity of how optional a messages arguments are) can cause megamorphic use of it. This can be unacceptable for many use cases, including ours.

seishun:
>Where they are doesn't matter, not changing them is what matters.

In that case, I don't understand your gripe with this solution. There is always going to be a property on the object for each field, either own or inherited.

englercj:
> In that case, I don't understand your gripe with this solution. There is always going to be a property on the object for each field, either own or inherited.

https://github.com/dcodeIO/protobuf.js/pull/380/files#diff-3046de64cab49373b52464f636b51688R132

No, there isn't. You explicitly have delete, which removes the property. That transitions the object to a new Hidden Class.

seishun:
>No, there isn't. You explicitly have delete, which removes the property.

Removing a property doesn't remove it from the prototype.

```js
> var o = Object.create({foo: 5})
undefined
> o.foo = 'bar'
'bar'
> delete o.foo
true
> o.foo
5
```

englercj:
> Removing a property doesn't remove it from the prototype.

Never said it did. I said removing the property from the object changes the Hidden Class. I said that the object `o` and the prototype `o.__proto__` have separate Hidden Classes. Doing `delete o` changes the Hidden Class used to represent `o` in the V8 engine.

seishun:
>I said removing the property from the object changes the Hidden Class. I said that the object o and the prototype o.__proto__ have separate Hidden Classes. Doing delete o changes the Hidden Class used to represent o in the V8 engine.

Now you're contradicting yourself. So does it or does it not matter where the property is?

Assuming your current stance is that it does:

>By placing value properties on the prototype, you essentially let V8 do the initialization of those properties for you on new rather than having them explicitly in the constructor. That article in no way disagrees with what I am saying. That article is suggesting that instead of running js to initialize the vars, let v8 do it for you.

Then yes, it does disagree with you. If a property is on the prototype, then it doesn't magically become the object's own property after construction – it remains inherited. The two snippets are semantically different.

dcodeIO:
Well, you have two hidden classes here: The single prototype that contains the default values as its "own" properties (the generated `ProtoBuf.Builder.Message`) and multiple instances of it, the actual runtime messages. The latter will fall back to "hashmap mode" in the worst case as soon as you call "delete" on any of their properties, hence disabling hidden class optimization for the exact instance it was used with.

The prototype (generated class) is not the issue here, the issue are the instances of them (those where delete is used on properties). It's not about changing the prototype, it's about changing objects that share their exact property layout (which is optimized through a hidden class in v8).

seishun:
@englercj Not very classy to edit your comments to make them look more convincing after they've been responded to.

>Being on the prototype is fine, being on the object is fine, changing either after construction is a no-no.

Are you implying that the snippet from Google assumes the properties `prop1_`, `prop2_` and `prop4_` will never be modified after construction?

>I recommend you read the articles I posted, because I don't think you understand what I am saying.

Have you read them? Because it seems you have skipped over or are intentionally ignoring the second-to-last paragraph in the first article you linked:

>Indeed worrying about polymorphism is usually futile. Instead benchmark your code on a realistic dataset, profile it for hotspots and if any of them are JS related - check out IR that optimizing compiler produces.

dcodeIO:
Forgot that this is a dick-measuring contest, my bad. Ignore my comment.

Braingines:
I feel sorry but I should join your 'dick-measuring contest' because that was my initiative to push this feature implementation

Dear all ProtoBuf.js maintainers, let's read this short article from ProtoBuf (and Cap'n'Proto) dev team:

https://developers.google.com/protocol-buffers/docs/overview#a-bit-of-history

<<Protocol buffers were initially developed at Google to deal with an index server request/response protocol. Prior to protocol buffers, there was a format for requests and responses that used hand marshalling/unmarshalling of requests and responses, and that supported a number of versions of the protocol. This resulted in some very ugly code, like:

 if (version == 3) {
   ...
 } else if (version > 4) {
   if (version == 5) {
     ...
   }
   ...
 }

>>

...

<<
Protocol buffers were designed to solve many of these problems:

New fields could be easily introduced, and intermediate servers that didn't need to inspect the data could simply parse it and pass through the data without needing to know about all the fields.>>

And what you just did? 

You implemented something like tricky V8 - based binary stream reader and writer that work excluding data scheme. 

THAT'S IT.

The whole feature set mentioned above were completely IGNORED the whole your implementation story, so please do not even try to tell that this is ProtoBuf implementation. 

This is ProtoBuf API - compatible V8 - based reader and writer implementation. That's all.

You have no possibility in code to get list of fields that were changed! 

What's the point to use your implementation? 

Your current version just save the traffic bytes, ok, well done guys, well done! 

I'm just wondering and can't imagine your .js code, how do you work with rich protocols (many message with many fields)? 

Looks like you do tens of if / else or switch / case statements in your code. In our case this is not tens, but hundreds of this useless code jumps because of your 'implementation'.

Now, If you have some little doubts about how your super - duper - faster - hidden V8 - based code works, just ask the ProtoBuf initial developer what does he think about implementation without empty reflection feature set and your '.js null - features'.

englercj:
@seishun I edit my posts when I have a typo, realize I said something wrong, or feel that I need to clarify. I'm not trying to "win" because this isn't an argument, I'm trying to educate you.

I think you are interpreting "Hidden Class" as the prototype, but that isn't what that is. Hidden Class !== Prototype.

> Are you implying that the snippet from Google assumes the properties prop1_, prop2_ and prop4_ will never be modified after construction?

Changing the value isn't the issue (unless you are changing the type), adding and removing properties that change the structure of the object is the issue. I've mentioned multiple times now that I have both profiled, and read the IR of our application. And I'm here to tell you we cannot suffer an deoptimization of this scale. For most applications or websites this probably doesn't matter. For hight-performance noded.js application processing millions of messages per second (like ours) it matters, a lot.

> Then yes, it does disagree with you. If a property is on the prototype, then it doesn't magically become the object's own property after construction – it remains inherited. The two snippets are semantically different.

They are different, but neither of them change the structure of the object after construction. The optimization mentioned there is a different one than what we have been discussing. `__proto__` is a reference to the prototype object which was constructed on `new`. The object's own properties are separate and set in the constructor function. Adding/removing properties to either object after they are constructed will result in a Hidden Class transition for that object that changed.

@dcodeIO I'm done now, I don't understand why this is spiraling out of control. I'm trying to be reasonable and explain how this works, but I just keep getting met with "nuh-uh" and ad hominem attack. Do what you feel is best, you have been honoring rules like these while building this lib for a bit now so I trust you.

In fact, its good to see that you actually do understand engine mechanics:

> Well, you have (at least) two (more with inheritance taken into account) hidden classes here: One for the single prototype that contains the default values as its "own" properties (the one-time generated ProtoBuf.Builder.Message) and (exactly, at least that is what we want to achieve) one for the multiple instances of it, the actual runtime messages. Runtime instances will fall back to "slow mode" in the worst case as soon as you call "delete" on any of their properties, hence disabling hidden class optimization for the exact instance it was used with. As far as I understand this is also the case when adding other new properties to them. Currently, we set everything to null to retain the layout.

Thank you for understanding, and we look forward to working on future library optimizations.

Braingines:
Does anybody of repository users can provide us the complete example of how they use optional / repeated fields on .js - side? 

I mean, what's the point to discuss the performance of wrong implementation code?

I completely do not understand the reasons why people use this. 

Just for compatibility with non - .js server side data exchange components? 

To reduce traffic only? 

If so, you can write your own strict binary reader / writer that exclude the data scheme in traffic, do not marshalling each time parse the message that's it. 

But it will be not the ProtoBuf implementation, sorry, guys.

The ProtoBuf technology means the complete feature set and here we have a lot of features missed in reflection infrastructure. 

The initial proposal and implementation mentioned in the top of this discussion is just the one attempt to implement this feature set in order to provide complete ProtoBuf technology implementation. 

This attempt faced with a lot of 'it works perfectly now _for us_, why we should use your code' (sick).

One more question, does anybody of repository maintainers used other ProtoBuf implementations? C++ / Java - based? If someone can positively answer this question, I see no more pros for push this code further, in production...

seishun:
>I'm trying to educate you

I understand how hidden classes work, thank you very much. You haven't told me anything new. You're missing my point that speculating about performance based on one known optimization is unreasonable, as the very article you linked clearly points out.

>Changing the value isn't the issue (unless you are changing the type), adding and removing properties that change the structure of the object is the issue.

Assigning to either `prop1_`, `prop2_` or `prop4_` of a newly-constructed instance of `Bar` will add a new own property. According to your (likely simplistic) model of how V8 works, this will cause a hidden class transition. That's a bad thing, right? I wonder why Google would suggest such a thing on a page labeled "Optimizing JavaScript code"?

mraleph:
@seishun the page you are linking to is really-really old, I think it even predates V8 itself

Please don't use it to justify any sort of optimizations.

Advice given in the section _"Initializing instance variables"_ might have been a very clear cut back in the day, but these days it comes with a very complicated trade-off - and usually it's impossible to predict whether it will improve the performance and memory consumption of your code. The opposite is also true - it's impossible to predict whether it will regress performance either. Requires benchmarking and detailed analysis. 

It is true however that if the field is almost never present then you might at least save some memory (potentially sacrificing some performance due to polymorphism). 

[Note: I did not look at the code in question, I am just commenting on the advice from _"Initializing instance variables"_ in general]

seishun:
>Requires benchmarking and detailed analysis. 

That's exactly what I've been saying the whole time.

>Please don't use it to justify any sort of optimizations.

I'm only using it to show that blindly following optimization advice can lead to contradiction.

seishun:
@dcodeIO It's been a month. Since the only person who was against this change due to performance concerns still hasn't provided benchmarks supporting their claims, can we merge this or at least move on to discussing other aspects besides performance?

michelgb:
Hello. I really need this code. I am using proto3 and I am having several issues with the defaults. Could someone please review it and merge?

r-tock:
+1

bootstraponline:
I don't think performance matters if we're not spec compliant. I'd like this PR to be merged.

Alternatively update the readme so people that need proto3 support are aware of the issue.

paralin:
@bootstraponline agreed, any progress on this?

bootstraponline:
> @bootstraponline agreed, any progress on this?

I think we're just waiting for someone with write access to press the merge button.

seishun:
Could you provide some explanation for why this was closed?

bootstraponline:
https://github.com/dcodeIO/protobuf.js/issues/485

dcodeIO:
Oh, this was obviously closed automatically while moving branches.

PR #381, opened 2016-01-05
Exclude tests dir from being published

superkhau:
The gtfs-realtime.proto uses Creative Common licensed code which is not
compatible with Apache licensed code. This fix prevents the publishing
of the tests dir (and it's licensed files), which allows third party
vendors to use the published project.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,4 @@
 .idea
 node_modules/
 npm-debug.log
+tests


superkhau:
@dcodeIO Please review. See comment at https://github.com/dcodeIO/protobuf.js/issues/379#issuecomment-169104073.

dcodeIO:
As far as I know it is good practice to ship the tests with npm. I still don't understand the trouble the file causes for you. If you do not want the tests to be included in your project, why not just add a build step to delete them?

superkhau:
I see you removed it already in #381. Thanks for addressing the concern. Can you do a release so I can get your latest changes on NPM?

>As far as I know it is good practice to ship the tests with npm. I still don't understand the trouble the file causes for you. If you do not want the tests to be included in your project, why not just add a build step to delete them?

+1 for best practices, but the licensing in the tests really prevents us from using the published module. The simplest fix was to skip all tests when publishing. It's not so vital to publish tests since most of them are run during dev time (ie. when people clone your project).

PR #398, opened 2016-03-01
Fix using constructor to copy a message containging oneofs

DouglasHeriot:
See test case I added for example of what was failing.
It throws an error complaining the the virtual oneof field doesn't exist.

Fixed by not trying to set a virtual oneof field in the first place, when setting a property from an object.
Required creating a _oneofsByName map for efficient lookup.


mhahn:
+1 ran into this yesterday as well

InfinitiesLoop:
+1 we have run into this

PR #402, opened 2016-03-08
Build fixes

warrenfalk:
Fix several build issues:
1. Fix a bug that is causing a huge number of tests to fail with "i is not defined"
2. Replace deprecated `@expose` with `@export` to silence warnings
3. Bring the non-minified dist scripts in sync with src


warrenfalk:
As an aside, is there a document or thread explaining the reasoning behind having derived files (like those in dist) be included in source control?  It looks like most pull requests contain changes to src plus the corresponding changes to the unminified dist files, but then, not always.

PR #421, opened 2016-04-16
improves perf of  and cloneRaw by several times

InfinitiesLoop:
Switching to Object.keys and enumerating is much faster because it avoids repeated calls to hasOwnProperty, and since these are protobuf messages, they have inherited members. In my local test with a medium size proto (~30 properties with 3 or 4 sub objects), this improved perf by 6x. I'm using this in a system that needs high performance, and this is almost all of the CPU bound work my process is doing.


mattness:
+1

PR #428, opened 2016-05-11
Check for file length to catch timed out requests

gregor:
A request timeout was not properly handled by Util.fetch

https://github.com/dcodeIO/protobuf.js/issues/427


dcodeIO:
To my understanding this would also wrongfully detect empty .proto files as errors, right?

gregor:
Why wrongfully? An empty proto file is not really a use case with value.

PR #444, opened 2016-06-15
update reference to js to copy for running the example

fabiofumarola:
When I tried to run the websocket example it returned me an error because it was looking for long.min.js, ByteBuffer.min.js and Protobuf.min.js in a wrong folder.


gwokae:
+1 same issue

PR #470, opened 2016-09-07
pbjs: write RPC method stream info to JSON output

### LINKED ISSUE #468: Streaming RPCs
GRPC adds the following syntax:

```
service MyService {
  rpc StreamSomething(stream MyRequest) returns (stream MyResponse);
}
```

As far as I can tell the ProtobufJS RPC wrapper doesn't support this right now. Can this be done? 

This works in GRPC with an object which adds a "Send" function, and a channel for receiving messages (basically a buffered stream).

A few interesting things:

 - An error will close the stream. This means a callback would be needed for close, with an optional error param.
 - Server-to-client, client-to-server, or bidirectional streams are possible.

If it's supported in protobufJS to do something like this, I can write another library that does websocket transport of RPC calls.

paralin:
The pbjs tool currently does not export RPC method stream information to the JSON format, however, the Builder expects it in the JSON input. This is fixed by setting `request_stream` and `response_stream` in the JSON output of pbjs.

Fixes #468

---
This change is [Reviewable](https://reviewable.io/reviews/dcodeio/protobuf.js/470)

### DIFF
--- a/cli/pbjs/targets/json.js
+++ b/cli/pbjs/targets/json.js
@@ -245,7 +245,9 @@ function buildService(svc) {
     svc.getChildren(ProtoBuf.Reflect.Service.RPCMethod).forEach(function(mtd) {
         rpc[mtd.name] = {
             "request": svc.qn(mtd.resolvedRequestType),
+            "request_stream": mtd.requestStream,
             "response": svc.qn(mtd.resolvedResponseType),
+            "response_stream": mtd.responseStream,
             "options": buildOptions(mtd.options)
         };
     });


paralin:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dcodeio/protobuf.js/470)*

paralin:
@dcodeIO this is a pretty minor patch, could you have a look?

Issue #368, opened 2015-12-01
Loading deep proto files having "imports"


rektide:
Hello,

Other examples of protobuf software seem to have some concept of a base directory where protobufs will be imported from, but I'm having trouble loading proto files in protobuf.js. Whenever I have an import, it seems like protobuf.js expects that proto to be in that directory or a subdirectory.

Specifically, I've been trying to get a Node project up that can talk to Etcd's [new v3 protobuf/grpc interfaces](https://github.com/coreos/etcd/blob/master/etcdserver/etcdserverpb/rpc.proto). I'm having trouble getting the .loadProto method to load this proto file and it's imports without fairly brutal hackery.

The proto files seem to assume there's a base directory. Rpc.proto is in this `etcdserver/etcdserverpb/rpc.proto` folder, yet it has these imports:
`import "gogoproto/gogo.proto";
import "etcd/storage/storagepb/kv.proto";`

So when I try to load the rpc.proto file, it joins the dir of rpc.proto with the specified, expecting for example a etcdserver/etcdserver/gogoproto/gogo.proto file, whereas Etcd/Go-protobuf-tooling seems cognizant of that gogoproto/ is a top level directory for the project.

I've had to add a bunch of symlinks to my project, such that I have rpc.proto symlinked into a "top-level directory", and a symlink for etcd/storage/storagepb/gogoproto to satisfy etcd/storage/storagepb/kv.proto's duplicate ask for `gogoproto/gogo.proto`.

Ideally, I'd like to be able to do loadProto on each dependency in order, and [I had a go at that](https://github.com/rektide/etcd-pump/commit/61e5d65bec202c66cc3a90528296f27c61cfbfc1#diff-eda7e77908868b8af64e4390580de22eR13) that I thought ought work. I had to add the above mentioned symlinks, and worse, it indicates that it's not me loading the imports ahead of time: now loadProto is searching for imports on the fs, rather than me providing it to it ahead of time. So I'd have to add `etcd/storage/storagepb/gogodeps/gogo.proto` as another explicit, ahead of time import on line 14 in my code, where I enumerate deps.

Looking at the data-structures protobuf.js generates, it seems to use full paths for dependencies. It seems like other protobuf tooling has some notion of a top level directory, that it can also use to search for files from. That concept could be added, or trivially/hackierly, perhaps protobuf.js could try recursing up the directory tree, looking for imports until satisfied.

[...]

PR #474, opened 2016-09-29
specify include path to resolve imported file

umegaya:
this pull request aims to add functionality to specify include path to resolve imported proto file, stated in #368 

include path can be specified via Builder object like following:

```
var builder = ProtoBuf.newBuilder({
    includePath:[
        './proto'
    ]
});
var protoBuilder = ProtoBuf.loadProtoFile('./proto/foo.proto', builder);
```

I know you don't want to support resolving path so much, but its kind of "nice to have" feature, right?


aidansteele:
This would be very useful for the `pbjs` CLI tool. I would like to be able to pass in multiple paths, e.g. `pbjs -p . -p /usr/local/opt/protobuf/include file.proto` and have it able to import both my proto definitions and Google's definitions bundled with `protoc`.

umegaya:
@aidansteele thank you for your suggestion! but it means all .proto files need to convert to .json file once, and program loads them rite? if so, I think loading .proto directly is more handy way, if purpose is running tool. especially we have a lot of .proto files like [this](https://github.com/AeonLucid/POGOProtos). I wonder why protobuf.js itself does not support this.

Issue #469, opened 2016-09-06
pbjs: proto3 syntax not declared anywhere in JSON output


paralin:
Throughout protobuf.js there are checks if the syntax is proto2 or proto3, but it seems that pbjs generating JSON representations of proto3 does not have anything mentioning proto3 in it.

Is this correct? I'm getting some errors now regarding serializing a object, it expects it to be a string for some reason.

paralin:
I think the issue is that the JSON representation, when generating over multiple proto files at once, the "package" at the top is null, and there is no "syntax" field as there is no top level proto file.

aybmab:
Hey @paralin, did you ever get around this?

dcodeIO:
Closing this for now.

[protobuf.js 6.0.0](https://github.com/dcodeIO/protobuf.js/releases/tag/6.0.0)

Feel free to send a pull request if this is still a requirement.

PR #483, opened 2016-10-31
Declare namespaces and syntax in pbjs json output

### LINKED ISSUE #482: Sub-namespaces in protobuf.js json builder
It's possible to have sub-namespaces with the pbjs json output. The builder incorrectly loads them as Messages. It would be good to instead label these as `className == Namespace`.

paralin:
This patch declares namespaces in json output by introducing the isNamespace flag. This is backwards compatible as the previous errant behavior would instantiate messages for namespaces. This will still happen if an old version of pbjs is used, but this works fine, anyway.

Syntax fields are also set in the output.

Fixes #482 and #469

---
This change is [Reviewable](https://reviewable.io/reviews/dcodeio/protobuf.js/483)

### DIFF
--- a/cli/pbjs/targets/json.js
+++ b/cli/pbjs/targets/json.js
@@ -58,6 +58,7 @@ json.description = description;
 function buildNamespace(ns, out) {
     var messages, enums, services;
     util.extend(out, {
+        "syntax"   : ns.syntax   || 'proto2',
         "options"  : out.options || {},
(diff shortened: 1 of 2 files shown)


paralin:
@dcodeIO I'm assuming you closed this for the rewrite?

dcodeIO:
Afaik this has been closed automatically when switching branches.

PR #498, opened 2016-11-28
set resolvedResponseType on resolve(), fixes #497

### LINKED ISSUE #497: Response type not resolved in service methods
Calling `resolve()` on a service method properly set `resolvedRequestType` but `resolvedResponseType` is always null.

.proto :  
```protobuf
syntax = "proto3";

package example1;

message MyRequest {
  string foo = 1;
}
message MyResponse {
  string bar = 1;
}
service MyTestService {
  rpc getTest(MyRequest) returns (MyResponse);
}
```

.js :  
```javascript
protobuf
  .load(proto)
  .then(root => {
    const service = root.lookup('example1.MyTestService');
    for (let name in service.methods) {
      const method = service.methods[name];
      method.resolve();
      console.log(method.resolvedRequestType instanceof protobuf.Type);
      console.log(method.resolvedResponseType instanceof protobuf.Type);
    }
  });
```

output :  
```bash
true
false
```

subk:


dcodeIO:
Thanks!

PR #501, opened 2016-11-29
ignore "duplicate name" errors on `pbjs`

endel:
**Scenario:**

When I try to generate a `bundle.json` from multiple proto files that require the same dependency, I get `"duplicate name"` exception. See below.

**file1.proto**

```
import "common.proto"
```

**file2.proto**

```
import "common.proto"
```

Command:

```
pbjs file1.proto file2.proto 
```

Outputs:

```
Error: duplicate name '...' in Namespace
```


dcodeIO:
Will look into this. Can't accept the PR when it just removes a valid error.

Issue #508, opened 2016-12-01
WriterPrototype use non-exist zero property


dreamershl:
Is it possible to add the string type support as before? Since javascript doesn't support long natively, my app always use the Long.toString(). When encode the message, the previous version will use the Long to parse the string back to Long.  But in the new version, this can't be supported anymore.   

1. longbits.js doesn't export this zero

        var zero = new LongBits(0, 0);

2. writer.js directly use it. when the value isn't valid uint64, the exception will be thrown 

       WriterPrototype.uint64 = function write_uint64(value) {
          var bits;
          if (typeof value === 'number')
             bits = value ? LongBits.fromNumber(value) : LongBits.zero;
          else if (value.low || value.high)
             bits = new LongBits(value.low >>> 0, value.high >>> 0);
        else
           bits = LongBits.zero;
        return this.push(writeVarint64, bits.length(), bits);
         };

PR #509, opened 2016-12-01
add the support to Long string #508

dreamershl:


dcodeIO:
This is working now. Note, though, that always converting hence and forth between longs and strings is slow. Ideally, when encoding a message multiple times, the value should be present as a long instead of a string to prevent multiple conversions.

dreamershl:
Thank u very much. the converting is only done when send/receive the message from the server in my system. So it should be fine.

PR #514, opened 2016-12-02
improve TypeScript support. add simple test script.

endel:
See https://github.com/dcodeIO/protobuf.js/issues/503


dcodeIO:
Nice, that seems to work!

PR #518, opened 2016-12-05
Disambiguation for `load` interface

eordano:
This change allows the typechecker to know the type that is going to be returned.

### DIFF
--- a/types/protobuf.js.d.ts
+++ b/types/protobuf.js.d.ts
@@ -322,7 +322,10 @@ declare module "protobufjs" {
     * @returns {Promise<Root>|Object} A promise if callback has been omitted, otherwise the protobuf namespace
     * @throws {TypeError} If arguments are invalid
     */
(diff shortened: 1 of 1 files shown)


dcodeIO:
Unfortunately, the .d.ts file is generated automatically through tsd-jsdoc and as thus needs a fix on jsdoc level.

dcodeIO:
Should work without type hints now (I believe nobody would use the non-promise return value anyway).

eordano:
Hey, sorry for the late reply. Thanks for fixing this!

Using typescript 1.8 it complains about "Object" not having a `then` property. And I wouldn't like to make the `as Promise...` correction, too verbose :)

dcodeIO:
The fix isn't published on npm, yet, so you are probably still using a .d.ts with the `Promise<Root> | Object` return signature. The fix above changed the return signature to `Promise<Root> | undefined` which properly infers to `Promise<Root>` when appending `.then(...)`.

PR #541, opened 2016-12-11
Fix to reflection documentation

### LINKED ISSUE #539: Reflection Documentation has error
There is an error in the documentation where a code example is given for reflection. The constructor for Field takes a name as the first argument and an Id as the second, currently, those are inverted.

jonathon-love:
Fixes #539

in other news, how do you feel about my editor stripping all this whitespace?


dcodeIO:
I feel jealous because my editor doesn't do this. Thanks again!

PR #558, opened 2016-12-14
Fix asJSON defaults option, make it work for repeated fields.

conorbranagan:
Fixing two issues in here:

1. When the `defaults` option was enabled, we were using `Object.keys(this)`
   which ends up returning an array that includes things like `$root` and
   asJSON. Instead we can use the `fields` we already have available.

2. For repeated values we were not setting the empty list default because the
   length check would break earlier. We're now passing through when the
   defaults option is set as well.

Added a test on this stuff as well.

### DIFF
--- a/src/message.js
+++ b/src/message.js
@@ -3,7 +3,7 @@ module.exports = Message;
 
 /**
  * Constructs a new message instance.
- * 
+ *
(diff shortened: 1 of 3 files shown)


dcodeIO:
Nice, thanks!

PR #561, opened 2016-12-15
Apply asJSON recursively

mmikulicic:
otherwise the options passed to `asJSON` (e.g. `{enum: String}`) don't take effect in children


dcodeIO:
Thanks!

PR #562, opened 2016-12-15
Make validate support optional fields

mmikulicic:


dcodeIO:
Optional fields should already be checked [here](https://github.com/dcodeIO/protobuf.js/blob/master/src/verify.js#L270) when present.

mmikulicic:
yes I added this check because that generated function was entered with m == null when an optional field was omitted. or at least the stack trace said that null has no such property. afk atm, I can attach a failing case when I'm back

dcodeIO:
I assume that the `null` value there is representing an optional message. I added a commit above, give it a try when you're back.

mmikulicic:
I confirm that `master` no longer exhibits the issue, abandoning this PR. Thank you!

PR #571, opened 2016-12-18
Fix directly using Buffer instead of util.Buffer

fnlctrl:
This seems to have caused webpack to load a Buffer shim when building for web.

### DIFF
--- a/src/field.js
+++ b/src/field.js
@@ -287,7 +287,7 @@ FieldPrototype.jsonConvert = function(value, options) {
             if (options.bytes === Array)
                 return Array.prototype.slice.call(value);
             if (options.bytes === util.Buffer && !util.Buffer.isBuffer(value))
-                return util.Buffer.from ? util.Buffer.from(value) : new Buffer(value);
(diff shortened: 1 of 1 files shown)


dcodeIO:
Thanks!

dcodeIO:
Btw, doesn't [this](https://github.com/dcodeIO/protobuf.js/blob/master/src/util/runtime.js#L26) cause the same?

fnlctrl:
Yes, this PR didn't solve the problem, and I forgot to leave a comment here.
Since it's dynamic require, I can't seem to think of a way to properly solve it.. So as a workaround, I now use `webpack.IgnorePlugin` to force disable loading Buffer/Long.
```js
plugins: [
    new webpack.IgnorePlugin(/^long$/),
    new webpack.IgnorePlugin(/^buffer$/)
]
```

dcodeIO:
This now uses a dynamic require implementation for all optional modules including browser and long. Give it a shot!

fnlctrl:
Works like a charm! Thanks for the awesome work! 👍👍👍

PR #607, opened 2017-01-02
corrected import typings for generated proto-classes

halex2005:


dcodeIO:
That's [not correct](https://github.com/dcodeIO/protobuf.js/blob/master/tests/data/test.d.ts#L13).

halex2005:
My PR is for v6.3.0, because there was no usage for $protobuf.Writer

I tested your comment with
```
./bin/pbjs -t static --out tests/data/test.js tests/data/test.proto
./bin/pbts --out tests/data/test.d.ts --no-comments tests/data/test.js
```

On current master it is ok, like your wrote, then this PR is not necessary.

On 6.3.1 pbts emit error:

```
ERROR: Unable to parse /home/hal/git/protobuf.js/tests/data/test.js: Unexpected token .
```

On 6.3.0 my PR is necessary.

So, when v6.4.0 will come out? (current master)

dcodeIO:
Closing this for now as 6.4.0, which is now released, should solve the issue.

PR #610, opened 2017-01-03
use Long type from $protobuf

halex2005:


dcodeIO:
This is probably the correct thing to do but it has the side effect that the (overly simple) Long interface from protobuf.js is used. Hope it still works correctly when a real long.js definition is in place.

PR #660, opened 2017-01-26
Changes callback on failed response decode in rpc service

Kenny-House:
Currently, the string "error" is passed to the callback on response decode errors (with the actual error passed where the decoded response would normally be). 

This updates the rpc service to pass the error as the first param to the callback.


dcodeIO:
Thanks!

PR #665, opened 2017-01-28
add support for node 4.2.0 to 4.4.7 with specific buffer polyfill

nddeluca:
This PR enables support for Node 4.3.2 (which I'm stuck with on AWS Lambda).

The `Buffer` object in Node 4.0, 4.1, 4.1, and 4.4 have a `Buffer.from` method defined which is inherited from `Uint8Array`.  The current polyfill only checks if `from` is present and the `from` method in these versions accept a typed array as an argument, which doesn't behave nicely.

This PR replaces `util.Buffer` with a new object on those node versions and brings compatibility to Node 4.2.0 to 4.4.7.  Unfortunately, `Buffer` can't be inherited from on 4.0 and 4.1, so this polyfill doesn't work there.

Ran tests locally with node 4.3.2 and they all passed and I'm also using this PR to decode protobuf messages in lambda through the static code generation.

Maybe node 4.3.2 should be added to travis to prevent regression?

### DIFF
--- a/src/util/minimal.js
+++ b/src/util/minimal.js
@@ -70,6 +70,24 @@ util.Buffer = (function() {
         if (!Buffer.prototype.utf8Write) // refuse to use non-node buffers (performance)
             return null;
 
+        // Node 4.2.0 to 4.4.7 Support.  Unfortunatly not 4.0.0 - 4.1.2 compatible
+        // See - https://github.com/nodejs/node/pull/3080 for reference
+        if (Buffer.from && Buffer.from.toString() === Uint8Array.from.toString()) {
+          function PolyBuffer(arg, encoding) {
+            var b = new Buffer(arg, encoding);
+            Object.setPrototypeOf(b, PolyBuffer.prototype);
+            return b;
+          }
+
+          PolyBuffer.from = function(value, encoding) { return new PolyBuffer(value, encoding); };
+          PolyBuffer.allocUnsafe = function allocUnsafe(size) { return new PolyBuffer(size); };
+
+          Object.setPrototypeOf(PolyBuffer.prototype, Buffer.prototype);
+          Object.setPrototypeOf(PolyBuffer, Buffer);
+
+          return PolyBuffer;
+        }
+
         /* istanbul ignore next */
         if (!Buffer.from)
             Buffer.from = function from(value, encoding) { return new Buffer(value, encoding); };


dcodeIO:
Hmm, what if we'd just change the polyfill to account for this:

```js
        if (!Buffer.from || Buffer.from === Uint8Array.from)
            Buffer.from = function from(value, encoding) { return new Buffer(value, encoding); };
```

This of course assumes that nobody is intentionally using broken .from.

dcodeIO:
As an alternative:

```js
        if (!Buffer.from)
            Buffer.from = function from(value, encoding) { return new this(value, encoding); };
        else if (Buffer.from === Uint8Array.from)
            Buffer.from = function from(a, b, c) {
                if (typeof b === "string" || arguments.length === 1)
                    return new this(a, b);
                 return Uint8Array.from(a, b, c);
            };
```

nddeluca:
@dcodeIO Hmm, I yeah I like that method better, but initially tried that but was getting
```
TypeError: Cannot assign to read only property 'from' of function Buffer(arg, encoding) {
  // Common case.
  if (typeof arg === 'number') {
    // If less than zero, o...<omitted>...
}
    at repl:1:27
    at REPLServer.defaultEval (repl.js:252:27)
    at bound (domain.js:287:14)
    at REPLServer.runBound [as eval] (domain.js:300:12)
    at REPLServer.<anonymous> (repl.js:417:12)
    at emitOne (events.js:82:20)
    at REPLServer.emit (events.js:169:7)
    at REPLServer.Interface._onLine (readline.js:211:10)
    at REPLServer.Interface._line (readline.js:550:8)
    at REPLServer.Interface._ttyWrite (readline.js:827:14)
```
when I tried to replace Buffer.from.  Seems strict mode prevents us from doing so?

dcodeIO:
Well, I assume your initial approach is the right one then. Going to get a node version for testing.

dcodeIO:
This commit changes how Buffer.from / Buffer.allocUnsafe are accessed. These are now private properties on `util`, either being just an alias or a polyfill where necessary.

Also added node 4.3.2 to Travis as you suggested.

nddeluca:
Nice refactor 👍 Closing this PR since master works for me now locally and on Lambda.

I also ran the test suite on node 4.0.0 and 4.1.2, and all tests pass -- so I believe you can change your comment to support all 4.x versions since you've refactored around the Buffer inheritance bug in the earlier versions that I wasn't able to solve.

PR #676, opened 2017-02-11
Full support for reserved type

schwiet:
Only this case was being handled:

`reserved 10 to 20;`

But all of the following are valid ([source](https://developers.google.com/protocol-buffers/docs/reference/proto3-spec#reserved)):

```
reserved 2, 15, 9 to 11;
reserved "foo", "bar";
```

Files with reserved names or comma separated numbers would not compile. This is my attempt to handle all cases in a way that is consistent with the existing codebase.

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -147,6 +147,39 @@ function parse(source, root, options) {
         return [ start, end ];
     }
 
+    function readReserved(){
+      var r_list = [];
+      var val;
+      var add_range = false;
+      var i;
+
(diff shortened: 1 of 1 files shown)


dcodeIO:
Please let me know if the latest commit solves your issue!

dcodeIO:
Should be working as of 6.6.4. Feel free to reopen if there are still any issues!

schwiet:
Yes, looks good, thanks!

PR #683, opened 2017-02-20
Fixing the packed=true by default problem for proto2 files.

serkangunes:
Fixing the packed=true by default problem for proto2 files. Packed should be false by default for repeated enums when the syntax is set to proto2.

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -336,7 +336,8 @@ function parse(source, root, options) {
             field.comment = cmnt(trailingLine);
         // JSON defaults to packed=true if not set so we have to set packed=false explicity when
         // parsing proto2 descriptors without the option, where applicable.
-        if (field.repeated && !isProto3)
+        var packed = field.getOption('packed');
+        if (field.repeated && packed === undefined && !isProto3)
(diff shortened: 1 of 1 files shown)


dcodeIO:
Manually merged your changes and also added some code to `Field#resolve` removing unnecessary packed options from parsing again.

serkangunes:
Ok shall I close this pull request then?

serkangunes:
I have just had a look at your fix. It does fix my issue however with syntax=proto2 you can still set the repeated fields to be packed. It is just by default unpacked.

```
var packed = field.getOption('packed');
// There reason packed === undefined check is here because it the user has set the packed option 
// value we should not override it as false.
if (field.repeated && packed === undefined && !isProto3)
```

dcodeIO:
`field.setOption("packed", false, /* ifNotSet */ true);` only sets the value if it is `undefined` ([see](https://github.com/dcodeIO/protobuf.js/blob/master/src/object.js#L166)).

Is that what you are referring to?

serkangunes:
Ah good one I didn't notice that parameter. I'll close this.

PR #699, opened 2017-03-10
fixed case for AMD module

zeachco:
I think this should fix https://github.com/dcodeIO/protobuf.js/issues/698


zeachco:
wont fix as this version is too outdated

PR #701, opened 2017-03-12
Add a switch to initialize default values

### LINKED ISSUE #700: Empty array causes "Object not extensible"
protobuf.js version: 6.0.0

When decoding messages with empty arrays or initializing a blank message with `new()`, all arrays are non extensible and inherited from the builder prototype.
It makes sense to freeze those array on the prototype, but when using `new()` or `decode()` we should have a working object, not frozen.

Can't we have usable objects without duplicating with `Builder.from(message.toJSON())` and setting `util.toJSONOptions.defaults` to true?

Basically this shouldn't throw any error:
```js
var message = new TestMessage({testList: []})
var buffer = TestMessage.encode(message).finish()
var decoded = TestMessage.decode(buffer)

// decoded.testList is undefined so it falls back to TestMessage.prototype.testList which is frozen       
decoded.testList.push('test str')
```
Replace `{testList: []}` with `{testList: ['something']}` and everything works back again.

I made a [small repro repo](https://github.com/fathyb/protobufjs-empty-arrays) to demonstrate what's going on.
Just run `make` and it should install everything and launch mocha. The first three tests should pass, the last two should fail.

fathyb:
This is in progress and should resolve #700


dcodeIO:
See: https://github.com/dcodeIO/protobuf.js/issues/700

PR #725, opened 2017-03-24
fix: pass output to callback if out is not a file

### LINKED ISSUE #724: Programmatic execution prints to stdout
protobuf.js version: 6.7.0 (current `HEAD` on master)

When running the sample from the README:
```js
var pbjs = require("protobufjs/cli/pbjs");

pbjs.main([ "--target", "json-module", "path/to/myproto.proto" ], function(err, output) {
    if (err)
        throw err;
    // do something with output
});
```
then `output` is `undefined` and the result gets printed on `stdout` instead.

joscha:
Currently, the output is not passed to the callback if the output target is not a file. It is written to stdout even with programmatic access.

This pull request fixes that by only writing to stdout if invoked non-programmatically and only passing the output to the callback if the target is not a file.

fixes #724

### DIFF
--- a/cli/pbjs.js
+++ b/cli/pbjs.js
@@ -278,11 +278,11 @@ exports.main = function main(args, callback) {
             if (output !== "") {
                 if (argv.out)
                     fs.writeFileSync(argv.out, output, { encoding: "utf8" });
-                else
+                else if (!callback)
                     process.stdout.write(output, "utf8");
             }
             return callback
-                ? callback(null)
+                ? argv.out ? callback(null) : callback(null, output)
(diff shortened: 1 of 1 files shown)


joscha:
coverage seem to have failed on `Node.js: 6, CXX=g++-4.8`

dcodeIO:
See: https://github.com/dcodeIO/protobuf.js/issues/724

Basically, I applied your fix to both pbjs and pbts.

joscha:
Nice! Too bad I overlooked pbts and thus don't get the contributors badge 😝

Issue #732, opened 2017-03-27
generated code incorrectly refers to $util instead of util


tamird:
In CockroachDB we have some protos in the `util` package, but the protobufjs generated code trips on this and generates code like this:
```javascript
            UnresolvedAddr.decode = function decode(reader, length) {
                if (!(reader instanceof $Reader))
                    reader = $Reader.create(reader);
                let end = length === undefined ? reader.len : reader.pos + length, message = new $root.cockroach.$util.UnresolvedAddr();
                while (reader.pos < end) {
                    let tag = reader.uint32();
                    switch (tag >>> 3) {
                    case 1:
                        message.network_field = reader.string();
                        break;
                    case 2:
                        message.address_field = reader.string();
                        break;
                    default:
                        reader.skipType(tag & 7);
                        break;
                    }
                }
                return message;
            };
```
Note that `$root.cockroach.$util` should be `$root.cockroach.util`.

dcodeIO:
Got a bit lost in espree's AST so please let me know if this fixes your issue(s). Might affect the PR.

tamird:
Fixes my issue, hurray! Rebased the PR as well.

PR #733, opened 2017-03-27
scripts: simplify, regenerate es6 too

tamird:
Ran into this while trying to investigate #732.


dcodeIO:
Looks good, but could you maybe change it so it only changes the generator script? For instance, google/protobuf/descriptor.json is intentionally missing the redundant `options` keys.

tamird:
What do you mean? Would you like me to modify gencommon to omit empty options?

PR #744, opened 2017-04-04
Update the typescript example in README.md

winksaville:
The update makes the example execute using typescript 2.2.2. The
following changes were made.

 - Changed lookupType parameter to `"awesomepackage.AwesomeMessage"`
 - Use `let` instead of `var` as its consider a better style
 - Add type `: any` for message
 - Add printing of message.awesomeField
 - Add example of decoding of buffer and printing result
 - Commented out redundant or non-executable lines

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,18 +444,21 @@ import * as protobuf from "protobufjs";
 import * as Long from "long"; // optional
 
 // browser only (alternatively)
-import * as protobuf from "./node_modules/protobufjs/index.js";
-import * as Long from "./node_modules/long/dist/long.js"; // optional
+//import * as protobuf from "./node_modules/protobufjs/index.js";
+//import * as Long from "./node_modules/long/dist/long.js"; // optional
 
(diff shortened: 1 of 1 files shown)


winksaville:
I had trouble getting the example working so I thought I provide this PR to help others. I was surprised I needed to explicitly type `message` and `decodedMessage` to `: any`. Is there something I did incorrectly to require the `: any`?

dcodeIO:
> I was surprised I needed to explicitly type message and decodedMessage to : any

This might be the case because you are accessing its properties with dot syntax (`message.awesomeField`) but messages dynamically created at runtime technically have no typings and are just `instanceof protobuf.Message`. Does it work when using `message["awesomeField"]`?

If you'd like to have proper typings for dynamic messages, too, you could still use [pbts-generated typings for its static counterpart](https://github.com/dcodeIO/protobuf.js/blob/master/cli/README.md#pbts) with your dynamic code.

winksaville:
Thanks for the feed back and updating README.md

PR #766, opened 2017-04-19
fix RPCImplCallback type declaration - make Error optional

rossenassenov:


dcodeIO:
index.d.ts is a generated file and any changes will be overwritten on rebuild. Actual source location is: https://github.com/dcodeIO/protobuf.js/blob/master/src/rpc.js#L28

`@param {Error|null}` then, probably.

rossenassenov:
Oh, I see. I wonder why it currently works for the response param, but not the error param ?

dcodeIO:
But is it actually optional? It cannot be `undefined`. There is either an `Error`, or none (`null`).

rossenassenov:
thanks for the fix !

PR #779, opened 2017-04-26
Issues/775

### LINKED ISSUE #775: add unknown field support
protobuf.js version: all, up to the current 6.8.0

proto2 preserve unknown fields for future serialization. This is also planned for proto3 in the near future: https://docs.google.com/document/d/1KMRX-G91Aa-Y2FkEaHeeviLRRNblgIahbsk4wA14gRk/edit#heading=h.w8dtggryroj4

I'm working on a merge request that will add unknown fields when decoding as

_unknownFields: { <id|wireType>: <byte array> }

This should be doable by add read_type_bytes() to Reader and using it instead of skipType. The _unknownField can be later serialized again or discarded by a new method discardUnknownFields().

Tests:
```js
var proto = "message Simple_v1 {\
  optional string name  = 1;\
  optional string value = 3;\
}\
message Simple_v2 {\
  optional string name  = 1;\
  optional int32  flags = 2;\
  optional string value = 3;\
}";

var msg = { inner: [{}, {}, {}] };

tape.test("unknown fields", function (test) {
    var root = protobuf.parse(proto).root,
        Simple_v1 = root.lookup("Simple_v1");
        Simple_v2 = root.lookup("Simple_v2");

    var s2 = Simple_v2.create({ name: "v2", flags: 2, value: "dummy" });
    var s1 = Simple_v1.decode(Simple_v2.encode(s2).finish());

    var restored = Simple_v2.decode(Simple_v1.encode(s1).finish());

    test.equal(s2.name, restored.name, "assert: even known fields are missing");

    test.equal(2, restored.flags, "are preserved by default");
    test.end();
});

tape.test("discarded unknown fields", function (test) {
    var root = protobuf.parse(proto).root,
        Simple_v1 = root.lookup("Simple_v1");
    Simple_v2 = root.lookup("Simple_v2");

    var s2 = Simple_v2.create({ name: "v2", flags: 2, value: "dummy" });
    var s1 = Simple_v1.decode(Simple_v2.encode(s2).finish());

    try {
        s1.discardUnknownFields();
    }
    catch (ex) {
        test.end("discardUnknownFields() exception: " + ex);
        return;
    }

    var restored = Simple_v2.decode(Simple_v1.encode(s1).finish());

    test.equal(undefined, restored.flags, "are removed from the message");
    test.end();
});
```

Does it sound OK?

pavelstudeny:
fixes https://github.com/dcodeIO/protobuf.js/issues/775

* all unknown fields are deserialized into a single array __unknownFields
* __unknownFields are serialized at the end
* discardUnknownFields() remove the unknown fields from the deserialized data
* a global option to ignore unknown fields is not present. if required, please suggest where to place it


src/writer.js
```diff
@@ -380,6 +380,16 @@ Writer.prototype.bytes = function write_bytes(value) {
 };
 
 /**
+ * Writes raw bytes with no wire type of length prefixed
+ * @param {Uint8Array} value bytes to add
+ * @returns {Writer} `this`
+ */
+Writer.prototype.rawBytes = function write_raw_bytes(value) {
```
pavelstudeny: seems that writeBytes (the one using util.Array) might be incorrect on older node versions, but it was probably never called until now

jcgertig:
is this able to just handle encoding unkown Ids that are passed in like say 
```
{
 SomeKnownKey: 1,
 9999: 'Some unknown id'
}
```

jcgertig:
Or do I have to put that id in the __unkown myself?

pavelstudeny:
@jcgertig it creates the __unknownFields automatically. Seeing this issue neither accepted nor rejected for more than a year, however, I'm wondering what's planned with this project now that Google provides [built-in JavaScript support](https://developers.google.com/protocol-buffers/docs/reference/javascript-generated)

chakmingli:
If you fork and apply this change, it will be really helpful for many people.

jeremyong:
@pavelstudeny Is there a reason you opted to make the unknown fields hidden?

pavelstudeny:
@jeremyong  do you mean that the `__unknownFields` have `enumerable` set to `false` by "hidden"? Yes, since they are unknown, there is not much you can do with them in an enumeration.

Issue #794, opened 2017-05-12
decodeDelimited: TypeError: Cannot set property 'buf' of undefined


milgner:
protobuf.js version: 6.7.3

`.proto` file encoded as an ES6 module.

Expected: `decodeDelimited(buffer)` returns a Message.

```
let message = MyMessage.decodeDelimited(buffer);
```

Instead it raises an error:
```
TypeError: Cannot set property 'buf' of undefined
    at Reader (node_modules/protobufjs/src/reader.js:28:14)
    at Function.decodeDelimited (my_message.js:192:56)
```

masonk:
Wrap your buf in a Reader. If you have an `ArrayBuffer`, slap a `Uint8Array` over it too.
```
const reader = new protobuf.Reader(new Uint8Array(msg.data));
let pb = MyMessage.decodeDelimited(reader);
```

milgner:
Thanks a lot, that helped! I only write Javascript every now and then but since that works, I guess the generated code must be missing a `new` before `$Reader`:

```javascript
    function decodeDelimited(reader) {
            if (!(reader instanceof $Reader))
                reader = $Reader(reader);
```

I'll see whether I can't locate the generator code in question and submit a PR...

PR #795, opened 2017-05-14
Fix Reader argument conversion in static module

milgner:
Since the proposed change in #794 worked, I adapted the generator.
After compiling my `.proto` with the modified version, I did not need to manually create a `Reader` anymore, as originally intended.


milgner:
The Travis config for Node 6 seems to be missing a `CODECLIMATE_REPO_TOKEN` environment variable?

dcodeIO:
Thanks! Yeah, the codeclimate repo token is a private config var and I still haven't disabled the respective call to it when it's not present.

PR #813, opened 2017-05-30
Use protobuf/minimal when pbjs target is static-module

### LINKED ISSUE #798: ES6 Static code import `protobufjs` isntead of `protobufjs/minimal`
protobuf.js version: 6.7.3

**Actual behavior**

When generating static code with `es6` wrapping like so:
```shell
$pbjs -w es6 -t static-module message.proto > message.js
```

The `message.js` script imports `protobufjs`:
```js
import * as $protobuf from "protobufjs";
```

**Expected behavior**
The `message.js` script should import `protobufjs/minimal`, isn't it?

betalb:
This should fix #798 

Change affects AMD wrapper too, to be consistent.

Some modifications to AMD loader configs may be required, i.e. `protobufjs/minimal` should be mapped to correct file. Other option is to pass `dependency` argument to pbts and override default module name.

### DIFF
--- a/cli/pbjs.js
+++ b/cli/pbjs.js
@@ -33,7 +33,7 @@ exports.main = function main(args, callback) {
             "force-long": "strict-long",
             "force-message": "strict-message"
         },
-        string: [ "target", "out", "path", "wrap", "root", "lint" ],
+        string: [ "target", "out", "path", "wrap", "dependency", "root", "lint" ],
         boolean: [ "create", "encode", "decode", "verify", "convert", "delimited", "beautify", "comments", "es6", "sparse", "keep-case", "force-long", "force-number", "force-enum-string", "force-message" ],
         default: {
             target: "json",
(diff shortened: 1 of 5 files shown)


cli/pbjs.js
```diff
@@ -99,6 +99,8 @@ exports.main = function main(args, callback) {
                 "                   es6       ES6 wrapper (implies --es6)",
                 "                   closure   Just a closure adding to protobuf.roots (see -r)",
                 "",
+                "  --dependency     Specifies which version of protobuf to require. Accepts any valid module id",
+                "",
```
dcodeIO: Currently, `$DEPENDENCY` is defined by the respective target and if I am not mistaken it is not (yet) a used argument.

betalb: If I understood your concern correct, it is actually used in `util.wrap` called by `static-module` and `json-module` targets. This option gets correctly passed down to `(json|static)-module` target.

I only see one concern at the moment: `static-module` target sepcifies default value for `dependency`, but `json-module` - doesn't

betalb: To be more specific. util.wrap uses `protobufjs` as default value when dependency option is not provided.
`static-module` target will use `protobufjs/minimal` as default if not overwritten by options, provided in arguments, but `json-module` is not doing this.

To make things consistent we can either remove default value from `static-module` code and specify default in `pbjs.js` or update `json-module` target

dcodeIO: Ideally, the argument should always be honored in case the user has a legitimate reason to override it, but the targets should have their defaults.

betalb: And it will. `dependency` will be passed inside of options object to target. Target will set `dependency` to `protobufjs/minimal` if not provided externally, and will pass options to `util.wrap`.

betalb:
Rebased onto master

PR #844, opened 2017-06-24
remove null from optional field type union

### LINKED ISSUE #834: Generated typescript file is not usable anymore
protobuf.js version: 6.8.0

Here's a part of my proto file:

```
message RequestUpdateDocument
{
        required ActionType Action = 1;
        optional WebMeetingDocument Document = 2;
        optional string Content = 3;
        optional int32 MeetingId = 4; //only for 3cxPhone clients. Not for WM user.
}
```

Here's a part of generated typescript file using pbts:

```
        /** Properties of a RequestUpdateDocument. */
        interface IRequestUpdateDocument {

            /** RequestUpdateDocument Action */
            Action: MyPhone.Notifications.ActionType;

            /** RequestUpdateDocument Document */
            Document?: MyPhone.Notifications.IWebMeetingDocument;

            /** RequestUpdateDocument Content */
            Content?: string;

            /** RequestUpdateDocument MeetingId */
            MeetingId?: number;
        }

        /** Represents a RequestUpdateDocument. */
        class RequestUpdateDocument {

            /**
             * Constructs a new RequestUpdateDocument.
             * @param [properties] Properties to set
             */
            constructor(properties?: MyPhone.Notifications.IRequestUpdateDocument);

            /** RequestUpdateDocument Action. */
            public Action: MyPhone.Notifications.ActionType;

            /** RequestUpdateDocument Document. */
            public Document?: (MyPhone.Notifications.IWebMeetingDocument|null);

            /** RequestUpdateDocument Content. */
            public Content: string;

            /** RequestUpdateDocument MeetingId. */
            public MeetingId: number;

            /**
             * Decodes a RequestUpdateDocument message from the specified reader or buffer.
             * @param reader Reader or buffer to decode from
             * @param [length] Message length if known beforehand
             * @returns RequestUpdateDocument
             * @throws {Error} If the payload is not a reader or valid buffer
             * @throws {$protobuf.util.ProtocolError} If required fields are missing
             */
            public static decode(reader: ($protobuf.Reader|Uint8Array), length?: number): MyPhone.Notifications.RequestUpdateDocument;

            /**
             * Decodes a RequestUpdateDocument message from the specified reader or buffer, length delimited.
             * @param reader Reader or buffer to decode from
             * @returns RequestUpdateDocument
             * @throws {Error} If the payload is not a reader or valid buffer
             * @throws {$protobuf.util.ProtocolError} If required fields are missing
             */
            public static decodeDelimited(reader: ($protobuf.Reader|Uint8Array)): MyPhone.Notifications.RequestUpdateDocument;
        }
```

You see that optional fields Document, Content, MeetingId have different types which makes this classes incompatible and this unusable. Class doesn't implement interface.

Expected behavior:

Interface and class should have same propeties. Probably it even makes sense to make class implement interface.

### LINKED ISSUE #837: Generated TypeScript message class not compatible with generated message interface
protobuf.js version: 6.8.0

Problem:
Message class instances cannot be assigned to interface typed variable (or function parameter);

Generated TypeScript definitions.
```ts
    interface IToken {
        message?: google.protobuf.IAny;
    }

    class Token {
        constructor(properties?: core.IToken);
        public message?: (google.protobuf.IAny|null);
        public static encode(message: core.IToken, writer?: $protobuf.Writer): $protobuf.Writer;
        public static encodeDelimited(message: core.IToken, writer?: $protobuf.Writer): $protobuf.Writer;
        public static decode(reader: ($protobuf.Reader|Uint8Array), length?: number): core.Token;
        public static decodeDelimited(reader: ($protobuf.Reader|Uint8Array)): core.Token;
        public static verify(message: { [k: string]: any }): (string|null);
    }
```

Code resulting in TypeScript Error:
```ts
const myEncodedToken = Token.encode(myToken);
```

TypeScript Error:
```
Error:(42, 37) TS2345:Argument of type 'Token' is not assignable to parameter of type 'IToken'.
  Types of property 'message' are incompatible.
    Type 'IAny | null | undefined' is not assignable to type 'IAny | undefined'.
      Type 'null' is not assignable to type 'IAny | undefined'.
```

A similar problem was fixed in response to this issue: https://github.com/dcodeIO/protobuf.js/issues/826.  The types now correctly reflect that an instantiated message class can have undefined fields.  But assignment to the interface (by calling `Message.encode` for example) is still impossible.

Example generated TypeScript interface which would fix the problem:
```ts
    interface IToken {
        message?: google.protobuf.IAny | null;
    }
```

Alexendoo:
Messages with optional fields will now implement their corresponding interfaces (IMessage)

fixes #834
fixes #837

### DIFF
--- a/cli/targets/static.js
+++ b/cli/targets/static.js
@@ -391,7 +391,7 @@ function buildType(ref, type) {
             push("");
             var jsType = toJsType(field);
             if (field.optional && !field.map && !field.repeated && field.resolvedType instanceof Type)
-                jsType = "(" + jsType + "|null|undefined)";
+                jsType = "(" + jsType + "|undefined)";
             pushComment([
                 field.comment || type.name + " " + field.name + ".",
                 "@member {" + jsType + "}" + escapeName(field.name),
--- a/tests/data/package.d.ts
+++ b/tests/data/package.d.ts
@@ -28,7 +28,7 @@ export class Package {
     public description: string;
     public author: string;
     public license: string;
-    public repository?: (Package.IRepository|null);
+    public repository?: Package.IRepository;
     public bugs: string;
     public homepage: string;
     public keywords: string[];
--- a/tests/data/package.js
+++ b/tests/data/package.js
@@ -104,7 +104,7 @@ $root.Package = (function() {
 
     /**
      * Package repository.
-     * @member {(Package.IRepository|null|undefined)}repository
+     * @member {(Package.IRepository|undefined)}repository
      * @memberof Package
      * @instance
      */
--- a/tests/data/test.d.ts
+++ b/tests/data/test.d.ts
@@ -121,7 +121,7 @@ export namespace jspb {
             constructor(properties?: jspb.test.IOptionalFields);
             public aString: string;
             public aBool: boolean;
-            public aNestedMessage?: (jspb.test.OptionalFields.INested|null);
+            public aNestedMessage?: jspb.test.OptionalFields.INested;
             public aRepeatedMessage: jspb.test.OptionalFields.INested[];
             public aRepeatedString: string[];
             public static create(properties?: jspb.test.IOptionalFields): jspb.test.OptionalFields;
@@ -196,7 +196,7 @@ export namespace jspb {
             constructor(properties?: jspb.test.IComplex);
             public aString: string;
             public anOutOfOrderBool: boolean;
-            public aNestedMessage?: (jspb.test.Complex.INested|null);
+            public aNestedMessage?: jspb.test.Complex.INested;
             public aRepeatedMessage: jspb.test.Complex.INested[];
             public aRepeatedString: string[];
             public static create(properties?: jspb.test.IComplex): jspb.test.Complex;
@@ -382,7 +382,7 @@ export namespace jspb {
         class TestClone {
             constructor(properties?: jspb.test.ITestClone);
             public str: string;
-            public simple1?: (jspb.test.ISimple1|null);
+            public simple1?: jspb.test.ISimple1;
             public simple2: jspb.test.ISimple1[];
             public bytesField: Uint8Array;
             public unused: string;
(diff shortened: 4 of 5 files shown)


djrenren:
Thanks! Hopefully this gets merged soon, cuz this bug has been giving me headaches.

vaiwa:
Thanks, How does it look with the merge?

Alexendoo:
@vaiwa instead of having optional fields typed as `Field|undefined|null`, they now match the interfaces in using `Field|undefined`. The result being you can use a Message as an IMessage rather than having to use type casting when using typescript

dcodeIO:
The reason that I'm hesitating here is that one can actually assign `null` without any consequences. All code explicitly checks for ´undefined` or `null`. Some dependents might already be doing that, so this change might, or might not, break their code.

Alexendoo:
Yeah this is a breaking change as far as type definitions go, but the benefits far outweigh the negatives (one can simply use undefined where they previously used null if migration is neccessary). Just `x|undefined` is the canonical way to describe optional in typescript though

djrenren:
To be fair, this change is only breaking under `strictNullChecks: true`

The real issue here is that without this change, the interfaces are unusable alongside the concrete types.

dcodeIO:
Does [this commit](https://github.com/dcodeIO/protobuf.js/commit/ed7e2e71f5cde27c4128f4f2e3f4782cc51fbec7) fix anything in this regard?

Conaclos:
Any news?

Szpadel:
@dcodeIO your commit makes things worse,
right now it broken our project that have `strictNullChecks: true`

Szpadel:
@Alexendoo could you create the same pull request to my [fork](https://github.com/Szpadel/protobuf.js)?
I'm trying to create working protobufjs version that we need in production until official version will be fixed.

Cixelyn:
Wait, @Szpadel isn't https://github.com/dcodeIO/protobuf.js/commit/ed7e2e71f5cde27c4128f4f2e3f4782cc51fbec7 sufficient?

Now all types are both possibly `null` and `undefined` which means that all the classes correctly implement the interface. This also more closely represents the API as described since both `null` and `undefined` are assignable @dcodeIO.

Szpadel:
@Cixelyn AFAIK protobuf is encoding empty repetable field as missing also all empty fields are just missing ones, therefore it is impossible to have null value here.

When you have `strictNullChecks: true` set in your project, then you cannot map fields directly to `field?: SomeType`, because this means that field is `SomeType | undefined` and protobuf field is `SomeType | null | undefined` and you are not allowing that null value here, of course you could refactor all your entities but this is just inconvenient especially when null value is impossible here.

dcodeIO:
Does this issue still exist in master? There have been a couple merges that might, or might not, prevent this. Otherwise, do you have an updated example causing the issue?

dudv:
@dcodeIO The issue is still present in the latest version. Are there any plans to fix that?

dcodeIO:
Haven't looked at this for a while, but what I can tell is that the correct typing here is that actual instances of `Message` have `OtherMessage | null` fields while message-likes have `OtherMessage | null | undefined`.

Seems that any attempts to resolve this didn't quite cut it because these either drop `null` or `undefined` from the equation in both cases, which isn't quite correct and leads to a "my fix is your bug" kind of situation.

This PR in particular changes

```
public repository?: (Package.IRepository|null);
```

to

```
public repository?: Package.IRepository;
```

on actual message instances, which is incorrect because instances don't have `undefined` but `null` there.

There might be smart ways to model this that I'm not aware of. Any ideas?

mfedderly:
Also related to #1171

On my own project, I stripped out the generated interfaces entirely (between pbjs and pbts), because I am only using the decode methods and everything will be of the class type not the plain javascript object type. Referencing the interface types also causes issues if you extend the generated classes with additional methods.

Would a separate PR be accepted to pbjs to skip generating the interface jsdoc comments? That gives people the flexibility to use the classes only and prevent what seems to be a common pitfall with protobufjs and typescript.

PR #845, opened 2017-06-24
Fixed codegen for maps

### LINKED ISSUE #843: Map fields with empty string as key are handled incorrectly
protobuf.js version: 6.8.0

UPDATED:
Current codegen for map-fields when string key is empty is incorrect.

ORIGINAL:
Current codegen for map-fields is invalid. It assumes that map keys are non empty (!) strings (!!).

1. Map CAN have empty string as indices.
2. ~~Map CAN be indexed by something different than string...~~

~~Both parsing and JS representation (Object instead of Map) are broken.~~

mkosieradzki:
Fixed #843


mkosieradzki:
@dcodeIO Any chance for merge?

mkosieradzki:
@dcodeIO Any chance for merge? It seems that not only I am affected by this one.

hugebdu:
@dcodeIO Any chance for merge?

PR #849, opened 2017-06-26
Proto path option

ricardohbin:
It's very common the proto dependencies between many resources to use some import path base to keep its relative paths with consistence.

This PR add "--proto_path" like feature to protobuf.js.

### DIFF
--- a/src/root.js
+++ b/src/root.js
@@ -103,17 +103,18 @@ Root.prototype.load = function load(filename, options, callback) {
             if (!util.isString(source))
                 self.setOptions(source.options).addJSON(source.nested);
             else {
+                var protoPath = options && options.protoPath || filename;
                 parse.filename = filename;
                 var parsed = parse(source, self, options),
                     resolved,
                     i = 0;
                 if (parsed.imports)
                     for (; i < parsed.imports.length; ++i)
-                        if (resolved = self.resolvePath(filename, parsed.imports[i]))
+                        if (resolved = self.resolvePath(protoPath, parsed.imports[i]))
                             fetch(resolved);
                 if (parsed.weakImports)
                     for (i = 0; i < parsed.weakImports.length; ++i)
-                        if (resolved = self.resolvePath(filename, parsed.weakImports[i]))
+                        if (resolved = self.resolvePath(protoPath, parsed.weakImports[i]))
                             fetch(resolved, true);
             }
         } catch (err) {
--- a/tests/api_root.js
+++ b/tests/api_root.js
@@ -92,4 +92,15 @@ tape.test("reflected roots", function(test) {
             test.end();
         });
     });
+    
+    test.test(test.name + " - proto path", function(test) {
+        var root = new Root();
+        test.plan(3);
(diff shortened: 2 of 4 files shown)


yorkie:
Can anyone review this patch, this seems to be good options.

dcodeIO:
Currently, the intended way is to override [Root#resolvePath](https://github.com/dcodeIO/protobuf.js/blob/master/src/root.js#L62) with a custom function. I don't see an issue with adding an option like the one proposed here, though, once the respective option interfaces and derived functions have been updated as well.

yorkie:
Hi @dcodeIO may I ask you give an example, override `resolvePath()`? I don't see any easy way to do this.

dcodeIO:
An example for a common root path instead of inspecting the actual origin is:

```js
var commonOrigin = "...";
myRoot.resolvePath = function(origin, target) {
  util.path.resolve(commonOrigin, target);
}
```

[See also](https://github.com/dcodeIO/protobuf.js/blob/master/lib/path/index.js#L57)

yorkie:
Yes, it's working and thank you, but I still have to say this is not friendly to users.

dcodeIO:
The idea there was to keep it as general as possible because the requirements differ significantly depending whether one would like to customize locations in a browser or under node. How it is now, it also allows regular expressions etc. and it's even possible to implement the CLI's `--path` [option with it](https://github.com/dcodeIO/protobuf.js/blob/master/cli/pbjs.js#L161).

ricardohbin:
Just to explain my motivation to do this patch: we are using a shared proto resources in various applications (and languages). With protoPath, protobufjs can work with .proto like golang - that is the same way that protobuf is designed to work with dependencies.

ricardohbin:
Some news about this one?

jtara1:
Had to read lib source code to make more sense of this
>An example for a common root path instead of inspecting the actual origin is:

> var myCommonOrigin = "...";
> var myRoot = ...;
> myRoot.resolvePath = function(origin, target) {
>   return protobuf.util.path.resolve(myCommonOrigin, target);
> }


To make the example a little more clear, you need to directly create an instance of `protobuf.Root` then use that to call `root.load` instead of `protobuf.load` to get another root instance.

eg (node.js)
```node.js
const fs = require('fs');

(async () => {
  console.log('start');
  
  try {
    const myCommonOrigin = '.';
    const myRoot = new protobuf.Root();
    myRoot.resolvePath = function (origin, target) {
      return protobuf.util.path.resolve(myCommonOrigin, target);
    };

    const schemaPath = 'my-path/schema.proto';
    let root = await myRoot.load(schemaPath);

    // Obtain a message type
    const AwesomeMessage = root.lookupType('mynamespace.MyMessageType');
    const buffer = fs.readFileSync('/home/j/Downloads/my-protobuf-encoded-binary');
    
    // Decode an Uint8Array (browser) or Buffer (node) to a message
    const message = AwesomeMessage.decode(buffer);
    
    // Maybe convert the message back to a plain object
    const object = AwesomeMessage.toObject(message, {
      longs: String,
      enums: String,
      bytes: String,
      // see ConversionOptions
    });

    console.log('done', object);
  } catch (e) {
    console.error(e);
  }
})();
```

PR #853, opened 2017-07-03
buffer: using allocUnsafeSlow instead of allocUnsafe

yorkie:
Take the following example to reproduce the problem:

```js
const buf = proto.encode(text).finish();
console.log(buf.length);  // 10/20/30 or other values, the real length of `text`
console.log(buf.buffer.byteLength); // 8192
```

The background of this example is to use the `ArrayBuffer` instance, `.buffer`. And `Buffer.allocUnsafeSlow()` could fix the problem :)

### DIFF
--- a/src/util/minimal.js
+++ b/src/util/minimal.js
@@ -397,7 +397,7 @@ util._configure = function() {
         function Buffer_from(value, encoding) {
             return new Buffer(value, encoding);
         };
-    util._Buffer_allocUnsafe = Buffer.allocUnsafe ||
+    util._Buffer_allocUnsafe = Buffer.allocUnsafeSlow ||
         /* istanbul ignore next */
         function Buffer_allocUnsafe(size) {
             return new Buffer(size);


dcodeIO:
It is actually intended that a (usually 8k) buffer pool is used because doing so increases allocation performance dramatically. `allocUnsafe` does this automatically under node. [This implementation](https://github.com/dcodeIO/protobuf.js/tree/master/lib/pool) is used in the browser.

yorkie:
But the property `buffer` seems not to be consistent with `Buffer` instance even thought I call slice to get a new `ArrayBuffer` object which length is equal to `Buffer` object.

dcodeIO:
The `buffer` property is a reference to the *underlying* (pooled) ArrayBuffer slab.

```js
var payload = ...; // Uint8Array
var myArrayBuffer = payload.buffer.slice(payload.byteOffset, payload.byteOffset + payload.length);
```

[See also](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice)

yorkie:
It works, thank you @dcodeIO :)

PR #857, opened 2017-07-05
Add methodName to rpc methods to allow to determine methods after code minification

Szpadel:
In current state when static code is minified, generated method names are replaced by single letter names.
This make almost impossible to implement correctly transport layer for services.
Those changes adds new property to methods: `methodName`, that contains string to original method name.
After this change you should determine called method using `methodName` instead of `name`, but old behavior is still preserved.

Also this commit create new exported type: `RpcServiceMethod` to allow to access new property in type safe way.

Runtime generated code was also extended with generated code.
Although methodName is appended in every runtime rpc call - I didn't found a way how to make it one time action in method generation.

Test case was also extended to make sure that `methodName` will containe the same value that `name`.

Bonus:
I fixed Travis build issue for you :-) I don't like to have that red X in my PR ;-)


Szpadel:
@dcodeIO any chance for merge?

examples/streaming-rpc.js
```diff
@@ -78,15 +78,15 @@ function performRequestOverTransportChannel(requestData, callback) {
 // Listen for events:
 
 greeter.on("data", function(response, method) {
-    console.log("data in " + method.name + ":", response.message);
+    console.log("data in " + method.nameName + ":", response.message);
```
avakhrenev: `nameName`?

dcodeIO:
Can't you just do:

```js
function rpcImpl(method, ...) {
  if (method == MyService.prototype.myMethod) { ... }
}
```

Szpadel:
@dcodeIO then you would need rpc transport layer to be dependent on all services.
and you would need implement handling of those messages manually for every service, so at this point it is easier to use pure protobuf serialization and implement all rpc manually.

MHDante:
@dcodeIO This is quite cumbersome when building a transport that works with any proto file.

dcodeIO:
Does [this](https://github.com/dcodeIO/protobuf.js/commit/996b3fa0c598ecc73302bfc39208c44830f07b1a#diff-88ce49885260ab905055158d9e68dd9cR653) solve it? Idea is that defining the `name` property explicitly should survive minification.

avakhrenev:
@dcodeIO how to get a full method name (in format of `${serviceName}/${methodName}` ) at runtime? The `name` property gives access to the method name only, the service name is still missing.

dcodeIO:
The original static-code minification issue has since been addressed by 996b3fa0: generated static service methods now explicitly define their `name` property with the original proto method name, so `rpcImpl` can continue to use `method.name` after minification.

PR #861, opened 2017-07-11
switch to unpkg for CDN urls

graingert:


README.md
```diff
@@ -67,17 +67,15 @@ var protobuf = require("protobufjs");
...
 ```
 
 Production:
 
 ```
-<script src="//cdn.rawgit.com/dcodeIO/protobuf.js/6.X.X/dist/protobuf.min.js"></script>
+<script src="https://unpkg.com/protobufjs@^6.8.0/dist/protobuf.min.js"></script>
 ```
 
-**Remember** to replace the version tag with the exact [release](https://github.com/dcodeIO/protobuf.js/tags) your project depends upon.
```
graingert: no need for this, unpkg automatically handles updates and redirects.

dcodeIO:
Could we make it so that it just mentions both?

graingert:
@dcodeIO not much point, because then you could not get rid of the dist directory.

dcodeIO:
Well, I know that some people might consider this bad practice, but I actually like the simplicitly it provides. Unlike npm/unpkg, you can even use dist files between versions or, if you forked it, from your own repo without publishing anywhere.

PR #862, opened 2017-07-11
including any.proto since it is referenced in type.proto

joakimrapp:
type.proto is importing any.proto, which is missing in master

### DIFF
--- a/google/protobuf/any.proto
+++ b/google/protobuf/any.proto
@@ -1,0 +1,139 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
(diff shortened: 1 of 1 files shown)


joakimrapp:
i did not patch version, needs to be done before deploy :)

joakimrapp:
i did not patch version, needs to be done before deploy :)

joakimrapp:
continuous-integration/travis-ci/pr failed, i do not understand why, travis config problem?

dcodeIO:
The reason why `any.proto` isn't in there is that it is [already bundled](https://github.com/dcodeIO/protobuf.js/blob/master/src/common.js#L43) with the library. Do you have a use case where this isn't working?

dcodeIO:
Closing this issue for now as it hasn't received any replies recently. Feel free to reopen it if necessary!

Issue #856, opened 2017-07-05
The change of "json-module"


yejinjian:
[the change commit](https://github.com/dcodeIO/protobuf.js/commit/4eac28c7d3acefb0af7b82c62cf8d19bf3e7d37b)

why "json-module" default use " protobuf/minimal "?
"protobuf/minimal" did not have function "Root()"

dcodeIO:
Related: https://github.com/dcodeIO/protobuf.js/issues/860

Issue #828, opened 2017-06-10
Has json-module with reflection changed since 6.8?


hnfmr:
protobuf.js version: 6.8

I have been using protobuf.js to generate json-module with reflection information, it's been working fine until 6.7.3 and a recent upgrade to 6.8 is throwing errors.

For example I use the following command to generate json-module:
```js
pbjs --target json-module -w commonjs --keep-case -o dest.js source.proto
```

The errors:

```
$ mocha test/protobuf.spec.js
/Users/xxx/xxx/node_modules/xxx/xxxx/dest.js:6
var $root = ($protobuf.roots["default"] || ($protobuf.roots["default"] = new $protobuf.Root()))
                                                                         ^

TypeError: $protobuf.Root is not a constructor
    at Object.<anonymous> (/Users/xxx/xxx/node_modules/xxx/xxxx/dest.js:6:74)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
```

Thanks!!

dcodeIO:
There have been a couple of changes along the road, but without additional context from your generated file it's hard to tell. `roots` and `Root` should be there. What's the require statement prior to your snippet?

hnfmr:
Wow thanks for the prompt reply.

Here is the header part of the generated `dest.js` using protobuf.js 6.8:

```
/*eslint-disable block-scoped-var, no-redeclare, no-control-regex, no-prototype-builtins*/
"use strict";

var $protobuf = require("protobufjs/minimal");

var $root = ($protobuf.roots["default"] || ($protobuf.roots["default"] = new $protobuf.Root()))
.addJSON({
  pb: {
    nested: {
```

Here is the same header part generated using protobufjs 6.7.3:

```
/*eslint-disable block-scoped-var, no-redeclare, no-control-regex, no-prototype-builtins*/
"use strict";

var $protobuf = require("protobufjs");

var $root = ($protobuf.roots["default"] || ($protobuf.roots["default"] = new $protobuf.Root()))
.addJSON({
  pb: {
    nested: {
```

dcodeIO:
Ah, for some reason this is now including the minimal library, which doesn't include reflection at all (so no `Root`).

Could be related to https://github.com/dcodeIO/protobuf.js/pull/813

hnfmr:
Yep, a comparison made it clear. The `protobufjs/minimal` is probably the cause. Thanks a lot!

hnfmr:
[...]

PR #864, opened 2017-07-11
Fixed dependency for json-module to use "light".

eventualbuddha:
"minimal" doesn't include `Root`, as mentioned in #828 and #856. Probably caused by #813.


dcodeIO:
Thanks!

viktor-hunko:
Please, add changes to npm, v.6.8.1 isn't available. thanks!

PR #870, opened 2017-07-21
Fixed: prevent codegen'ing a function with an invalid name.

eventualbuddha:
This can happen when, for example, an RPC method name is the uppercase of a keyword like "Delete". I tried to retain the performance of the original, but since the test just uses the same names over and over again and they're memoized it's not a super representative set of test data to really understand how the performance changed.

### DIFF
--- a/lib/codegen/index.js
+++ b/lib/codegen/index.js
@@ -1,6 +1,28 @@
 "use strict";
 module.exports = codegen;
 
+var makeValidFunctionName = function() {
+    var memoized = {};
+
+    return function makeValidFunctionName(name) {
+        if ((name in memoized) && memoized[name]) {
+            return memoized[name];
+        }
+
+        try {
+            Function("return function " + name + "(){};");
+            memoized[name] = name;
+        } catch (err) {
+            if (/^[a-z]+$/i.test(name)) {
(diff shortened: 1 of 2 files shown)


eventualbuddha:
Ping.

eventualbuddha:
Is this project unmaintained? Do I need to do something before this can be considered?

dcodeIO:
Thanks, but this doesn't fix the underlying issue since these function might still be called by other parts of the library, using their original name. Instead, we should make sure that invalid names are handled on the reflection level.

dcodeIO:
The commit above should prevent the parse error in runtime code while still naming the method `delete`, hence to be accessed through `myService["delete"](...)`.

eventualbuddha:
Cool, that seems like a good approach. Thanks for looking at this!

eventualbuddha:
@dcodeIO so is your solution finished? Anything I can do to help?

dcodeIO:
The changes are live, latest on npm is 6.8.3. Have you tested with this one?

eventualbuddha:
Oh cool, I’ll try it out tomorrow.  I had assumed that since the issue was still open that the problem remained.

dcodeIO:
Closing for now, feel free to report any remaining issues.

PR #902, opened 2017-09-01
proto3: skip defaults

stelcheck:
Defaults are not supported in proto3, and therefore should
not be appended to a `.proto` file when using `-t proto3`
flag for `pbjs`.

### DIFF
--- a/cli/targets/proto.js
+++ b/cli/targets/proto.js
@@ -238,6 +238,8 @@ function buildFieldOptions(field) {
                     return;
                 break;
             case "default":
+                if (syntax === 3)
+                    return;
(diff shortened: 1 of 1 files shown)


dcodeIO:
Does this actually lead to an error somewhere? Or are `default` options silently skipped anyway?

stelcheck:
```plaintext
protoc test.proto --csharp_out=csharp
test.proto: Explicit default values are not allowed in proto3.
```

And no files are being created. So I assume this is a hard error.

dcodeIO:
Thanks!

stelcheck:
My pleasure!

PR #911, opened 2017-09-14
Allow users to configure additional origin paths for resolving imports.

notbdu:
* Gives users the option of specifying additional origin paths for resolving imports
* Added test case to demonstrate


dcodeIO:
The reason for this API being overly simple is that it is nearly useless in browsers. Hence it has been designed in a friction-less way so everyone can override it with what is required.

notbdu:
Ah, gotcha. Thanks @dcodeIO. 

Closing.

PR #913, opened 2017-09-17
Fix codegen if type name starts with "Object"

### LINKED ISSUE #842: [TypeScript] Invalid codegen if message name starts with Object
protobuf.js version: 6.8.0

When message name starts with Object in every place where this type is returned lowercase version of message name is injected causing errors.

## Repro
Minimal proto file
```
syntax = "proto3";

message ObjectAsd {
}
```

Run following command: `pbjs my.proto -t static-module -w commonjs -es6 -o my.js && pbts my.js -o my.d.ts`

Output my.d.ts file will contain:
```js
public static create(properties?: IObjectAsd): objectAsd;
```
but should contain
```js
public static create(properties?: IObjectAsd): ObjectAsd;
```

AndiDog:
Fixes #842

### DIFF
--- a/cli/lib/tsd-jsdoc/publish.js
+++ b/cli/lib/tsd-jsdoc/publish.js
@@ -295,7 +295,7 @@ function getTypeOf(element) {
     name = name.replace(/\bfunction(?:\(\))?\b/g, "Function");
 
     // Convert plain Object back to just object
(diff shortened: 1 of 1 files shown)


AndiDog:
Build seems to have failed only because of

> No CODECLIMATE_REPO_TOKEN found.

Issue #839, opened 2017-06-21
Well-known types support (Struct, Value)


mkosieradzki:
protobuf.js version: 6.8.0

I am really surprised that in a library targeting Javascript there is no support for the most important wellknown-types: Struct and Value.

Those types have been *specifically* designed to allow the best javascript interop.

Struct is mapping to a generic JSON object
Value is mapping to typescript 'any' value

Is this by-design or is it just an oversight?

dcodeIO:
There is some support for these. For reference: https://github.com/dcodeIO/protobuf.js/blob/master/src/common.js

Still lacking appropriate [wrappers](https://github.com/dcodeIO/protobuf.js/blob/master/src/wrappers.js), though.

mkosieradzki:
@dcodeIO Thanks a lot. Any plans for full support (including wrappers) in a predictable future? Or maybe  is it something up for grabs?

xealot:
This would be tremendous. Is there work planned for this?

If you were able to give some pointers on what needs doing perhaps we could open a PR for this support.

mavrick:
Can we add the protobuf/struct.proto files? would be nice to be able to use them with this lib.

mailaneel:
This is what we have at the moment

StructWrapper.ts
```ts
// tslint:disable-next-line:max-line-length
// @see https://github.com/googleapis/nodejs-common-grpc/blob/67a4cdc109cf3283dbebd487ff672f1fdf3f19bf/src/service.ts

import * as is from 'is';

export class StructEncode {

  seenObjects: Set<{}>;
  removeCircular: boolean;
  stringify?: boolean;

  constructor(options?) {
    // tslint:disable-next-line:no-parameter-reassignment
    options = options || {};

    this.seenObjects = new Set();
    this.removeCircular = options.removeCircular === true;
    this.stringify = options.stringify === true;
  }

  encodeStruct(obj) {
    const convertedObject = {
      fields: {},
    };

    this.seenObjects.add(obj);

    for (const prop in obj) {
      if (obj.hasOwnProperty(prop)) {
        const value = obj[prop];

        if (is.undefined(value)) {
          continue;
        }

        convertedObject.fields[prop] = this.encodeValue(value);
      }
    }

    this.seenObjects.delete(obj);

    return convertedObject;
  }

  encodeValue(value) {
    let convertedValue;

    if (is.null(value)) {
      convertedValue = {
        nullValue: 0,
      };
    } else if (is.number(value)) {
[...]

PR #929, opened 2017-10-13
Customer wrapper for Struct

tdterry:


tdterry:
I didn't mean to create this PR against the upstream repo, but since I did, I'll leave it here for feedback. This PR adds a custom wrapper for the Struct type to convert between plain JS objects and the protobuf Struct type.

jctrouble:
Any chance of getting this landed?

mrene:
Thanks for implementing this @tdterry. I found that this breaks optional objects by not null checking them in the generated code. I fixed it [here](https://github.com/zia-ai/protobuf.js/commit/77a4e77da2ba68ada26715605272814eacdd2e51) (and implemented timestamp in there was well). Are you still working on this? Would be fun to get this into the mainline tree but I'm wondering whether I should start another PR or not.

ntindall:
@mrene @tdterry what is the state of this `feature/struct` branch? It looks like it is about ready to go. What do we need to do to get this merged/published?

This has been an unforseen gotcha in something that we are trying to do...

dcodeIO:
tracked by #839

Issue #946, opened 2017-11-21
 Error: illegal value '{'


ccdunder:
protobuf.js version: 6.8.0 and master

Actual: The following proto fails to parse.
Expected: The following proto parses successfully.
```proto
// test.proto
message Foo {
  option (some.custom_option) = {
    a: { b: "bar" c: "baz" }
  };
  ...
}
```

```
Error: illegal value '{' (test.proto, line 3)
    at illegal (protobufjs/src/parse.js:87:16)
    at readValue (src/protobufjs/src/parse.js:126:19)
    at parseOptionValue (src/protobufjs/src/parse.js:558:59)
    at parseOption (src/protobufjs/src/parse.js:544:9)
    at parseCommon (src/protobufjs/src/parse.js:247:17)
    at parseType_block (src/protobufjs/src/parse.js:298:17)
    at ifBlock (src/protobufjs/src/parse.js:279:17)
    at parseType (src/protobufjs/src/parse.js:297:9)
    at parseCommon (src/protobufjs/src/parse.js:252:17)
    at parse (src/protobufjs/src/parse.js:717:21)
```

dcodeIO:
isn't `a: {` wrong, and it should be just `a {`? At least that's what `parseOptionValue` expects.

ccdunder:
Both are valid & successfully parsed by google/protobuf.

ccdunder:
Which makes sense if you consider something simpler like
```
  option (some.custom_option) = {
    a: 42
  };
```

PR #951, opened 2017-11-27
Support parsing nested option-values with or without ':'

ccdunder:
Issue #946


dcodeIO:
Thanks! I've taken the liberty to fix the remaining issues.

PR #962, opened 2017-12-13
Extend Error Messages

michael-ar:
- several errors i run into fairly often where context would be helpful to debug


dcodeIO:
Looking good, thanks!

PR #968, opened 2018-01-06
Adds a new option to IParseOptions, alternateComment.

gideongoodwin:
If this new flag is true, the tokenizer will change the way it parses comments from the input. Instead of requiring /// or /** */ (docblock style) comments only, it will additionally accept // and /*   */ comments. It will consecutive lines of // comments into one multi-line comment.

The motivation here is that we want to parse comments in customer api descriptions which don't adhere to the docblock style.

### DIFF
--- a/src/tokenize.js
+++ b/src/tokenize.js
@@ -202,7 +202,7 @@ function tokenize(source, alternateCommentMode) {
     function findEndOfLine(cursor) {
         // find end of cursor's line
         var endOffset = cursor;
-        while (endOffset < length && charAt(endOffset) !== '\n') {
+        while (endOffset < length && charAt(endOffset) !== "\n") {
             endOffset++;
         }
         return endOffset;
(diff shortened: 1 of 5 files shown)


src/tokenize.js
```diff
@@ -202,35 +233,93 @@ function tokenize(source) {
...
-                        curr = charAt(offset);
-                    } while (prev !== "*" || curr !== "/");
-                    ++offset;
-                    if (isDoc) /** Comment */
+                }
+                if (!alternateCommentMode) {
+                    // standard comment parsing
+                    if (charAt(offset) === "/") { // Line
+                        // check for triple-slash comment
+                        isDoc = charAt(start = offset + 1) === "/";
```
dcodeIO: What do you think of doing something like

```js
isDoc = (charAt(start = offset) === "/" && ++offset) || alternateCommentMode;
```

here so it always handles `///` and advances properly, and otherwise falls back to `//` in alternateCommentMode? Just asking because it seems that duplicating this section could potentially be avoided.

I'd say it would be fine if `///` in non-alternateCommentMode also coalesces consecutive lines.

gideongoodwin: Thanks for looking. I agree that avoiding duplication would be good, however I am hesitant to change existing behavior in case other consumers of this library depend on it.

In the spirit of your comment, I consolidated the `/*` handling for original and alternate parsing modes. The `//`/`///` handling is a bit too divergent to consolidate further.

dcodeIO:
Looking good, thank you!

PR #975, opened 2018-01-19
Preserve comments when serializing/deserializing with toJSON and fromJSON

gideongoodwin:


dcodeIO:
This looks good to me, but we'd probably need a way to filter out comments in pbjs now. Thinking of an option like `--retainComments` to keep them explicitly. Wdyt?

PR #983, opened 2018-02-06
Preserve comments when serializing/deserializing with toJSON and fromJSON.

gideongoodwin:
This version has an option controlling toJSON behavior. By default we'll skip comments when serializing to JSON, or they can be preserved by passing the new option. Thanks!

### DIFF
--- a/src/mapfield.js
+++ b/src/mapfield.js
@@ -18,9 +18,10 @@ var types   = require("./types"),
  * @param {string} keyType Key type
  * @param {string} type Value type
  * @param {Object.<string,*>} [options] Declared options
+ * @param {string} comment Comment associated with this field
  */
-function MapField(name, id, keyType, type, options) {
-    Field.call(this, name, id, type, options);
+function MapField(name, id, keyType, type, options, comment) {
+    Field.call(this, name, id, type, undefined, undefined, options, comment);
 
     /* istanbul ignore if */
     if (!util.isString(keyType))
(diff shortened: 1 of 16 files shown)


src/mapfield.js
```diff
@@ -18,9 +18,10 @@ var types   = require("./types"),
  * @param {string} keyType Key type
  * @param {string} type Value type
  * @param {Object.<string,*>} [options] Declared options
+ * @param {string} comment Comment associated with this field
  */
-function MapField(name, id, keyType, type, options) {
-    Field.call(this, name, id, type, options);
+function MapField(name, id, keyType, type, options, comment) {
+    Field.call(this, name, id, type, undefined, undefined, options, comment);
```
gideongoodwin: Please confirm, this looks like an old bug calling base class constructor with wrong params.

dcodeIO: I believe omitting `rule` and `extend` and instead passing `options` should be handled [here](https://github.com/dcodeIO/protobuf.js/blob/master/src/field.js#L56). Might well be, though, that specifying `undefined` explicitly is better optimizable by JS engines?

gideongoodwin: Ah I didn't see the parameter shuffling in the Field constructor...I think it's clearer to explicitly pass undefined values for these, what do you think? Since `options` can be undefined here as well, the duck typing in Field() gets a little nondeterministic if these values are omitted.

gideongoodwin: Added additional parameter shuffling to handle the comment parameter.

dcodeIO:
Thanks!

PR #994, opened 2018-03-05
964 enum circular dep react native

LeonardPauli:
React native default setup is unable to handle the circular dependency. Delaying the use solves this (7 char change, but changed to legacy function style to match code style).

See https://github.com/dcodeIO/protobuf.js/issues/964.

### DIFF
--- a/src/enum.js
+++ b/src/enum.js
@@ -5,7 +5,7 @@ module.exports = Enum;
 var ReflectionObject = require("./object");
 ((Enum.prototype = Object.create(ReflectionObject.prototype)).constructor = Enum).className = "Enum";
 
-var Namespace = require("./namespace"),
+var getNamespace = ()=> require("./namespace"),
     util = require("./util");
 
 /**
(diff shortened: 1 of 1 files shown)


src/enum.js
```diff
@@ -5,7 +5,7 @@ module.exports = Enum;
 var ReflectionObject = require("./object");
 ((Enum.prototype = Object.create(ReflectionObject.prototype)).constructor = Enum).className = "Enum";
 
-var Namespace = require("./namespace"),
+var getNamespace = function () { return require("./namespace") },
```
dcodeIO: Isn't there another way, like reordering imports? Asking because there's already a more general [workaround in place](https://github.com/dcodeIO/protobuf.js/blob/master/src/namespace.js#L428) for similar cases that doesn't require calling a function each time, and instead sets these things up on instantiation.

LeonardPauli: I see, neat! See update.

Ps. had some fun with `node --max_inlined_bytecode_size=0 test.js`, not sure if the extra function call would have mattered in execution, though using existing logic is better!

pbaer:
Change looks good, hope this can be merged into master soon.

dcodeIO:
Could you check that master, with my additional commit, is working as intended with React Native?

pbaer:
@dcodeIO the latest version on npmjs is still 6.8.6, when can we expect 6.8.7 with this fix to be published? Thanks.

dcodeIO:
Thanks for the reminder. I had to do some significant changes to dependencies and the build step in the last two commits. Can you verify that master is working for you?

pbaer:
I can confirm that master is in a good state for us (@SlyryD and I are on the same team).

PR #995, opened 2018-03-07
fix usage of global object in browser environment

quanterion:
Browser code should not rely on things that are not available in browser environments. Otherwise it throws error **global is undefined**.

https://github.com/angular/angular-cli/issues/9827#issuecomment-369578814

### DIFF
--- a/lib/prelude.js
+++ b/lib/prelude.js
@@ -13,7 +13,10 @@
     }
 
     // Expose globally
-    var protobuf = global.protobuf = $require(entries[0]);
+    var protobuf = $require(entries[0]);
+    if (typeof global !== "undefined") {
+        global.protobuf = protobuf;
+    }
 
     // Be nice to AMD
     if (typeof define === "function" && define.amd)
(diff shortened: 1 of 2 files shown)


dcodeIO:
The expectation here is that the sources won't be used in something else than node, while the dist files can be used in various environments. Anyway, there are multiple options to solve such issues:

1. If a project runs in the browser, depend on the dist files instead of the sources.
2. We could add a `browser` entry to `package.json` so automate this, but I am not sure that each and every bundler honors this.
3. Feature-check in the node sources, possibly making bundlers with default settings pull in unnecessary and possibly large shims. (That's the issue the linked comment above overlooks.)

quanterion:
It would be great to implement 2nd option. Webpack is the most popular bundler in the wild, I think it should compatible with it. I can try to make a PR if you hint me how to do that. I can't understand how to add browser entry for full and minimal libraries.

dcodeIO:
> I can't understand how to add browser entry for full and minimal libraries.

Good point. Maybe the package should expose the dist files for everything, and just use the sources for building.

quanterion:
Yeah.. As I can see in many modern libraries like RxJS, angular, typescript etc. They all expose dist files only

dcodeIO:
A related fix landed in 7c391d63 and 812b38dd, so this patch is no longer applicable.

PR #999, opened 2018-03-15
Ignore optional commas in aggregate options

### LINKED ISSUE #383: options parser exception parsing multiple options in same code block
A proto file with options like:

```
option (toplevel_3.options) = {
  opt1: true,
  opt2: "Support multiple options in the same block"
};
```

Causes exceptions when parsing: 

```
 Parse error at line 5: illegal option name: (toplevel_3.options).,
    Error: Parse error at line 5: illegal option name: (toplevel_3.options).,
```

drigz:
More information on the "Aggregate Syntax" here:
https://github.com/google/protobuf/issues/1148

These are accepted by the proto compiler but rarely used - I found an
example in the
[Cloud Endpoints docs](https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#google.api.HttpRule.FIELDS.string.google.api.HttpRule.rest_method_name).

Fixes #383.

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -567,6 +567,7 @@ function parse(source, root, options) {
                     else
                         setOption(parent, name + "." + token, readValue(true));
                 }
+                skip(",", true);
             } while (!skip("}", true));
         } else
             setOption(parent, name, readValue(true));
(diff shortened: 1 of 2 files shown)


dcodeIO:
Thanks! :)

PR #1001, opened 2018-03-18
Support Service in Root.fromDescriptor

mightyguava:
I'm working on a server that can dynamically discover and map gRPC services to HTTP using [ServerReflection](https://github.com/grpc/grpc-go/blob/master/reflection/grpc_reflection_v1alpha/reflection.proto). The ReflectionObject parsed from the descriptor returned by the reflection service is missing the service stubs.


dcodeIO:
Thanks! :)

PR #1018, opened 2018-04-14
Get node executable path from process.execPath

kellycampbell:
This is required if pbts is called without node in the path. 

I ran into this issue while creating protobufjs build rules for the bazel build system.


dcodeIO:
Thanks!

PR #1025, opened 2018-04-25
remove return type requirement of pbjs callback

vuryleo:
The callback function of pbjs main should not enforce a return value.


dcodeIO:
Iirc, it *may* return an exit code, [see](https://github.com/dcodeIO/protobuf.js/blob/master/cli/pbjs.js#L20). Not sure why it stated `{}`, hmm.

dcodeIO:
Thanks!

PR #1031, opened 2018-04-27
Use same node when shelling out to jsdoc

alexeagle:
Right now we assume that the node in the process.env['PATH'] will be the right one. If the user uses a certain version of node to invoke pbts (or doesn't have node on their PATH) then this does the wrong thing. We should spawn another copy of the same node binary that's already running.


alexeagle:
Oh, I see that @kellycampbell found this already, this is a dupe of https://github.com/dcodeIO/protobuf.js/pull/1018

PR #1033, opened 2018-04-29
Allow abstract methods

thegecko:
The tsd-jsdoc currently doesn't allow you to create typings for abstract methods.

This means you cannot create typed target templates to inherit from, this PR fixes that.

### DIFF
--- a/cli/lib/tsd-jsdoc/publish.js
+++ b/cli/lib/tsd-jsdoc/publish.js
@@ -645,7 +645,9 @@ function handleFunction(element, parent, isConstructor) {
         insideClass = isClassLike(parent);
         if (insideClass) {
             write(element.access || "public", " ");
-            if (element.scope === "static")
+            if (element.virtual)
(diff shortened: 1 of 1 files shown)


cli/lib/tsd-jsdoc/publish.js
```diff
@@ -645,7 +645,9 @@ function handleFunction(element, parent, isConstructor) {
         insideClass = isClassLike(parent);
         if (insideClass) {
             write(element.access || "public", " ");
-            if (element.scope === "static")
+            if (element.virtual)
+                write("abstract ");
+            else if (element.scope === "static")
```
dcodeIO: Looking good, except that I don't know if jsdoc might mark something as virtual that is also static. Wdyt of checking for static first?

thegecko:
Hmm, should that have failed CI?

thegecko:
Sorry, need to open this PR in a different way.

PR #1037, opened 2018-05-01
Added ability to generate abstract members

thegecko:
Recreation of #1033 

The tsd-jsdoc currently doesn't allow you to create typings for abstract methods.

This means you cannot create typed target templates to inherit from, this PR fixes that.


thegecko:
I don't believe the CI failure is due to changes in this PR, can it be re-run?

thegecko:
bump

Issue #1034, opened 2018-04-29
Unable to add imports for pbts generated types


thegecko:
protobuf.js version: master

When using a custom target for `pbjs` I expect to be able to have generated classes inherit from a custom base class (as opposed to something from the protobufjs codebase, e.g. protobuf.Rpc.Service).

This can be achieved by writing out a require in the template, but when it comes to generating the types with `pbts`, the requires are lost and tsc fails.

I see there is a hardcoded import added by pbts:

https://github.com/dcodeIO/protobuf.js/blob/master/cli/pbts.js#L146

JSDoc doesn't seem to support imports/requires, so is there a way to add custom imports?

dcodeIO:
I think this'd require some sort of postprocessing at this point, but a mechanism that somehow inherits custom imports from JS in the generated TSD would be ideal, probably.

thegecko:
Ultimately I'm trying to create an abstract class for implementing RPC server templates and this is the final hurdle.

I can approach this in different ways:
 - Create a PR containing a new generic target and base class alongside `protobufjs.Rpc.Service`
 - Keep it a custom implementation, which will require the `imports` to be fixed:
    - Fix imports by inventing a new extension to the jsdoc markup (`@import` perhaps) to be processed by `pbts`
    - Fix imports by allowing a list of imports to be passed to the `pbts` tool

I think I'd prefer to see the import syntax included in the target as it keeps the concerns together. Therefore the jsdoc extension approach could be better. I can then share my server work if of interest later.

Any preference?

thegecko:
Perhaps abuse the `@external` tag:

http://usejsdoc.org/tags-external.html

dcodeIO:
Closing this issue because it has been inactive for more than two years and the project has changed substantially.

If this is still relevant, please open a fresh issue with current reproduction steps, version details, and any updated context.

Please also consider that maintaining protobuf.js takes time and effort. If the project is valuable to you or your organization, sponsorship is appreciated.

Thanks!

PR #1038, opened 2018-05-01
Custom pbts type imports

thegecko:
With reference to #1034, this is an example of how extra imports could be added to the generated types using the pbts cli tool.

It's far from an ideal solution, but works.


dcodeIO:
Thanks! Merging and going to take a look :)

PR #1046, opened 2018-05-18
Fix package browser field descriptor

SlyryD:
The [package browser field descriptor](https://github.com/defunctzombie/package-browser-field-spec) expects an entry point, rather than the bundle file path.

This change specifies an entry point for the three different libraries.

Addresses issue https://github.com/dcodeIO/protobuf.js/issues/1045

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
     "typescript"
   ],
   "main": "index.js",
-  "browser": "dist/protobuf.js",
   "types": "index.d.ts",
   "bin": {
     "pbjs": "bin/pbjs",
--- a/package.json
+++ b/package.json
@@ -19,9 +19,9 @@
   ],
   "main": "index.js",
   "browser": {
-    "protobufjs": "./src/index.js",
-    "protobufjs-light": "./src/index-light.js",
-    "protobufjs-minimal": "./src/index-minimal.js"
+    "protobufjs": "./index.js",
+    "protobufjs/light": "./light.js",
+    "protobufjs/minimal": "./minimal.js"
   },
   "types": "index.d.ts",
   "bin": {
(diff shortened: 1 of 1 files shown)


SlyryD:
I verified I was able to produce a browserify bundle with each of the three libraries as a dependency, producing 3 bundles of differing lengths as expected.

package.json
```diff
@@ -18,7 +18,11 @@
     "typescript"
   ],
   "main": "index.js",
-  "browser": "dist/protobuf.js",
+  "browser": {
+    "protobufjs": "./index.js",
```
dcodeIO: Are these aliases actually necessary? From my understanding, as the library is imported from the outside already, having them on the inside shouldn't do anything? For example, there is a `light.js` at the root of the package already that is picked up on `require("protobufjs/light")`.

package.json
```diff
@@ -18,11 +18,7 @@
     "typescript"
   ],
   "main": "index.js",
-  "browser": {
-    "protobufjs": "./index.js",
-    "protobufjs/light": "./light.js",
-    "protobufjs/minimal": "./minimal.js"
-  },
+  "browser": "index.js",
```
dcodeIO: I think the browser entry can simply be removed. As you pointed out, previous usage in the package.json was essentially wrong.

package.json
SlyryD: Removing the browser entry works, and specifying all three works. Specifying only index.js does not work.

package.json
SlyryD: Agreed, I will do that. Thanks!

dcodeIO:
Thanks :)

SlyryD:
Hi Daniel, is there anything else delaying the release of npm package v6.8.7? Thanks!

PR #1050, opened 2018-05-23
Prevent swallowing errors from imported protobufs

benesch:
Previously, when parsing in synchronous mode (e.g., via the pbjs CLI),
syntax errors in imported files would be swallowed. Protobuf compilation
would then proceed with partial type information, resulting in very
confusing error messages. For example, consider the following protobuf
files:

    # a.proto
    import "b.proto";

    message A {
      B b = 1;
    }

    # b.proto
    ;; syntax error ;;

    message B {
    }

Invoking pbjs on a.proto would fail to report the syntax error in
b.proto, and instead fail with an error like

    Error: no such Type or Enum 'B' in type 'A'

because compilation proceeded by swallowing tye syntax error, without
having parsed the entirety of b.


src/root.js
```diff
@@ -85,13 +85,13 @@ Root.prototype.load = function load(filename, options, callback) {
 
     // Finishes loading by calling the callback (exactly once)
     function finish(err, root) {
+        if (sync)
+            throw err;
```
dcodeIO: Doesn't this throw even if there is no error?

benesch: Yep. But that was true before, too. It works because `finish` is always invoked with a non-null error in synchronous mode.

hugebdu:
@dcodeIO Daniel, any chances to merge this?
10x

skeet70:
Would also appreciate this being merged. I'm sharing protobuf files with a scala project that has scalapb annotations and imports a related protobuf file. I just got a bunch of unrelated type errors until I manually made this change in `node_modules` and saw that the `scalapb.proto` was trying to import from the wrong path.

skeet70:
Maybe @JustinBeckwith could take a look?

PR #1065, opened 2018-06-06
Service: add support for RPCs with raw messages

kohlmannj:
**Note:** In an attempt to simplify the code diff, this PR does _not_ contain updated `dist/*` files.

## TODO
- [x] **Help Wanted:** fix failing tests 😅
- [ ] **Write tests for new functionality**
- [ ] Commit updated build assets (i.e. run `npm run build` and commit)

Opening this PR now so I can gather early feedback from the maintainers.

## Description
The Service constructor now accepts a boolean-type `rawMessages` argument, which is `false` by default. This arg is also assigned to a public member, `this.rawMessages`.

The `rpc.Service.rpcCall() ` method now checks `self.rawMessages` to determine whether or not it should:

- **Encode the request data** before making the RPC call
- **Decode the response data** afterwards

That is, if `self.rawMessages === true`, we'll call the RPCImpl with the un-encoded, "raw" `requestData` object, and will expect to receive an un-encoded `response` object.

If `self.rawMessages === false`, we'll encode and decode messages, as in the current implementation.

## Motivation
I’m using Protobuf.js to implement a JavaScript interface for **an existing RPC implementation*** which does not use protocol buffers (it uses serialized JSON instead).

Fortunately, the “protocol” part of Protobuf still offers significant value in this use case. (Seriously, it’s revelatory, and what a great project Protobuf.js is besides!)

Since the existing RPC implementation doesn’t use Protobuf, I’m having `pbjs` generate static modules without `create()`, `encode[Delimited]()`, `decode[Delimited]()`, `verify()`, and `convert()` methods:

```bash
$ pbjs -t static-module --no-create --no-encode --no-decode --no-verify --no-convert --no-delimited ./src/*.proto > ./lib/index.js
```

I'm also generating TypeScript definitions, which covers our needs for message verification.

\* Technically I'll be writing an `RPCImpl` function which wraps the "real" RPC interface. Either way, it expects serialized JSON, not a Protobuf encoded message.

—

In practice, I found that these generated classes, sans `encode[Delimited]()` and `decode[Delimited]()` methods, **did not work with Protobuf.js’s Service class**. `rpc.Service.rpcCall()` assumes that [`requestCtor` has a `encode[Delimited]()` method](https://github.com/dcodeIO/protobuf.js/blob/6.8.6/src/rpc/service.js#L96) and that [`responseCtor` has a `decode[Delimited]()` method](https://github.com/dcodeIO/protobuf.js/blob/6.8.6/src/rpc/service.js#L111). Since my generated message constructors lack both those methods, I get a runtime error when attempting to make an RPC.

## Benefits
By adding the `rawMessages` argument and public field to `Service`, developers can now implement Protobuf without the “buf” part, using Protobuf.js’s excellent feature set to generate Service subclasses and message type definitions.

Furthermore, because `rawMessages` is an additional Service constructor option, and `false` by default, this enhancement is fully backwards compatible.

Finally, I have also overloaded the `RPCImpl` type alias to account for raw messages, so developers still benefit from type checking when defining an RPC implementation in TypeScript.

### `rpc.Service` Implementation

- Add optional, boolean-type `rawMessages` positional argument to constructor
- Add boolean-type `this.rawMessages` public field

### `rpc.Service.rpcCall()` Implementation

- Check `this.rawMessages` before attempting to call `requestCtor.encode[Delimited]()` or  `responseCtor.decode[Delimited]()`
- **Note:** Service instances constructed with the default `rawMessages = false` will still try to call these methods; if they don't exist in this case, we'll throw an Error, as in the current implementation

### `rpc.Service` Type Definitions

- Add `rawMessages` argument and public field to type definitions (along with JSDoc comments)

### `Service extends NamespaceBase` Type Definitions

- Add `rawMessages` arg to `create()` method definition

### `RPCImpl` Type Definition

- Allow `requestData` to be of type `{}` in addition to `Uint8Array` and `null`

### `RPCImplCallback` Type Definition

- Allow `response` to be of type `{}` in addition to `Uint8Array` and `null`

### `pbjs` Static Targets Implementation

- Add `rawMessages` JSDoc comments and arguments to generated code

### `test/data` Type Definitions

- Add `rawMessages` to various class definitions within `test/data/rpc*.d.ts` files

### DIFF
--- a/src/rpc/service.js
+++ b/src/rpc/service.js
@@ -38,8 +38,9 @@ var util = require("../util/minimal");
  * @param {RPCImpl} rpcImpl RPC implementation
  * @param {boolean} [requestDelimited=false] Whether requests are length-delimited
  * @param {boolean} [responseDelimited=false] Whether responses are length-delimited
+ * @param {boolean} [rawMessages=false] Whether to disable message encoding and decoding
  */
-function Service(rpcImpl, requestDelimited, responseDelimited) {
+function Service(rpcImpl, requestDelimited, responseDelimited, rawMessages) {
 
     if (typeof rpcImpl !== "function")
         throw TypeError("rpcImpl must be a function");
@@ -63,6 +64,12 @@ function Service(rpcImpl, requestDelimited, responseDelimited) {
      * @type {boolean}
      */
     this.responseDelimited = Boolean(responseDelimited);
+
+    /**
+     * Whether to disable message encoding and decoding.
+     * @type {boolean}
+     */
+    this.rawMessages = Boolean(rawMessages);
 }
 
 /**
@@ -91,9 +98,12 @@ Service.prototype.rpcCall = function rpcCall(method, requestCtor, responseCtor, 
     }
 
     try {
+        var encodeMethod = requestCtor[self.requestDelimited ? "encodeDelimited" : "encode"];
+        var decodeMethod = responseCtor[self.responseDelimited ? "decodeDelimited" : "decode"];
+
         return self.rpcImpl(
             method,
-            requestCtor[self.requestDelimited ? "encodeDelimited" : "encode"](request).finish(),
+            self.rawMessages ? request : encodeMethod(request).finish(),
             function rpcCallback(err, response) {
 
                 if (err) {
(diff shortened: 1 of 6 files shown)


src/rpc/service.js
```diff
@@ -93,9 +100,11 @@ Service.prototype.rpcCall = function rpcCall(method, requestCtor, responseCtor,
     try {
         return self.rpcImpl(
             method,
-            requestCtor[self.requestDelimited ? "encodeDelimited" : "encode"](request).finish(),
+            self.rawMessages
+                ? request
+                : requestCtor[self.requestDelimited ? "encodeDelimited" : "encode"](request).finish(),
             function rpcCallback(err, response) {
-
+                debugger;
```
ivan-shaban: Looks like it shouldn't be here?

rise0chen:
Why close this PR.
It is useful

PR #1066, opened 2018-06-07
Remove dependency on vulnerable package and fix lint errors

SlyryD:
Updated to gh-pages@^1.2.0, to remove dependency on [vulnerable](https://github.com/brianloveswords/base64url/pull/25) base64url@2.0.0.

Fixed lint errors.

### DIFF
--- a/cli/pbts.js
+++ b/cli/pbts.js
@@ -139,8 +139,8 @@ exports.main = function(args, callback) {
         });
 
         function getImportName(importItem) {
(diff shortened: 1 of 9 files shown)


SlyryD:
Seems like my version of npm is creating pinned versions of modules in the "requires" field of package-lock.json. For example,
`"lodash": "^4.14.0"`
becomes
`"lodash": "4.17.10"`

I'm not certain whether this is consistent with the current version of npm that is being used in the repository.

Issue #850, opened 2017-06-27
handle 'type.googleapis.com/' string


FoghostCn:
protobuf.js version: 6.8.0

in google's any.proto file I find this document:

> The pack methods provided by protobuf library will by default use
 'type.googleapis.com/full.type.name' as the type URL and the unpack
 methods only use the fully qualified type name after the last '/'
 in the type URL, for example "foo.bar.com/x/y.z" will yield type
 name "y.z".

and I have seen your code about `any` type
```js
wrappers[".google.protobuf.Any"] = {
    fromObject: function(object) {
        // unwrap value type if mapped
        if (object && object["@type"]) {
            var type = this.lookup(object["@type"]);
            ........
        }
        return this.fromObject(object);
    },
    toObject: function(message, options) {
        // decode value if requested and unmapped
        if (options && options.json && message.type_url && message.value) {
            var type = this.lookup(message.type_url);
        .......
        return this.toObject(message, options);
    }
};
```
`fromObject` and `toObject` method didn't handle `type.googleapis.com/` string 

actually `Msg.decode(buffer).toJSON()` didn't auto decode `Any` type with `type_url =type.googleapis.com/myMsg` but can auto decode `.myMsg`

did I missing something?

dobryanskyy:
Hi @FoghostCn , did you manage to find out how to make this working?

FoghostCn:
@dobryanskyy I have to override the two method `fromObject` and `toObject` with this
```js
protobufjs.wrappers['.google.protobuf.Any'] = {
    fromObject(object) {
        if (object && object["@type"]) {
            const t = object["@type"];
            const type = this.lookup(t);
            if (type) {
                return this.create({
                    type_url: `type.googleapis.com/${t}`,
                    value: type.encode(type.fromObject(object)).finish()
                });
            }
        }
        return this.fromObject(object);
    },
    toObject(message, options) {
        if (options && options.json && message.type_url && message.value) {
            const type = this.lookup(message.type_url.replace('type.googleapis.com/', ''));
            if (type) {
                message = type.decode(message.value);
            }
        }
        if (!(message instanceof this.ctor) && message instanceof protobufjs.Message) {
            const object = message.$type.toObject(message, options);
            const name = message.$type.fullName;
            object["@type"] = name.charAt(0) === '.' ? name.substr(1) : name;
            return object;
        }
        return this.toObject(message, options);
    }
};
```

dobryanskyy:
great, thanks

[...]

PR #1068, opened 2018-06-08
fix: google.protobuf.Any type_url fixes

nickpestov:
### Background
While using protobuf.js we've come across compatibility issues to do with `google.protobuf.Any`. Namely, when dealing with Any types, according to Google's documentation (referenced below), `types.googleapis.com` should be used as a default prefix within the `type_url`. The `fromObject` function referenced below failed to lookup types that had a prefix in their `@type` when converting from JSON to protobuf.

Addresses #850.

#### From Google's `google.protobuf.Any` docs
```
// The pack methods provided by protobuf library will by default use
// 'type.googleapis.com/full.type.name' as the type URL and the unpack
// methods only use the fully qualified type name after the last '/'
// in the type URL, for example "foo.bar.com/x/y.z" will yield type
// name "y.z".
```

#### And, regarding JSON
```
// The JSON representation of an `Any` value uses the regular
// representation of the deserialized, embedded message, with an
// additional field `@type` which contains the type URL. Example:
//
//     package google.profile;
//     message Person {
//       string first_name = 1;
//       string last_name = 2;
//     }
//
//     {
//       "@type": "type.googleapis.com/google.profile.Person",
//       "firstName": <string>,
//       "lastName": <string>
//     }
//
// If the embedded message type is well-known and has a custom JSON
// representation, that representation will be embedded adding a field
// `value` which holds the custom JSON in addition to the `@type`
// field. Example (for message [google.protobuf.Duration][]):
//
//     {
//       "@type": "type.googleapis.com/google.protobuf.Duration",
//       "value": "1.212s"
//     }
//
```

### In this PR
- In `fromObject`, only use type name after the last '/' when doing a lookup and make sure the `type_url` prefix is preserved. Otherwise, lookups that included `type_url` path prefixes (such as the default `type.googleapis.com` prefix) would fail.
- In `toObject`, use `type.googleapis.com` default prefix if no `type_url` prefix is used.
- Adjusted 2 tests to comply with the above and the below statement from Google's docs:
```
  // * The last segment of the URL's path must represent the fully
  //   qualified name of the type (as in `path/google.protobuf.Duration`).
  //   The name should be in a canonical form (e.g., leading "." is
  //   not accepted).
```


alexander-fenster:
@murgatroid99 I fixed linting problems, this PR is ready to be reviewed now.

PR #1076, opened 2018-06-28
adds wrapper and timestamp support

antonversal:
The PR is a concept but workable. There are luck of tests.
Now to use this types you need to create a message from object like this:
```js
Message.fromObject({ 
  intField: {
    value: 42
  },
  dateField: {
    seconds: 42,
    nanos: 42
  }
})
```

so the PR simplifies usage: 
```js
Message.fromObject({ 
  intField: 42,
  dateField: new Date()
})
```

I need to know if it could be merged? 
If so ping me in a right direction, what I should add and what to fix.

Thanks!

### DIFF
--- a/src/converter.js
+++ b/src/converter.js
@@ -8,6 +8,20 @@ var converter = exports;
 var Enum = require("./enum"),
     util = require("./util");
 
+var WRAPPER_TYPES = [
+  'BytesValue',
+  'BoolValue',
+  'UInt32Value',
+  'Int32Value',
+  'UInt64Value',
+  'Int64Value',
+  'FloatValue',
+  'DoubleValue',
+  'StringValue'
+]
+
+var TIMESTAMP_TYPE = 'Timestamp'
+
 /**
  * Generates a partial value fromObject conveter.
  * @param {Codegen} gen Codegen instance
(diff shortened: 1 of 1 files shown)


mickeyreiss:
This is working for us still. Would love to get this merged.

afterthought:
I have tested this in a [fork](https://github.com/afterthought/protobuf.js) and ended up tweaking it slightly. Maybe our use case is different, but we wanted to be able to round-trip JSON to/from protobuf. Maybe there is a better way to code this, but I just ensured that if the value was a string that 
it implicitly converts to a date first.

          gen
                ("if(d%s===null)", prop)
                ("m%s={}", prop)
                ("else {")
                ("var finalDate = (typeof d%s===\"string\") ? new Date(d%s) : d%s", prop, prop, prop)
                ("m%s={seconds: Math.floor(finalDate.getTime() / 1000),", prop)
                ("nanos: finalDate.getMilliseconds() * 1000000}")
                ("}");

alexander-fenster:
(copying my note from #1495)

Folks, for those who follows this PR and, in general, follows the topic of proper proto3 JSON serialization/deserialization according to the protobuf [spec](https://developers.google.com/protocol-buffers/docs/proto3#json), I wanted to post a quick update here.  For reasons not directly related to anything in this PR, I won't be able to complete it here soon, so I released the proto3 JSON serialization/deserialization code as a separate npm [package](https://www.npmjs.com/package/proto3-json-serializer).

```sh
$ npm install proto3-json-serializer
```

At this moment, we only support instances of `protobuf.Message` which should be enough for most use cases. Feel free to try it and create issues in that repository in case if you need anything to be fixed; contributions are also welcome :)

This separate package will cover the workflows I need at this time; if I have some free time to backport its functionality directly to protobuf.js (e.g. by generating `.fromProto3JSON` / `.toProto3JSON` together with `.fromObject` / `.toObject`, I'll send a separate pull request here, but I cannot promise any ETA on that.

PR #1084, opened 2018-07-14
fix: wrong sourceRoot

orisano:


JustinBeckwith:
Thanks for the PR!  However - I think this is right the way it is.  `index.js` is setup as `main` in the `package.json`, and this (for better or for worse), lives outside of the `src` directory.

Issue #879, opened 2017-07-26
Overwriting basic functions in custom classes


yagop:
protobuf.js version: 6.8.0

Hello, I'm trying to overwriting `toObject` in a custom class, but it seems is not being possible or I'm doing it wrong.

```js
const protobuf = require("protobufjs");

const proto = `
package main;

message Decimal {
  required sint64 mantisa = 1;
  required sint32 exponent = 2;
}

message Message {
  required Decimal a = 1;
  required Decimal b = 2;
}`;

const root = protobuf.parse(proto).root;

const Message = root.lookupType('main.Message');
const Decimal = root.lookupType('main.Decimal');

const DecimalCtor = function (...args) {
  console.log('DecimalCtor constructor called with:', ...args);
};

DecimalCtor.toObject = function (...args) {
  console.log('DecimalCtor.toObject called with:', ...args);
};

DecimalCtor.prototype.toObject = function (...args) {
  console.log('DecimalCtor.prototype.toObject called with:', ...args);
}; 

Decimal.ctor = DecimalCtor;

const msg = Message.create({
  a: Decimal.create({
    mantisa: 1,
    exponent: 0
  }),
  b: Decimal.create({
    mantisa: 1,
    exponent: 0
  })
});

const buffer = Message.encode(msg).finish();
const decoded = Message.decode(buffer);

Message.toObject(decoded);
```
Output:
```
➜  /tmp node proto.js
DecimalCtor constructor called with: { mantisa: 1, exponent: 0 }
DecimalCtor constructor called with: { mantisa: 1, exponent: 0 }
DecimalCtor constructor called with:
DecimalCtor constructor called with:
```

yagop:
:thinking:

ape-casear:
You can overwrite function `toObject` by using `Protobuf.wrappers`
```
const protobuf = require('protobufjs')

protobuf.wrappers['.main.Decimal'] = {
    toObject(message, options) {
        const originOutput = this.toObject(message, options)
        if (message.type === 2) {
            originOutput.content = UserProfile.toObject(UserProfile.decode(Buffer.from(message.content, 'base64')))
        }
        return originOutput
    },
    fromObject(object) {
        console.log('user fromObject')
        return this.toObject(object)
    },
}

const proto = `
package main;

message Decimal {
  required sint64 mantisa = 1;
  required sint32 exponent = 2;
  optional uint32 type = 3;
  optional bytes content = 4;
}

message User {
[...]

PR #1095, opened 2018-07-25
Message.toObject decode bytes to UTF8 String 

yagop:
Using option `{ bytes: 'utf8' }` in `Message.toObject` will decode bytes to String (`Buffer.toString()`). Previos options are preserved.


dcodeIO:
Note that `Buffer` is not available in the browser. Any reason for not using the `string` type?

dcodeIO:
My feeling on this one is that one should use another library for custom encodings that are not part of protobuf itself.

yagop:
I was trying to overwrite `.toObject` like on #879 but I think its not possible, it is?

Gcaufy:
I agree to add an `{ bytes: 'utf8' }` option.

Because in C++, they like to use bytes as string. bytes can be used as `utf-8` string directly.

Gcaufy:
in case someone else still need this:

https://github.com/Gcaufy/protobufjs.shim

PR #1097, opened 2018-08-03
Add more comment test cases, some failing.

jaslong:
The following new test cases were added. "Block comment under
line comment" fails in both normal and alternateCommentMode.
* One-line block comment. PASS
* Block comment under line comment. FAIL
* Line comment under block comment. PASS

### DIFF
--- a/tests/data/comments-alternate-parse.proto
+++ b/tests/data/comments-alternate-parse.proto
@@ -71,3 +71,14 @@ enum Test3 {
 
     FOUR = 4; /// Other value with a comment.
 }
+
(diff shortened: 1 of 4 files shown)


jaslong:
I discovered a bug when some weird comment format is used. The bug manifests when a block comment is immediately under a line comment. For example:

// This should be ignored but it's not.
/* The real comment. */
message M {}

Also, the library simply crashes in this format:
/*
 *The real comment.
 */
message M {}

I don't think I'll have time to fix this, so wanted to share the test cases with you.

PR #1101, opened 2018-08-19
Fixed: ignore optional semicolon in pb.options

shjwudp:
support https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/proto/examplepb/a_bit_of_everything.proto options style


mattinsler:
+1

Issue #957, opened 2017-12-02
Load .pbtxt files using protobuf.js


lutzroeder:

dcodeIO:
There is no support for data in text format, yet. Would make a nice [extension](https://github.com/dcodeIO/protobuf.js/tree/master/ext), though.

sk-:
@lutzroeder any chance on sharing a beta version of your changes. If you share some instructions on how to use it, I could give it a try and report back with any issues.

lutzroeder:
@sk- Have a look at [package.json](https://github.com/lutzroeder/netron/blob/master/package.json) and #1270 and #1112. It would be helpful if you can get some feedback from maintainers what would be needed to get this merged.

sk-:
Thanks @lutzroeder I tested it and works, although when using typescript there are some problems as some of the types are not exported.

This is the error I'm getting:
`error TS2694: Namespace '"node_modules/protobufjs/index"' has no exported member 'TextReaders'`

lutzroeder:
@sk- can you submit a fix? @dcodeIO mentioned keeping this modular is desirable. As mentioned earlier not sure the issue here is more bug reports that complicate integration. If maintainers dropped the ball on this project due to other priorities or aren't working towards integrating the PR this might be a dead end and code needs to move in an entirely different direction. What is missing is feedback from someone that would be able to merge the change on what needs to be done next.

PR #1112, opened 2018-09-22
Static .pbtxt decode support (#957)

lutzroeder:


jackalcooper:
Hi, Lutz. Thanks for your great work on this feature. 
I tried to use branch `pbtxt` to decode but I got `decodeText is not a function`. I'm wondering how can I verify or decode pbtxt with current codebase.

lutzroeder:
@jackalcooper what steps are you using to reproduce this error? [Netron](https://github.com/lutzroeder/netron/blob/master/tools/tf) has a few samples how this is being used.

jackalcooper:
@lutzroeder 
Here is the code:
```
protobuf.load("path_to.proto", function(err, root) {
    var ProtoDef = root.lookupType("ProtoDef");
    var payload = "net: {}";
    var errMsg = ProtoDef.decodetext(payload)
```

jackalcooper:
Do I have to use cli to generate the code?

lutzroeder:
Yes, currently only `--static-module` is supported.

lutzroeder:
@JustinBeckwith, any advise on how to get this PR reviewed and merged?

JustinBeckwith:
I am *very* new to this project, and mostly here helping triage and push buttons :) @dcodeIO, @nicolasnoble and @murgatroid99 are the actual experts.

dcodeIO:
One thing that bothers me about this is that I expect most users to use the library for binary data and I am not sure in general where text format is useful on the web. This makes me question whether adding a new code path that remains unused in 99% of cases, from the minimal build upwards, is worth it. Is there maybe a way to make this an extension, similar to `ext/descriptor`?

lutzroeder:
@dcodeIO Moved the code so `dist` won't be impacted and the text reader in `ext/prototxt` can be loaded separately for browser.  Looks good or are there other recommended changes?

jackalcooper:
@dcodeIO It is more than useful when building web tools for machine learning frameworks.

lutzroeder:
@dcodeIO, @nicolasnoble and @murgatroid99, what are next steps to get this PR reviewed and merged?

julien-c:
I can understand why one wouldn't want to merge this in the main codepath but we should at least make it easy to find and use if needed.

lutzroeder:
@julien-c @dcodeIO's comment has already been addressed and code moved to `ext`. Also, this PR has been replaced with #1270.

julien-c:
Awesome.

zzusunjs:
> Do I have to use cli to generate the code?
Hi, jackalcooper. Did you fix the problem ? Could you please show some code snippets about how to load .pbtxt file with protobuf.js ? 🙏🙏

PR #1122, opened 2018-10-11
Treat second part of a reserved range as exclusive

dflemstr:
Reserved ranges on a namespace level are exclusive in their second element: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto#L119


dcodeIO:
Thanks!

PR #1131, opened 2018-11-02
Add test case where parse, tokenizer fails with empty option

agirton:
I don't have a fix just yet, trying to understand the code first, but wanted to open this PR for any guidance on the correct place to fix this test case. My application is interacting with a service that has an empty option therefore the parser fails. I figured his library should fail gracefully if the user has not set any option.

It seems the best place to fix this would be in the parse.js file in the `parseOption > parseOptionValue` on line 561 and 564.

### DIFF
--- a/tests/comp_option.js
+++ b/tests/comp_option.js
@@ -1,0 +1,25 @@
+var tape = require('tape');
+
+var protobuf = require('..');
+
+var proto =
+  'syntax = "proto3";\
+package myservice;\
+message DoSomethingRequest {\
+}\
+message DoSomethingResponse {\
+}\
+service MyService {\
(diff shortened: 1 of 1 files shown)


alexander-fenster:
Hi @agirton,

This looks like a bug: the `parseOption` function (in src/parse.js:533) expects an option value inside. This needs to be fixed in the option parse logic.

The test you suggested is actually incorrect because the `// Some TODO comment` will eat all the remaining characters of the string. If you remove this comment and just leave the option empty, it will be a valid test for this bug.

agirton:
@alexander-fenster I've updated test and I believe I found a fix, let me know if this looks ok to you.

PR #1140, opened 2018-11-17
 Added a more API-like interface to pbjs and pbts

cromefire:
For easier usage in things like gulp

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ cli/node_modules/
 docs/
 coverage/
 sandbox/
+.idea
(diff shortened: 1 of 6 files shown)


cromefire:
You can now do:
```typescript
import {main as pbjs} from "protobufjs/cli/pbjs";
pbjs({encode: false}, () => {
    // do something
}, {content: "syntax = \"proto3\";", name: "foo.proto"})
```
Same for `pbts`

cromefire:
Also the `main` functions are not affected by this, so there are no breaking changes

cromefire:
Is there anything blocking this from being merged @alexander-fenster ?

alexander-fenster:
@cromefire Just the lack of my free time - apologize for that! I'll go ahead and review/merge later this week. I do use `pbjs` from my code using the existing API (just the command line options as an array) so I appreciate these changes, just need some time to read through the code :)

cromefire:
The best thing is that you don't have to save the file, but you can retrieve it as `Buffer`

PR #1148, opened 2018-12-16
Other: Generate rpc list for custom rpc implementation

xuxihai123:


nicolasnoble:
So, this one comes from a good intention, but probably has the wrong idea to fix the problem.

The grpc package allows users to generate code using protoc, or allows users to dynamically generate code at runtime using protobufjs.

This "rpc list" really seems to be trying to let protobufjs and pbjs be the command line tool for statically generating code for grpc specifically.

The right way to solve this one would be to emulate what protoc does: support external plugins that you can specify from the command line which will allow you to inject any sort of information you want to the output file, instead of hardcoding something specific for grpc. The final change might be much more involved than this, unfortunately.

PR #1152, opened 2018-12-19
Parse repeated fields in custom options

rcfox:


tests/comp_options-textformat.js
```diff
@@ -7,21 +7,28 @@ import \"google/protobuf/descriptor.proto\";\
...
+  repeated string d = 4;\
+  repeated MyOptions e = 5;\
 }\
 extend google.protobuf.FieldOptions {\
   MyOptions my_options = 50000;\
 }\
 message Test {\
   string value = 1 [(my_options) = { a: \"foo\" b: \"bar\" }];\
-  string value2 = 2 [(my_options) = { a: \"foo\" b { c: \"bar\" } }];\
+  string value2 = 2 [(my_options) = { a: \"foo\" c { a: \"bar\" } }];\
```
rcfox: Note: I modified this test slightly so that it would parse correctly with protoc while maintaining the spirit of the original test.

tests/comp_options-textformat.js
```diff
@@ -7,21 +7,28 @@ import \"google/protobuf/descriptor.proto\";\
...
   string value3 = 3 [(my_options) = { a: \"foo\", b: \"bar\" }];\
+  string value4 = 4 [(my_options) = { d: [\"foo\", \"bar\"] }];\
+  string value5 = 5 [(my_options) = { e: [ { d: [\"foo\", \"bar\"] } ] }];\
```
eyalpost: This isn't valid protobuf. array notation [] is not valid here

rcfox: (Sorry, I seem to have missed the notification for this comment when it came in a month ago.)

I don't think that's true. My company's code base uses it. Also, I made sure that this protobuf message was accepted by protoc before submitting the pull request.

I'm fine with dropping the pull request though, as we've decided against dealing with protobufs in Javascript. It's up to you.

murgatroid99:
This array syntax in this PR is not consistent with the protobuf specification. In order to use a repeated options field, you should set the option more than once, to different values. For example:

```protobuf
string value4 = 4 [(my_option) = "a", (my_option) = "b"];
```

PR #1153, opened 2018-12-20
throw when an invalid enum value is passed to fromObject

gilly3:
Addresses dcodeIO/protobuf.js#789


alexander-fenster:
I do understand where it is coming from but I have several concerns about throwing in this place. 

It might be possible that the object this code parses was created using a newer `.proto` file with some new values added to the enum. Adding fields to an enum is not considered a breaking `.proto` change so throwing in this case is probably not what users would expect. This is different from the case described in an issue, where the code throws when someone passes object where something else is expected.

I don't feel comfortable introducing this new behavior since it might just break existing users' code.

@gilly3 Are you OK with these points?

gilly3:
Good point, thank you.  I agree.

Sorry for the delayed response... (holiday weekend here).

Issue #1130, opened 2018-10-31
CamelCase again


VitalyName:
protobuf.js version: 6.8.8

<please describe the expected and actual behavior>

I found problem with camelcase where  _  before digit. If the field name was "aaa_bbb_1_ccc", then after generation it turns out to be aaaBbb_1Ccc. But why is here second "_" (before digit)?

Jhowden:
Just hit this as well. Is this a regression?

alexander-fenster:
Hi @VitalyName and @Jhowden,

The main problem with the requests like that, as discussed in #1154, is that these changes are not just bug fixes, but are actually breaking changes so I feel we cannot do anything like that without bumping a major version; we don't know if any existing code breaks after doing that. I think, previously, the change related to such field names happened when moving from v5 to v6 which was a major version update. So even if it's clearly a bug, we might as well just leave it there for now.

(I'll close this issue since it's essentially a duplicate of #1009 which was filed earlier)

PR #1154, opened 2018-12-23
fix: ensure camelCase implementation is compatible with Protobuf's ToJsonName

### LINKED ISSUE #1009: Camel case conversion bug with numbers in field names
protobuf.js version: 6.8.6

The snake to camel case conversion should remove all underscores from the field name and match the behavior of lodash's `_.camelCase` method. Currently, if there is a protobuf field with the name of `street_1`, protobufjs's camel casing of this field results in `field_1` while lodash results in `field1`. I see lodash being correct here as a camel cased string should have no `_` left in it after conversion.

```js
const _ = require('lodash');
const text = "street_1"

var camelCaseRe = /_([a-z])/g;
const camelCase = function camelCase(str) {
    return str.substring(0, 1)
         + str.substring(1)
               .replace(camelCaseRe, function($0, $1) { return $1.toUpperCase(); });
};
console.log(camelCase(text));
console.log(_.camelCase(text));
```
`camelCase` function from https://github.com/dcodeIO/protobuf.js/blob/c482a5b76fd57769eae4308793e3ff8725264664/src/util.js#L90-L101

While I feel the current behavior is a bug, fixing it could introduce a breaking change. It may be preferable to introduce a new argument to the `load` function so that a custom case transformation could be passed in to be used in lieu of `utils.camelCase`. 

I am happy to submit a PR, but would like to know what the desired resolution would be, to fix the behavior of `utils.camelCase` or allow for custom case transformations via a `load` argument.

eyalpost:
Added a few test cases that did not work well in the previous version
Should fix #1009 and #1130

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -87,17 +87,28 @@ util.ucFirst = function ucFirst(str) {
     return str.charAt(0).toUpperCase() + str.substring(1);
 };
 
-var camelCaseRe = /_([a-z])/g;
-
 /**
  * Converts a string to camel case.
  * @param {string} str String to convert
  * @returns {string} Converted string
  */
 util.camelCase = function camelCase(str) {
-    return str.substring(0, 1)
-         + str.substring(1)
-               .replace(camelCaseRe, function($0, $1) { return $1.toUpperCase(); });
+    var res = "";
+    var upCaseNext = false;
+    var ch;
+
+    for (var i = 0; i < str.length; i++) {
+        ch = str[i];
+        if (ch === "_") {
+            upCaseNext = true;
+        } else if (upCaseNext) {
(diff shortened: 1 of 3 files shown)


nicolasnoble:
This most likely would cause breaking changes in dependents...

murgatroid99:
To expand on that, if any existing user has a field name that matches one of the patterns of a name that transforms differently as a result of this change, their code that references those fields will break after this change.

eyalpost:
Well, currently this creates a mess when the same proto file is exposed by different technologies and the same client tries to use them. 
I think you should leave this open and target a fix for the next major version when a breaking change is ok

PR #1170, opened 2019-02-19
Support fields ids as keys in plain objects

o-samwal:
This PR is an adaptation of the Closure Library's KeyOption provided by [goog.proto2.ObjectSerializer](https://github.com/google/closure-library/blob/master/closure/goog/proto2/objectserializer.js)

It allows to read, produce and verify a plain object using its fields ids as keys.
Unlike Closure Library, by default the names of the fields are used. To use the ids, it is necessary to specify it with an optional boolean parameter.

We use the ids of the fields in our payloads to reduce their size and pseudo-obfuscate the data.
We would like to accomplish the same with dcodeio/protobuf.js.

With the message
```javascript
message MyMessage {
  string foo = 1;
  uint32 bar = 2;
}
```
We can create this plain object
```javascript
{"foo": "value", "bar": 15}
```
Or this one by using `root.lookupType("MyMessage").toObject(obj, {useId: true})` :
```javascript
{"1": "value", "2": 15}
```

We can still convert `{"foo": "value", "bar": 15}` into a valid MyMessage message using 
`root.lookupType("MyMessage").fromObject(obj)`

We can also convert `{"1": "value", "2": 15}` into a valid MyMessage message using 
`root.lookupType("MyMessage").fromObject(obj, true)`.
The fields names are resolved during the conversion.

Finally, we can verify a plain object with ids as keys by using
`root.lookupType("MyMessage").verify({"1": "value", "2": 15}, true)`.

Like before, the message instances continue to use the fields names.
There is no breaking change : if no extra parameter is specified, the conversions continue to use the fields names.


alexander-fenster:
@o-samwal Hi there! 👋 

Can you share the specific use case why you might need to use field IDs as keys in JSON representation of the protobuf objects?

PR #1183, opened 2019-03-14
allow file-level options everywhere in the file

### LINKED ISSUE #1111: Add support for file level option
protobuf.js version: 6.8.8

`Error: illegal token 'option'` is thrown when a file level `option` is specified. Per the spec, some options should be specified at the file level.

https://developers.google.com/protocol-buffers/docs/proto#options

To reproduce, create a proto file with a file level `option` and then use pbjs to try to convert it to json. It will: `Error: illegal token 'option'`

alexander-fenster:
Fixes #1111. According to proto3 syntax defined [here](https://developers.google.com/protocol-buffers/docs/reference/proto3-spec), an option can appear anywhere in the file. Don't break if we see an option after messages.

I also updated `tests/data/test.proto` to have a test coverage for this use case.

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -714,10 +714,6 @@ function parse(source, root, options) {
 
             case "option":
 
-                /* istanbul ignore if */
-                if (!head)
-                    throw illegal(token);
-
                 parseOption(ptr, token);
                 skip(";");
                 break;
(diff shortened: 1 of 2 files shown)


mattinsler:
+1

PR #1189, opened 2019-03-22
Add working rpcImpl with grpc node package

jbpringuey:
After spending a few hours trying to figure out how to connect those 2 great framework, I think it would be a great idea to add that to the doc to save people time in the future. 
Thanks !


shynome:
Thanks for your document, it has help me

PR #1198, opened 2019-04-10
Ensure descriptor conversion can handle default values

### LINKED ISSUE #1196: Descriptors extension fails to take into account default values
I get the following error using `fromDescriptor`:

```
TypeError: Cannot read property 'add' of undefined
    at Function.fromDescriptor (node_modules/protobufjs/ext/descriptor/index.js:219:66)
    at Function.fromDescriptor (node_modules/protobufjs/ext/descriptor/index.js:92:42)
```

Here's the code that throws the exception:

```javascript
            if (descriptor.field[i].hasOwnProperty("oneofIndex")) // eslint-disable-line no-prototype-builtins
                type.oneofsArray[descriptor.field[i].oneofIndex].add(field);
```

The reason it's throwing an exception is that the `oneofIndex` property is present, but is set to its default value (`0`) - ie, it is effectively unset, but the above code fails to take that into account.

I've tried setting `defaults: false` when loading the file, it doesn't seem to help. The problem may be that I'm importing the `descriptor.proto` into a `proto3` syntax file, and so default values are being forced because there's no such thing as unset in `proto3`? It seems like the above code should be updated anyway to handle this in future.

jroper:
Fixes #1196

If the descriptor has been serialized/deserialized by something that set the default values, then an error occurs during reading the descriptor. This fixes it.

### DIFF
--- a/ext/descriptor/index.js
+++ b/ext/descriptor/index.js
@@ -215,7 +215,7 @@ Type.fromDescriptor = function fromDescriptor(descriptor, syntax) {
         for (i = 0; i < descriptor.field.length; ++i) {
             var field = Field.fromDescriptor(descriptor.field[i], syntax);
             type.add(field);
-            if (descriptor.field[i].hasOwnProperty("oneofIndex")) // eslint-disable-line no-prototype-builtins
+            if (descriptor.field[i].oneofIndex)
                 type.oneofsArray[descriptor.field[i].oneofIndex].add(field);
         }
     /* Extension fields */ if (descriptor.extension)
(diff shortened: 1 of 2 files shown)


jroper:
I wasn't sure what to do about tests here - I did add something to the tests for the descriptor, but that isn't run by the main build as far as I can see, and it doesn't even use a test framework. I'd be happy to make it use tape, just let me know what should be done.

dcodeIO:
See https://github.com/protobufjs/protobuf.js/issues/1196#issuecomment-4482913811

PR #1208, opened 2019-04-19
fix: run npm audit fix

JustinBeckwith:
This is a `package-lock.json` only change which was the result of an `npm audit fix`.


JustinBeckwith:
@nicolasnoble @murgatroid99 bump :)

PR #1211, opened 2019-04-24
Move @types to devDependencies

pgherveou:
Hey there, 
Thanks for maintaining this great package.
Not sure this PR makes sense but I just found out that typescript types were marked as dependencies instead of dev dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,11 +51,11 @@
     "@protobufjs/path": "^1.1.2",
     "@protobufjs/pool": "^1.1.0",
     "@protobufjs/utf8": "^1.1.0",
-    "@types/long": "^4.0.0",
-    "@types/node": "^10.1.0",
     "long": "^4.0.0"
   },
   "devDependencies": {
+    "@types/long": "^4.0.0",
+    "@types/node": "^10.1.0",    
     "benchmark": "^2.1.4",
     "browserify": "^16.2.3",
     "browserify-wrap": "^1.0.2",


dcodeIO:
In the past, putting these in devDependencies didn't play out well because these are, somewhat, actually production dependencies: protobuf.js ships its own types which partially re-export interfaces from long and node types, so if protobuf.js was installed as a normal dependency, its shipped types would always be broken unless someone `cd node_modules/protobufjs && npm i`.

pgherveou:
I see, thanks
For your quick response, maybe adding them as peer could give a hint that they are missing, what do you think? If this is a bas idea we can close this PR

dcodeIO:
Anoher thought here is that long.js will soon ship its own types as well (I think it's just not published yet),, which essentially boils down to having them in production builds anyway. Not sure about peer dependencies. How would this look if someone installs just `protobufjs` - would they see a warning that they must resolve manually?

pgherveou:
> would they see a warning that they must resolve manually?

I am not sure about that, it might depend on your editor as well...
I also stumbled upon this article while making a google search on types & peer-deps.

https://medium.com/@KevinBGreene/surviving-the-typescript-ecosystem-working-with-types-and-type-definitions-3539baf26627#4447

It looks like both deps and peer-deps are common practices. 
In our case, the latest would have the advantage of letting users install a more recent version of  `@types/node`  but as you mentioned might it might not be as easy to use as the current solution where all the types are bundled with protobuf.js

PR #1213, opened 2019-04-28
build: be more explicit about tested versions of nodejs

JustinBeckwith:
At a glance, it's hard to understand what versions of node.js this library is being tested against.  This makes things a tad more explicit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,11 @@ language: node_js
 node_js:
   - node
   - lts/*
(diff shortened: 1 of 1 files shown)


dcodeIO:
For a bit of background: The 4.x versions are there because node's Buffer API implements Uint8Array from 4.3.2 upwards iirc. The others are latest `lts` and latest stable `node`.

Issue #1232, opened 2019-05-30
Messages with own of throw  `m.hasOwnProperty is not a function` error


mhamrah:
protobuf.js version: 6.8.8

I have a protobuf message that includes a oneOf:

```
 oneof identifier {
    BySubsidiary by_subsidiary = 4;
    ByCompany by_company = 7;
  }
```

where these one-of message typesare sub-messages of the parent request message:

```
message MyRequest {
  message ByCompany {
    string company_uuid = 1;
  }
 oneof identifier {
    BySubsidiary by_subsidiary = 4;
    ByCompany by_company = 7;
  }
 // ...
}
```
In constructing a payload to make a call to the operation which uses the request message:

```
{
byCompany: {
              companyUuid:"47E5C2A5-5DF6-4639-9891-87EF8A20436C"
            } 
}
```

I get the following error: ```m.hasOwnProperty is not a function```.

The culprit seems to be in the decode method generated by the library. This following block handles the oneOf switch:

```
(function anonymous(Writer,types,util
) {
return function ListEmploymentsRequest$encode(m,w){
  if(!w)
  w=Writer.create()
  if(m.byEmployments!=null&&m.hasOwnProperty("byEmployments"))
  types[0].encode(m.byEmployments,w.uint32(10).fork()).ldelim()
  if(m.byEmployees!=null&&m.hasOwnProperty("byEmployees"))
  types[1].encode(m.byEmployees,w.uint32(18).fork()).ldelim()
  if(m.byPayGroups!=null&&m.hasOwnProperty("byPayGroups"))
  types[2].encode(m.byPayGroups,w.uint32(26).fork()).ldelim()
  if(m.bySubsidiary!=null&&m.hasOwnProperty("bySubsidiary"))
  types[3].encode(m.bySubsidiary,w.uint32(34).fork()).ldelim()
  if(m.pageSize!=null&&m.hasOwnProperty("pageSize"))
  w.uint32(40).int32(m.pageSize)
  if(m.pageToken!=null&&m.hasOwnProperty("pageToken"))
  w.uint32(50).string(m.pageToken)
  if(m.byCompany!=null&&m.hasOwnProperty("byCompany"))
  types[4].encode(m.byCompany,w.uint32(58).fork()).ldelim()
  return w
}
})
```

[...]

PR #1233, opened 2019-05-31
Use Object.hasOwnProperty instead of prototype method

mhamrah:
This PR addresses a situation uncovered with #1232 where the encode method calls `hasOwnProperty` of an object which may not have a prototype. This situation came about when mapping an express request to a grpc call, where a `oneOf` object is a sub-message of a parent object. Using `Object.hasOwnProperty.call` appears to a safer approach which doesn't make assumptions about the underlying object's inheritance.

### DIFF
--- a/src/encoder.js
+++ b/src/encoder.js
@@ -46,7 +46,7 @@ function encoder(mtype) {
         // Map fields
         if (field.map) {
             gen
-    ("if(%s!=null&&m.hasOwnProperty(%j)){", ref, field.name) // !== undefined && !== null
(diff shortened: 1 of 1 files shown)


alexander-fenster:
Thanks @mhamrah!

mbaranovski:
@alexander-fenster  Is this going to be released anytime soon?

alexander-fenster:
@mbaranovski I don't have rights to release it, please ask @dcodeIO who is the owner of this library :)

Issue #716, opened 2017-03-23
cliDependencies and npm dependency


joscha:
protobuf.js version: 6.6.5

When using the CLI version of protobuf.js, some additional dependencies are installed (they are defined without versions in `cliDependencies` of the `package.json`.
There are a few issues with this approach that I hope we can find a solution for:
* The project I am using is using `yarn`, meaning that wherever `pbjs` was installed into does not have `npm` available on CI.
* Locally there is npm available, but quite often the install fails:
```yarn gen-js v0.21.3
$ ./proto/gen.js.sh
installing jsdoc@^3.4.2
installing tmp@0.0.31
installing espree@^3.1.3
installing escodegen@^1.8.1
module.js:472
    throw err;
    ^

Error: Cannot find module 'espree'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/joscha/Development/wala/wala-sodaqone_3/node_modules/protobufjs/cli/targets/static.js:7:18)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
error Command failed with exit code 1.
```
* The dependencies that are installed on-the-fly are not part of any `yarn.lock` or `npm-shrinkwrap.json`, meaning that the installation can 1) not be cached easily 2) not be reproduced easily 3) yarn will blow away any installed packages not part of the lock file any time `yarn install` is run 4) `npm shrinkwrap` will wet itself because there are additional packages installed.

A couple solutions come to mind:
* Make these actual `dependencies` - after all they are used at runtime
* Create a `protobuf.js-cli` package that has a dependency on the `protobuf.js` package and on all the CLI dependencies. That way this package could be lightweight for the people that are not interested in the CLI.

What do you think?

dcodeIO:
I introduced this because I noticed that a few people forked protobuf.js just without the cli to avoid the additional dependencies. The cleanest solution would be a -cli package probably.

joscha:
I'd be happy to open a PR to make that happen - not sure how I would open a PR for the new CLI package though, do you want to create an empty repo for it in your name?

dcodeIO:
[...]

PR #1234, opened 2019-06-01
feat!: move command line tool to a new package named protobufjs-cli

taylorcode:
I ran into issue #716 when trying to integrate pbjs and pbts into my build system. We must mirror our dependencies for reproducible builds.

@dcodeIO recommended that we convert cli/ into a new package. I see that the work here was started, but left in an incomplete state.

This PR attempts to finish decoupling protobufjs-cli from the parent package, so that it can be installed separately and so we can remove the behind the scenes npm install.

I recommend reviewing commit by commit, I tried to keep them as single purpose as possible. Please consider this is my first time contributing to this project.

To test my changes I ran `npm run test` in the parent package. To test the binaries exposed by protobujs-cli package and verify that they do not depend on the parent package, I moved the /cli folder outside of the parent repo and invoked the binaries directly. I confirmed with the following commands:

```pbjs --target static-module --wrap es6 protos/sample.proto```

provided a proto with the content:
```
syntax = "proto3";

package sample_js_init_data;

message TestMessage {
    bool test_bool = 1;
    string test_string = 2;
}
```

It logs generated js.

When output to a file `generated_js/sample.js`, and running the following command:
```pbts generated_js/sample.js```

It logs the generated typings file content.

### CODE REVIEW

**Comment by taylorcode** [comment]
Hi, can I get a review of this :)? thanks

**Comment by alexander-fenster** [comment]
This feels like a breaking change since `pbjs` / `pbts` won't be available just after installing `protobufjs` if I'm reading it right. I don't feel we have an authority to approve this here :) @dcodeIO can you take a look? :)

**Comment by dcodeIO** [comment]
I'd suggest to collect all the breaking changes (iirc there are more currently blocked for similar reasons that are worth merging) and make it a new major release. A bonus might be to first collect everything not breaking into a minor, but certainly implies quite some work, so depends on how viable it is and if someone is willing to pick this up :)

**Comment by taylorcode** [comment]
@dcodeIO I'm happy to do this.

To clarify, are you suggesting:

- make these changes in a backward compatible manner (leaving pbts and pbjs exposed on the top level package)
- Ask you to do a minor release
- Unexpose pbts/pbjs from top level
- Major release

**Comment by dcodeIO** [comment]
Suggestion was to do backward compatible fixes first, leading to a new minor release (6.8.9), and afterwards do anything breaking leading to a new major release (7.0.0, switch to semver going forward). I'm not sure how important doing a minor update first is, though. Might be that it's not worth it.

**Comment by taylorcode** [comment]
Hey @dcodeIO I made one small improvement to my changes to report an error if protobufjs can't be resolved from the protobufjs-cli package.

I looked over the code again and I don't think that it makes sense to do a minor release because all of the changes are related to separating out the protobufjs-cli package, and the non-breaking changes do not add any value by themselves.

The non-breaking changes include:
- resolving protobufjs from the parent directory or from the node_modules folder
- adding the package.json file to /cli and removing package-standalone.json

If this sounds reasonable to you, how can we proceed? Should I increment the version number to 7.0.0 for both protobufjs and protobufjs-cli?

**Comment by dcodeIO** [comment]
> If this sounds reasonable to you, how can we proceed? Should I increment the version number to 7.0.0 for both protobufjs and protobufjs-cli?

A few questions:

* Is `cli/resolve-protobufjs.js` the recommended way of resolving peer deps or did you come up with it? In the latter case, is there maybe a recommended pattern?
* Some changes to package-lock's might be unnecessary, for example lib/eventemitter doesn't really need one. Should be maybe limit package-lock's in source control to those packages with external non-dev dependencies?

Overall this PR looks good to me, but I didn't test it locally. Would be great if multiple people could verify that it works without issues. Incrementing the version here makes sense to me as well, even if we might collect additional PRs before doing an actual release.

**Comment by taylorcode** [comment]
> Is cli/resolve-protobufjs.js the recommended way of resolving peer deps or did you come up with it? In the latter case, is there maybe a recommended pattern?

In my opinion the best way to solve this problem is to use [Lerna](https://github.com/lerna/lerna) or [Yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) for developing multiple packages in the same repo. If we used Lerna we would be able to reference protobufjs by the module name "protobufjs" for both modes (local development and installed as an npm package). I've used Lerna for projects in the past and it works well.

I'm happy to follow up after this PR and see what it would take to move this repo to Lerna.

> Some changes to package-lock's might be unnecessary, for example lib/eventemitter doesn't really need one.

yep, nice catch. The package-lock.json for eventemitter was added by accident, I must have run `npm i` inside of that folder.

Why do you feel that the package-lock.json is unnecessary? If we decided to use Lerna, we would have a package-lock for each subpackage, and lerna would manage updating them.

> Would be great if multiple people could verify that it works without issues.

I will post my manual test plan here. Any automated tests you recommend I am happy to add as well.

> Incrementing the version here makes sense to me as well, even if we might collect additional PRs before doing an actual release.

I updated it to 7.0.0

**Comment by taylorcode** [comment]
To test my changes:

I published `protobufjs-taylorm` and `protobufjs-cli-taylorm` as npm packages, created a new folder with the following package.json:
```
{
  "name": "protobufjs_fork_tests",
  "version": "1.0.0",,
  "scripts": {
    "pbjs_test": "pbjs --target static-module --wrap es6 sample.proto > sample.js",
    "pbts_test": "pbts sample.js"
  },
  "dependencies": {
    "protobufjs-cli-taylorm": "^7.0.4",
    "protobufjs-taylorm": "^7.0.4"
  }
}
```

And sample.proto:
```
syntax = "proto3";

package sample_js_init_data;

message TestMessage {
    bool test_bool = 1;
    string test_string = 2;
}
```

I then ran `npm run pbjs_test` and `npm run pbts_test`. The first generated sample.json, the second output the .d.ts file content to my console.

To test inside of the repo, I added sample.proto inside of cli/, and invoked the binaries directly with the same arguments.

**Comment by taylorcode** [comment]
Hi @alexander-fenster, @dcodeIO, can I have your review? thanks

File: cli/require-protobufjs.js [hunk 1]
```diff
@@ -0,0 +1,19 @@
+"use strict";
+
+module.exports = function requireProtobufjs() {
+    try {
+        // local development, i.e. forked from github
+        require.resolve("..");
```
alexander-fenster: Can you educate me why do you need these two no-op calls to `require.resolve()`? My understanding is that they would've made sense to check if the module is available, but in both cases the regular `require` follows - and will fail if there is no module found.

alexander-fenster: (is the behavior of `require.resolve` different from the one of `require`? I was under impression that `require` would throw the same `Error` if the requested module is not found)

cli/package.json
```diff
@@ -1 +1,35 @@
-{"version": "6.7.0"}
\ No newline at end of file
+{
+  "name": "protobufjs-cli",
+  "description": "Translates between file formats and generates static code as well as TypeScript definitions.",
+  "version": "7.0.0",
```
alexander-fenster: Just wanted to confirm that we want to start this separate package from v7.0.0 - it's up to @dcodeIO to decide.

cli/package.json
```diff
@@ -1 +1,35 @@
...
+  "types": "index.d.ts",
+  "bin": {
+    "pbjs": "bin/pbjs",
+    "pbts": "bin/pbts"
+  },
+  "scripts": {
+    "prepublish": "node scripts/prepublish"
+  },
+  "peerDependencies": {
+    "protobufjs": "~6.7.0"
```
alexander-fenster: `"^7.0.0"`? (just to avoid having two sets of `pbjs` / `pbts` from different packages)

cli/require-protobufjs.js
taylorcode: you are correct, require.resolve is unnecessary. I removed those calls now

cli/package.json
taylorcode: I've updated this to ^7.0.0

cli/package.json
taylorcode: hey @dcodeIO can you advise here so that we can get this merged, if all else looks good. Thanks

taylorcode:
Hey @dcodeIO and @alexander-fenster, my team has another improvement to protobuf.js that is based on this fork.

Should we wait to open a new PR until this is merged in? Thanks

cli/require-protobufjs.js
```diff
@@ -0,0 +1,17 @@
+"use strict";
+
+module.exports = function requireProtobufjs() {
+    try {
+        // local development, i.e. forked from github
+        return require("..");
+    } catch (e) {
+        // installed as a peer dependency
+        try {
+            return require("protobufjs-taylorm");
```
alexander-fenster: Seems like `-taylorm` is not supposed to be here (and below). Looks like your test package name made its way here :)

alexander-fenster:
@taylorcode I cannot create new release (I work for a Google team that works on gRPC, client libraries, and protobuf, so we asked for write access here to help with stuff, but I cannot and don't want to make releases :) ) - it's all up to Daniel @dcodeIO as for when to merge and release.

Having that said - as a user of `pbjs`/`pbts`, let me try playing with this locally to verify all this stuff works as expected. I'll write back when I check it.

cli/require-protobufjs.js
taylorcode: oh crap. fixing that now

taylorcode: fixed

taylorcode:
sounds good thanks @alexander-fenster. I'll wait for @dcodeIO, and just have my teammate open a new PR after this one lands.

taylorcode:
hey folks... can I have someone review this?, it's been more than 3 months since I opened this PR.

taylorcode:
Hi @dcodeIO reminder on this PR. I’m happy to make any changes.

My team has a few other fixes that we are working on and would like to merge in instead of maintaining our own fork.

gcangussu:
@dcodeIO We are also having problems that would be fixed by this. We use Bazel to get deterministic builds, so installing things from the network without the package-lock on build time is a very big problem for determinism.

Currently we are using [protobufjs-cli-taylorm](https://www.npmjs.com/package/protobufjs-cli-taylorm) without any problem on our project.

gjasny:
Is there a straight-forward way to swap in a fixed fork into things like grpc where protobuf.js is a transitive dependency?

alexander-fenster:
@gjasny We (at Google / gRPC / Google Cloud client libraries) will make sure we release a new version. It will take a while to make it work though, but we're on it. No ETA right now, but we do need to keep it rolling as much as you do - our libraries depend on protobuf.js and we need these unreleased features as well :)

taylorcode:
That's great to hear @alexander-fenster!

In the meantime, we've decided to fork this repo - https://github.com/dropbox/protobuf.js

We also had to fix https://github.com/protobufjs/protobuf.js/issues/1249 because we have many proto files with shared dependencies, and the fact that pbjs only produces one bundle lead to an explosion of duplicated dependencies.

On the no-bundle branch:
https://github.com/dropbox/protobuf.js/tree/no-bundle

Commits:
https://github.com/dropbox/protobuf.js/commit/cfb0048ba3787a99175ed2ef33ea9098e6dc4f8f
https://github.com/dropbox/protobuf.js/commit/c5f0c0b4df2a7696a0cbdd31fa191973d6806fc7

I'm wondering, at Google is there any plan to consolidate this library https://github.com/grpc/grpc-web?

We want to use grpc-web here at Dropbox, but would prefer to use protobufjs because it is much leaner and generates much smaller bundles. I found some discussion on the idea of using grpc-web for making requests to the server here:
 https://github.com/grpc/grpc-web/issues/80#issuecomment-426517834

gcangussu:
@taylorcode is there a package published at NPM with both the cli split and the `no-bundle` option? I tried searching for "dropbox protobufjs", but got nothing.
We are also having the duplicated code issue with the way protobufjs generates code.

staceysern:
We've moved our official fork to https://github.com/dropbox/protobuf.js and we've published to protobufjs-dbx and protobufjs-cli-dbx.

jtcorbett:
@dcodeIO Friendly ping on this. This issue if affecting our team as well.

alexander-fenster:
@jtcorbett Scheduling this to our next triage meeting.

cli/package.json
```diff
@@ -1 +1,35 @@
...
+  "types": "index.d.ts",
+  "bin": {
+    "pbjs": "bin/pbjs",
+    "pbts": "bin/pbts"
+  },
+  "scripts": {
+    "prepublish": "node scripts/prepublish"
+  },
+  "peerDependencies": {
+    "protobufjs": "^7.0.0"
```
murgatroid99, line 21: This doesn't make sense as a peer dependency. It is really a direct dependency: `pbjs` and `pbts` directly load `protobufjs`. As of recent versions of npm, [peer dependencies are not installed automatically](https://docs.npmjs.com/files/package.json#peerdependencies), so this will cause failures if people try to use `protobufjs-cli` as a standalone tool without also having their own dependency on `protobufjs`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "protobufjs",
-  "version": "6.8.8",
+  "version": "7.0.0",
```
murgatroid99, line 3: @alexander-fenster Now that the release-please bot is handling version bumps, how should this be handled?

murgatroid99:
After other issues are addressed, this PR will need to be updated from master again to fix the merge conflicts.

alexander-fenster:
@bcoe After this is merged, the repository will effectively become a monorepo. Will release-please properly support it, or do we need to write a separate releaser for this?

staceysern:
I mentioned it above in the comments but just want to reiterate that Dropbox moved its official fork to dropbox/protobuf.js. There's a cleaner, more complete branch of this at https://github.com/dropbox/protobuf.js/tree/decouple_cli_package which should probably be used instead of the @taylorcode fork.

taylorcode:
Well... it looks like either I messed up, or Github has become confused. I attempted to get taylorcode:cli_package even with dropbox:decouple_cli_package so that I wouldn't need to open a new PR. But it looks like this failed, as Github thinks I force merged my changes into the protobufjs repo. Oddly it also updated the title of this PR to the last commit message in dropbox:decouple_cli_package.

Here is the new PR:
https://github.com/protobufjs/protobuf.js/pull/1459

If someone knows how to fix this (re-open and update the original PR) let me know.

taylorcode:
Ok folks, this PR is now on top of current master, but it looks like we have a build problem. I've moved some dependencies to cli/package.json that are only used inside of the new protobufjs-cli package. However it looks like the CI build only installs dependencies in the parent package. Fwiw I've confirmed locally that `npm run build` and `npm run test` are working if you `cd cli && npm install`.

alexander-fenster:
Thanks @taylorcode, I'll try to figure out what's going on with the CI here.

gcangussu:
@taylorcode @alexander-fenster @staceysern I've took the liberty of fixing the error we were having here: https://github.com/protobufjs/protobuf.js/pull/1480 (don't merge please)

I have opened a PR on @taylorcode's fork: https://github.com/taylorcode/protobuf.js/pull/3. He can rebase his branch with `ProtoBuf7` and then merge it. Then it should start working here.

bcoe:
Hey @taylorcode, just a gentle nudge.

We'd love to move towards the next major release of _protobuf.js_, and we see splitting out the CLI as its own module as one of the significant changes we'd like to make.

Any chance you could work with @gcangussu to merge his fix for tests, and then we can get this over the finish line.

gcangussu:
@bcoe Would an admin be capable of editing the _from_ branch on this PR to `gcangussu:cli_package`? Then we can merge without losing the commits and this PR history. The resulting branch @taylorcode could do by doing my suggestion would be the same as mine (`gcangussu:cli_package`, rebased @taylorcode branch to `ProtoBuf7` and added commits to fix issues).

bcoe:
> Would an admin be capable of editing the from branch on this PR 

@gcangussu (CC: @alexander-fenster), I don't believe we can edit the branch, but we can make a new branch and cherry-pick the changes -- this seems like a good path forward.

gcangussu:
I'm fine with that, let me know if I can help with anything.

taylorcode:
pardon my delay folks. I rebased in @gcangussu's fixes.. I think it worked.

pauldraper:
Checks pass...are there are outstanding issues?

alexander-fenster:
@pauldraper We're working on an automation update that will release this package when it becomes a monorepo (after this one is merged). Please give us a little bit more time and it will go live!  (cc: @bcoe)

package.json
alexander-fenster: release-please now supports monorepos.

Cc: @bcoe

bcoe (approved):

alejandroclaro:
Any news about the release date?

Issue #1210, opened 2019-04-23
API documentation website 404


kaiwk:
It seems one can't access this website:

https://dcode.io/protobuf.js

dcodeIO:
Ah, right, that changed when moving the repo to its own org. New link is: https://protobufjs.github.io/protobuf.js/

kaiwk:
I found the link in the wiki page. Thanks, problem solved.

T3sT3ro:
the wiki link for API documentation V6 point to 404: https://github.com/protobufjs/protobuf.js/wiki  
so does the link on https://www.npmjs.com/package/protobufjs#additional-documentation

PR #1235, opened 2019-06-02
docs: fix stale links to API documentation

jlisee:
As mentioned in #1210 the website changed but there were two places
still referring to the old site.


JustinBeckwith:
Thank you so much for the fix!

PR #1236, opened 2019-06-02
Fix links to ReflectionObject

jlisee:
Change `ReflectionObjecthtml` -> `ReflectionObject.html`

### DIFF
--- a/Namespace.html
+++ b/Namespace.html
@@ -2223,7 +2223,7 @@
             
 <dt>
     <div class="nameContainer">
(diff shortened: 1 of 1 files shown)


dcodeIO:
This file is generated, so any changes would be lost again on running jsdoc the next time. But there must be a way to fix the generator.

jlisee:
> This file is generated, so any changes would be lost again on running jsdoc the next time. But there must be a way to fix the generator.

I was afraid of that, but I was thinking to myself: how could a generator drop a random `.`, but I guess it can.

dcodeIO:
Related: https://github.com/dcodeIO/jaguarjs-jsdoc/pull/1

PR #1265, opened 2019-07-14
parsing from the protobuf json

dilipvamsi:
Added a field `standard` in json options for parsing.


Faendrek:
Hi there, I take it this is to solve the problem with to/from Object in the JS generated files. I can't seem to get your PR version changes to be reflected in the UI when I compile my project's Protobuf files using your protojs, is there anything special / flags I have to pass to the compile command? I'm running `pbjs -t static-module -o output.js my_proto.proto`?

PR #1270, opened 2019-07-28
Static .pbtxt decode support (#957)

lutzroeder:
@dcodeIO, @nicolasnoble and @murgatroid99 what is needed to get this reviewed and merged?

### DIFF
--- a/cli/pbjs.js
+++ b/cli/pbjs.js
@@ -44,7 +44,7 @@ exports.main = function main(args, callback) {
             "force-message": "strict-message"
         },
         string: [ "target", "out", "path", "wrap", "dependency", "root", "lint" ],
-        boolean: [ "create", "encode", "decode", "verify", "convert", "delimited", "beautify", "comments", "es6", "sparse", "keep-case", "force-long", "force-number", "force-enum-string", "force-message" ],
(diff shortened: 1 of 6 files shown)


lutzroeder:
@dcodeIO, @nicolasnoble and @murgatroid99 what is needed to this reviewed and merged?

lutzroeder:
@alexander-fenster can you help with what are the next steps to get this merged? Is there any way I can help? This change has been used for quite a while now and I would like to avoid creating another npm package.

lutzroeder:
@alexander-fenster can you help with what are the next steps to get this merged? Is there any way I can help? This change has been used for quite a while now and I would like to avoid creating another npm package.

alexander-fenster:
@lutzroeder Thank you! We try to review and merge pending stuff on Fridays to have this moving, even though I must admit it moves much slower than I would want :) (it's not a primary job for neither of us here at Google)  I'll add this to the list for the next Friday. Due to the size of code to be reviewed, it might take some time though, but we're really doing our best here.

As a feedback of what can be added - my first ask would be to add tests. At least a test that reads/writes some nontrivial proto object from/to `.pbtxt` would've made the reviewers more comfortable using this. You can try to adapt any test that reads/writes `.json` format and maybe put your `.pbtxt` into `test/data/` folder to use from a test.

Issue #677, opened 2017-02-12
Automatically wrapping & unwrapping wrappers.proto types?


RodH257:
protobuf.js version: 6.6.3

Hey, first off this library is really great - love the typescript integration! Thanks for building it. 

I'm trying to work out how to retain 'null' values across the wire. We have some instances where a value can be null and it means something different to empty string or 0 value. Google has the wrappers https://github.com/google/protobuf/blob/master/src/google/protobuf/wrappers.proto that let you do this, and protobuf.js handles retaining the null well, however it doesn't automatically wrap and unwrap the values

Lets say I have
```
message ChangeEvent {
  string source = 1;
  google.protobuf.StringValue code = 2;
}
```
I'd like to be able to pass in:
```js
const changeEventWithCode = {
  source = 'test',
  code = 'code',
}

const changeEventWithoutCode = {
  source = 'test',
  code = null,
}
```

and have them both encode & decode to the same thing. However it seems if I want to set the code string, I have to do:
```
const changeEventWithCode = {
  source = 'test',
  code = {
    value: 'code',
  },
}
```

I was hoping fromObject and toObject would handle this, but they don't - is there any way I can hook in some customisation to do this, or would this be a good feature to add to the framework?

dcodeIO:
There is no special handling for Wrappers, Any, Duration, Timestamp etc. currently.

What could be done, though, is to set up [custom classes](https://github.com/dcodeIO/protobuf.js#using-custom-classes) for these types that are capable of handling these cases. Haven't thought this through, yet, unfortunately.

RodH257:
I was thinking along those lines as well, I was having a bit of difficulty working out how to get the custom classes injected in when using codegen, but I'll have another crack

dcodeIO:
This adds (hidden) options to specify custom implementations of `fromObject` and `toObject`. Basically, when there is an option `__fromObject`, it is used instead of the generated converter with `this.fromObject` referencing the original function. Likewise, a `__toObject` option does the same but for `toObject`.

There is a commented out example here: https://github.com/dcodeIO/protobuf.js/commit/0c6e639d08fdf9be12677bf678563ea631bafb2c#diff-2c36c683dff5798fefed561dbec17803R54

Additional functionality can now be implemented right within [src/common.js](https://github.com/dcodeIO/protobuf.js/blob/master/src/common.js) where these types are provided.

dcodeIO:
[...]

PR #1271, opened 2019-07-31
Skip object check for types with a defined custom wrapper.

alec-mccormick:
This PR would give the ability to pass a primitive value for a non primitive type as long as a custom wrapper is defined for it.

# Use Case
We would like to be able to define fields that can be multiple different primitive types, however the library currently will throw an error if a message field resolves to a type and the value is not an object. If a custom wrapper is defined for that type, then it should be assumed that the wrapper will handle serialization.

Message:
```protobuf
message StringOrDouble {
  oneof kind {
    string stringValue = 1;
    double doubleValue = 2;
  }  
}

message MyMessage {
  StringOrDouble value = 1;
}
```

Protobuf Wrapper:
```javascript
{
  '.StringOrDouble': {
    fromObject(value) {
      return (typeof value === 'string') ? this.create({ stringValue: value }) : this.create({ doubleValue: value });
    },
    toObject(message) {
      return (typeof message.stringValue === 'string') ? message.stringValue : message.doubleValue;
    }
  }
}
```

### DIFF
--- a/src/converter.js
+++ b/src/converter.js
@@ -6,7 +6,8 @@
 var converter = exports;
 
 var Enum = require("./enum"),
-    util = require("./util");
+    util = require("./util"),
+    wrappers = require("./wrappers");
 
 /**
  * Generates a partial value fromObject conveter.
@@ -32,7 +33,9 @@ function genValuePartial_fromObject(gen, field, fieldIndex, prop) {
                     ("break");
             } gen
             ("}");
-        } else gen
+        } else if (wrappers[field.resolvedType.fullName]) gen
+            ("m%s=types[%i].fromObject(d%s)", prop, fieldIndex, prop);
+          else gen
             ("if(typeof d%s!==\"object\")", prop)
                 ("throw TypeError(%j)", field.fullName + ": object expected")
             ("m%s=types[%i].fromObject(d%s)", prop, fieldIndex, prop);


pgherveou:
@alexander-fenster, @JustinBeckwith,  do you think you guys or another contributors could review and merge this?

pgherveou:
@nicolasnoble any chance you could look at this PR? It is a quick fix and solves some interesting issues. see my comment here see https://github.com/protobufjs/protobuf.js/issues/677#issuecomment-51799720

dardanos:
Please merge this 🙏

betmix-matt:
@dcodeIO or @alexander-fenster any chance this could get some love?  

It's a dependency of work arounds like:
https://github.com/protobufjs/protobuf.js/issues/677

Sandeep-Hirani:
@dcodeIO @alexander-fenster @bcoe Review please

stephenh:
I know pings are annoying @dcodeIO / @alexander-fenster  / @gitjuba but FWIW I'm the maintainer of ts-proto, and have pre-reviewed / :+1: this super-small patch.

We'd like to see this merged b/c it allows us to support modeling `StringValue`/etc. in TypeScript messages as just `string | undefined`.

Currently we already support this for our own encoder methods, but a number of our users use NestJS, which goes through grpcjs to @grpcjs/proto-loader, and proto-loader really wants to go through protobufjs's `converter.js`-generated encoders.

Usually we can use protobufjs's `wrappers` config to interject our own `fromObject` / `toObject` to map our "idiomatic TS"-isms back to what protobufjs supports.

Except for non-object primitives, due to this one extra-cautious `typeof ... !== object` check that this PR removes, _only_ if the type has a `wrapper` configured, i.e. an indication that the user knows what they are doing, and hence the `type ... !== object` check imo should be safe to remove, and this PR safe to merge.

Lmk if you have questions or concerns. Thanks!

Marko298:
We really need this change

zamarawka:
Any way to get it merged?

hyunjinjeong:
This simple PR hasn't been merged or even reviewed for 5 years now 😢

c76a15868f38 (approved):

marketguru-dev (approved):

PR #1276, opened 2019-08-10
add enum value description comments to pbts

rocwind:
brings the jsdoc enum value description to the typescript definition file


initial-wu:
@rocwind it looks like that you need handle some conflicts ?

PR #1286, opened 2019-08-25
Release 5.0.4 with fix for enums with reserved fields

### LINKED ISSUE #1284: Ignore reserved fields for enums
protobuf.js version: 5.0.3

<please describe the expected and actual behavior>
We have a proto message which uses the **reserved** keyword in an enum:

```
enum HeaderType {
    // Prevent removed tag re-use
    // Uncomment after fabric-baseimage moves to 3.5.1
    // reserved 7;
    // reserved "PEER_RESOURCE_UPDATE";
    reserved 7, 9;
    reserved "PEER_RESOURCE_UPDATE", "TOKEN_TRANSACTION";

    MESSAGE = 0;                     // Used for messages which are signed but opaque
    CONFIG = 1;                      // Used for messages which express the channel config
    CONFIG_UPDATE = 2;               // Used for transactions which update the channel config
    ENDORSER_TRANSACTION = 3;        // Used by the SDK to submit endorser based transactions
    ORDERER_TRANSACTION = 4;         // Used internally by the orderer for management
    DELIVER_SEEK_INFO = 5;           // Used as the type for Envelope messages submitted to instruct the Deliver API to seek
    CHAINCODE_PACKAGE = 6;           // Used for packaging chaincode artifacts for install
    PEER_ADMIN_OPERATION = 8;        // Used for invoking an administrative operation on a peer
}
```
While ProtoBuf5 does not actually support reserved fields, the parser ignores it for other token types.

ProtoBuf6 actually supports parsing reserved fields and fixed the issue for enum in #950 .

It's not easy for us to move to the latest release, so we'd like to push a fix for ProtoBuf5

```js
<please provide a code snippet for reproduction>
```

```
<please paste the stack trace of the error if applicable>
```

mastersingh24:
Fixes #1284 and then updates for a 5.0.4 release


mastersingh24:
Did not see any other PRs for ProtoBuf5, so took the liberty of pushing a fix as well as updating for a 5.0.4 release.

mastersingh24:
@nicolasnoble   (since I know you from grpc-node) ... any interest in merging something like this on a back-level release?

Issue #1322, opened 2019-11-21
parser: failing on empty statements (extra semicolons)


rodaine:
protobuf.js version: 

```sh
→ npm version
{
  protobufjs: '6.8.8',
  npm: '6.12.0',
  ares: '1.15.0',
  brotli: '1.0.7',
  cldr: '35.1',
  http_parser: '2.8.0',
  icu: '64.2',
  llhttp: '1.1.1',
  modules: '72',
  napi: '4',
  nghttp2: '1.38.0',
  node: '12.1.0',
  openssl: '1.1.1b',
  tz: '2019a',
  unicode: '12.1',
  uv: '1.28.0',
  v8: '7.4.288.21-node.16',
  zlib: '1.2.11'
}
```

For this proto file containing extraneous semicolons:

```proto
syntax="proto3";

message Foo {
  int32 x = 1;; // <-- two semicolons
}
```

The official `protoc` (v3.7.1) compiler will correctly parse and ferry the descriptors to a plugin:

```sh
→ protoc --js_out=`mktemp -d` foo.proto
→ echo $?
0
```

However, `pbjs`'s parser chokes on the extra semicolon:

```sh
→ pbjs foo.proto                       
/usr/local/lib/node_modules/protobufjs/cli/pbjs.js:245
            throw err;
            ^

Error: illegal token ';' (foo.proto, line 4)
    at illegal (/usr/local/lib/node_modules/protobufjs/src/parse.js:94:16)
    at parseType_block (/usr/local/lib/node_modules/protobufjs/src/parse.js:335:31)
    at ifBlock (/usr/local/lib/node_modules/protobufjs/src/parse.js:286:17)
    at parseType (/usr/local/lib/node_modules/protobufjs/src/parse.js:304:9)
    at parseCommon (/usr/local/lib/node_modules/protobufjs/src/parse.js:259:17)
    at parse (/usr/local/lib/node_modules/protobufjs/src/parse.js:728:21)
    at process (/usr/local/lib/node_modules/protobufjs/src/root.js:107:30)
    at fetch (/usr/local/lib/node_modules/protobufjs/src/root.js:166:13)
    at Root.load (/usr/local/lib/node_modules/protobufjs/src/root.js:194:13)
    at Root.loadSync (/usr/local/lib/node_modules/protobufjs/src/root.js:235:17)
```

[...]

PR #1323, opened 2019-11-21
parser: correctly handle empty statements (extra semicolons)

rodaine:
Fixes: #1322 

First commit should demonstrate failing tests, subsequent commits are the fix.

As a minor optimization, a `skip_all` function was added to recursively skip a token. Removing the `skip_all` changes should still work (with minor tweaks), so let me know if you'd prefer not to include it.


technotronic12:
@rodaine I've stumbled upon this problem also. Why wasn't it merged? did you find a workaround?

rodaine:
> @rodaine I've stumbled upon this problem also. Why wasn't it merged? did you find a workaround?

This PR and issue were never handled by the maintainers despite being open for about a year. My workaround was to not use this library and stick with the official JS code generator instead.

Issue #1306, opened 2019-10-03
generated .d.ts syntax error


wildfire810:
protobuf.js version: 6.8.8

D:\Project\skybox\Engine\Monju\Protocol\ProtoBuf\Protocol.d.ts
    Error:Error:line (74)TS2693: 'string' only refers to a type, but is being used as a value here.
    Error:Error:line (74)TS2531: Object is possibly 'null'.
    Error:Error:line (81)TS2304: Cannot find name 'fromObject'.
    Error:Error:line (81)TS2552: Cannot find name 'object'. Did you mean 'Object'?
    Error:Error:line (89)TS2552: Cannot find name 'toObject'. Did you mean 'Object'?
    Error:Error:line (89)TS2552: Cannot find name 'message'. Did you mean 'Message'?
    Error:Error:line (89)TS2552: Cannot find name 'options'. Did you mean 'Option'?
    Error:Error:line (89)TS2339: Property 'IConversionOptions' does not exist on type 'typeof import("D:/Project/skybox/Engine/Monju/node_modules/protobufjs/index")'.
    Error:Error:line (95)TS2552: Cannot find name 'toJSON'. Did you mean 'JSON'?
    Error:Error:line (164)TS2693: 'string' only refers to a type, but is being used as a value here.
    Error:Error:line (164)TS2531: Object is possibly 'null'.
    Error:Error:line (171)TS2304: Cannot find name 'fromObject'.
    Error:Error:line (171)TS2552: Cannot find name 'object'. Did you mean 'Object'?
    Error:Error:line (179)TS2552: Cannot find name 'toObject'. Did you mean 'Object'?
    Error:Error:line (179)TS2552: Cannot find name 'message'. Did you mean 'Message'?
    Error:Error:line (179)TS2304: Cannot find name 'options'.
    Error:Error:line (179)TS2339: Property 'IConversionOptions' does not exist on type 'typeof import("D:/Project/skybox/Engine/Monju/node_modules/protobufjs/index")'.
    Error:Error:line (185)TS2304: Cannot find name 'toJSON'.
    Error:Error:line (254)TS2693: 'string' only refers to a type, but is being used as a value here.
    Error:Error:line (254)TS2531: Object is possibly 'null'.
    Error:Error:line (261)TS2304: Cannot find name 'fromObject'.
    Error:Error:line (261)TS2304: Cannot find name 'object'.
    Error:Error:line (269)TS2304: Cannot find name 'toObject'.
[...]

PR #1333, opened 2019-12-21
fix:  jsdoc/jsdoc#1643

ceeji:
According to #1306, we need to upgrade jsdoc to >= 3.6.1, else `pbts` will not work innode 12.

All tests passed.


ceeji:
In fact for `pbts`'s issue, only CLI's `package.json` need to be upgraded. But i also upgrade `jsdoc`'s version in root's dev dependency as it might cause problem in the future and node 12 is a very important target.

Issue #1275, opened 2019-08-07
pbts generates invalid typescript when used on Windows


eric1221bday:
protobuf.js version: 6.8.8

<please describe the expected and actual behavior>

When on Windows, pbts generates invalid typescript even though pbjs is outputting the correct annotated js file.

Using the test proto from the protobufjs repo [here](https://github.com/protobufjs/protobuf.js/blob/master/tests/data/comments.proto).

The expected typescript file would contain
```js
public static verify(message: { [k: string]: any }): (string|null);
```
as seen [here](https://github.com/protobufjs/protobuf.js/blob/master/tests/data/comments.d.ts).

Instead the type file contains
```ts
public static verify(message: [ 'object' ].<string, any>): (string|null);
```
which is invalid typescript code.

The commands I ran for the conversion were
```sh
yarn run pbjs -t static-module -w commonjs -o protobufjs/comments.js protos/comments.proto
yarn run pbts -o protobufjs/comments.d.ts protobufjs/comments.js
```

medns:
I have the same issue

FrobtheBuilder:
This happens to me too.

FrobtheBuilder:
By the way, I think this is a compatibility issue with the latest version of Node. I'm using 12.4.0, and pbts hasn't really been updated at all in a year or so. The dependencies that it installs aren't at all related to parsing the actual jsdocs, so I don't think that's it either. I'm going to try on some previous versions and see which one works in a bit.

JiangJie:
> By the way, I think this is a compatibility issue with the latest version of Node. I'm using 12.4.0, and pbts hasn't really been updated at all in a year or so. The dependencies that it installs aren't at all related to parsing the actual jsdocs, so I don't think that's it either. I'm going to try on some previous versions and see which one works in a bit.

Waiting for your good news.

sw-clough:
I was getting the same error when using node 12.8.0. I've now rolled-back to the latest LTS release, 10.16.3, and now my files are correct.

chris-grabcad:
I'm using `protobuf.js` in an electron app - which we're currently upgrading to v5 of electron, which has a fixed compatible version of node 12.0.0 (electron v7 uses 12.4.0).

So workarounds like using an older version of node don't really work for us.

[...]

PR #1338, opened 2020-01-03
Issue 1275 - Upgrade jsdoc

chris-grabcad:
fixes issues with using pbts on Windows platforms
with modern versions of node 12.0.0 +

I have tried this locally on the project I work on with node 12.0.0 and it fixes problems as described here

https://github.com/protobufjs/protobuf.js/issues/1275#issuecomment-570588830

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,12 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/parser": {
+      "version": "7.7.7",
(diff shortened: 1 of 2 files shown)


dcodeIO:
Thanks, sounds good. Preferring updating jsdoc over the other options, but I'm too paranoid to merge package-lock changes. Currently looking into upgrading ALL the dependencies, and seeing how that turns out.

chris-grabcad:
Ok - thanks for fixing ... I will look out for an updated version so I can stop using our local fork 👍 .

PR #1344, opened 2020-01-12
fix: treat fixed64 as an unsigned long

derektom14:
https://developers.google.com/protocol-buffers/docs/proto3
By the spec, `fixed64` should be unsigned. However, the current code only considers `uint64` to be unsigned, and will force `fixed64` fields to become signed.

This change causes `fixed64` to also be treated as unsigned.


alexander-fenster:
The fix looks good but we'd like to have a regression test to see if this change could break anyone (in this case it will go to 7.x), and prevent it from breaking in future. I can write this test.

TODO(@alexander-fenster): write a test

PR #1087, opened 2018-07-17
* support default value for map key & value

shrimpz:
message TestMap
{
    map<int32, int32> x = 1;
}

tm.x[0] = 10;
tm.x[10] = 0;

Issue #1293, opened 2019-09-11
Empty strings read 


paambaati:
protobuf.js version: `3.6.1`

A Go-written message has an empty string as a `map` value, but the Node version fails when deserializing it because it sees it as `undefined`.

Proto file — https://github.com/liftbridge-io/liftbridge-grpc/blob/5694b15f251d2ff16d7d4c3e8d944aab327d3ef0/api.proto#L93-L104

The value for a `Message` on the Go side looks like this —

```
offset:69 key:"some-key" value:"some-vlaue" timestamp:1568177809419877000 subject:"test11" headers:<key:"reply" value:"" > headers:<key:"subject" value:"test11" >
```

**Reproduction steps**
1. `docker run -p 4222:4222 -ti nats:latest --debug --trace` in a window.
2. `go get github.com/liftbridge-io/go-liftbridge` and then `$GOPATH/bin/liftbridge --raft-bootstrap-seed --nats-servers nats://localhost:4222 --level debug` in another window.
3. Clone my repo from `https://github.com/paambaati/node-liftbridge.git` in yet another window.
4. `yarn install` or `npm install`
5. `yarn run debug` or `npm run debug`

When trying to read this message on the Node.js side, I get this error —

```
Error [AssertionError]: Assertion failed
    at new goog.asserts.AssertionError (~/node-liftbridge/node_modules/google-protobuf/google-protobuf.js:1166:22)
    at Object.goog.asserts.doAssertFailure_ (~/node-liftbridge/node_modules/google-protobuf/google-protobuf.js:1186:9)
    at Object.goog.asserts.assert [as assert] (~/node-liftbridge/node_modules/google-protobuf/google-protobuf.js:1193:55)
    at Function.jspb.Map.deserializeBinary (~/node-liftbridge/node_modules/google-protobuf/google-protobuf.js:3881:18)
    at ~/node-liftbridge/grpc/generated/api_pb.js:2259:18
    at jspb.BinaryReader.readMessage (~/node-liftbridge/node_modules/google-protobuf/google-protobuf.js:3517:5)
    at Function.proto.proto.Message.deserializeBinaryFromReader (~/node-liftbridge/grpc/generated/api_pb.js:2258:14)
    at Function.proto.proto.Message.deserializeBinary (~/node-liftbridge/grpc/generated/api_pb.js:2214:30)
    at deserialize_proto_Message (~/node-liftbridge/grpc/generated/api_grpc_pb.js:59:25)
[...]

PR #1348, opened 2020-01-20
fix: handling of map entries with omitted key or value

### LINKED ISSUE #843: Map fields with empty string as key are handled incorrectly

### LINKED ISSUE #960: map doesn't work for empty values
protobuf.js version: 6.8.3

protobuf.js throws an error when trying to decode any map encoded using google's protobuf for c#

```js
message= Message.decode(buffer);
```

```
Error: invalid wire type 6 at offset 896
    at n.skipType (reader.js:375)
```

schiffermtq:
According to [1], map encoding must be compatible with a repeated message
using indices 1 and 2 for key and value. In particular this implies that
both key and value may be omitted when they are equal to the default
value - which some protobuf implementations like protobuf-c actually do.

The comments in the added testcase are based on the output of
protobuf-inspector [2].

[1] https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility
[2] https://github.com/jmendeth/protobuf-inspector

This is a cleaned up version of #1087:
- Added detailed commit message
- Improved coding style
- Added test case

Supersedes #845
Supersedes #1087
Fixes #843
Fixes #960 
Fixes #1087


src/decoder.js
```diff
@@ -37,22 +37,48 @@ function decoder(mtype) {
...
+                ("k=\"%s\"", types.defaults[field.keyType]);
+            else if (types.defaults[field.keyType] !== undefined) gen
+                ("k=%s", types.defaults[field.keyType]);
+            else gen
+                ("k=null");
+
+            if (type === "string") gen
+                ("value=\"%s\"", types.defaults[type]);
+            else if (types.defaults[type] !== undefined) gen
+                ("value=%s", types.defaults[type]);
```
johncantrell97: if type === "bytes" then this tries to insert the default value of an empty array with %s and it ends up inserting empty string "" instead of "[]".  This is because %s calls String([]) when we need %j here so we use JSON.stringify([]).

schiffermtq: Thanks, fixed.

paambaati:
@schiffermtq Thank you so much for this PR! This should also fix #1293

paambaati:
@nicolasnoble @alexander-debater Anything we can do to get this PR reviewed/merged?

paambaati:
@nicolasnoble @alexander-fenster Anything we can do to get this PR reviewed/merged?

adriancable:
Bump.

adriancable:
YES! Thank you so much.

PR #1349, opened 2020-01-20
Prevent jsdoc upgrade

### LINKED ISSUE #1342: pbts doesn't generate enum typings when the enum is defined after its use
protobuf.js version: 6.8.8

Take the following .proto file:

```
syntax = "proto3";

message Test {
	Status status = 0;
}

enum Status {
	STATUS_ERROR = 0;
	STATUS_OK = 1;
}
```

With the following commands:
```
pbjs -t static-module -w es6 -o ./test.js ./test.proto
pbts -o ./test.d.ts ./test.js
```

The generated .d.ts file will be unusable, as it references the type `Status`, buts its definition is missing. Using commonjs instead of es6 yields the same results.

Reversing the order of the definitions leads to a different result: while a Status definition is emitted, now ITest is missing instead.

schiffermtq:
pbts currently doesn't work with jsdoc 3.6.x and will emit broken
definitions when trying to use it.

As the toplevel package-lock.json is not respected by `npm install` as
executed by the CLI utils, fix the jsdoc version to 3.5.5 in
package.json.

Fixes #1342

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "gulp-uglify": "^3.0.0",
     "istanbul": "^0.4.5",
     "jaguarjs-jsdoc": "github:dcodeIO/jaguarjs-jsdoc",
-    "jsdoc": "^3.5.5",
+    "jsdoc": "3.5.5",
     "minimist": "^1.2.0",
     "reflect-metadata": "^0.1.12",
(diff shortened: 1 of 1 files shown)


schiffermtq:
Hmm, the Node 12 incompatibility is unfortunate.

Issue #1353, opened 2020-02-01
Cutting a release


JustinBeckwith:
Greetings folks!  I'd like to be able to cut a release if possible.  After we have a new release or two out, and stabilized, I'd also like to start automating releases so we're not dependent on any individual to keep things going, rather group owned credentials.  

- Is the process for cutting a release documented today? I see `npm run release`, which appears to create a tag and compiled assets.  It's unclear how this dovetails with running `npm publish`. 

- Would it be possible to add [`google-wombot`](https://www.npmjs.com/~google-wombot) as an admin on the `protobufjs` package on npm?  Alternatively, could we add `justinbeckwith`, `fenster`, or `bcoe` so we have someone with admin?

- Would it be possible to expand the list of users with admin on the repo, to include `JustinBeckwith`, `alexander-fenster`, and `bcoe`, so we can set up a little more automation?

@dcodeIO - I realize this is askin a lot, but we really do want to make sure things can keep rolling here, especially dependency updates :) 

@alexander-fenster @bcoe as FYI

bcoe:
would love to pitch in as best I can, I've been using the automation machinery our team has been building on projects I manage, such as [yargs](https://github.com/yargs/yargs/pull/1521).

dcodeIO:
There isn't really a release process in place as of today. Mostly a bunch of scripts to do it manually, with `npm run release` being a shorthand to run all the relevant tasks before issuing a publish.

I'm open to your suggestions and google-wombot seems like a great tool. An alternative might be to do something like we do over at AssemblyScript, which is solely based on GitHub Actions, not much more than [a few lines](https://github.com/AssemblyScript/assemblyscript/blob/master/.github/workflows/publish.yml) of GH actions code. Publishes nightlies once a day if there have been changes, respectively creates releases if the version in package.json changes.

JustinBeckwith:
Both approaches sound awesome to me - the biggest thing I'd like to see is automation so we don't have to bug you :) 

I take the [`semantic-release`](https://www.npmjs.com/package/semantic-release) release approach in my projects, which cuts a new release _every time_ a relevant PR is landed.  You cna see an example of combining that with GitHub Actions [here](https://github.com/JustinBeckwith/gcx/blob/master/.github/workflows/ci.yaml#L39). 

[...]

PR #1354, opened 2020-02-03
Move CI to GitHub Actions

dcodeIO:
This PR intends to replace our Travis-CI integration with a functionally equivalent GitHub Actions workflow. NFCI, except that it opens the door towards extending what we have with a release mechanism eventually.

See: https://github.com/protobufjs/protobuf.js/issues/1353


dcodeIO:
Interesting...

PR #1360, opened 2020-02-19
docs: fix the JSON example in README.md

fabien-h:
The json example was not equivalent to the .proto file. The message type should be nested in a namespace.


alexander-fenster:
@fabien-h I'll be happy to merge it, can you please merge the latest master into your branch? Thanks!

Issue #1221, opened 2019-05-13
util.isNode / Error: not supported


jumpinvestor:
protobuf.js version: 6.8.8
Ubuntu: 19.04
Node: 12.2.0
Angular: 7.2.15 with server-side rendering enabled

I'm seeing "Error: not supported" popping up in an Angular 7 app on Node 12.2.0. Just upgraded Node from 10.15.3, where it did not have the error.

<img width="1086" alt="Screen Shot 2019-05-13 at 3 35 06 PM" src="https://user-images.githubusercontent.com/14829156/57658795-fe166180-7594-11e9-81f4-e64d352ce27c.png">

After a little digging, it looks like this references _/src/root.js_:233-234:

<img width="966" alt="Screen Shot 2019-05-13 at 3 50 00 PM" src="https://user-images.githubusercontent.com/14829156/57659329-d32d0d00-7596-11e9-9349-36a128b363c2.png">

and _/src/util/minimal.js_:55:

<img width="968" alt="Screen Shot 2019-05-13 at 3 38 05 PM" src="https://user-images.githubusercontent.com/14829156/57659002-96144b00-7595-11e9-8644-9ecf1686eab5.png">

After commenting out the check in root.js, everything seems to work fine. Is there another solution?

dcodeIO:
The check is here: https://github.com/protobufjs/protobuf.js/blob/master/src/util/minimal.js#L55

There might be something in your stack invalidating it, or did node change something?

jumpinvestor:
I think it might be related to Angular's server-side rendering. The util object captures global.document but not global.process, which could make sense for SSR:

<img width="463" alt="Screen Shot 2019-05-14 at 8 23 34 AM" src="https://user-images.githubusercontent.com/14829156/57710490-e46f2b80-7621-11e9-935a-b783fe28326b.png">

MichaelSolati:
I'm having a similar, so what I found worked (which I don't think would be a hard/serious change) is in `src/util/minimal.js` I reordered the `util.global`. It was:

```JavaScript
util.global = typeof global !== "undefined" && global
           || typeof window !== "undefined" && window
           || typeof self   !== "undefined" && self
           || this; // eslint-disable-line no-invalid-this
```

And I swapped it to this:

```JavaScript
util.global = typeof global !== "undefined" && global
           || typeof window !== "undefined" && window
           || typeof self   !== "undefined" && self
           || this; // eslint-disable-line no-invalid-this
```

[...]

PR #1363, opened 2020-02-20
fix: updated isNode check (#1221)

diegoblattner:
* Fixes https://github.com/protobufjs/protobuf.js/issues/1221 when used within SSR.
`process` is a global variable since node [0.1.7](https://nodejs.org/api/globals.html#globals_process) and versions since [0.2.0](https://nodejs.org/api/process.html#process_process_versions).


dcodeIO:
Haven't checked and might be a false alarm, but referencing `process` directly as proposed here might cause the bundler to pull in a `process` shim, increasing distribution size, and potentially breaking browser builds due to assuming they'd run under node. Thought I mention :)

alexander-fenster:
Thanks @dcodeIO, let me check this.  This is basically why we decided to merge stuff on Fridays and let it bake for a while before we make an actual release next week :)   I'll update the issue and/or send a fix if needed.

alexander-fenster:
So, several things here:
- it indeed includes a shim for `process` now (I checked with `webpack`). It pulls in the whole `./node_modules/process/browser.js` which adds 1.6K of code when minimized. That's sad.
- it may or may not break some builds indeed (if someone just checks for `typeof process !== "undefined"` to check if it's Node or not, this code may stop working within the bundle), even though this check is actually wrong;
- adding a shim for `process` can be disabled in the bundler config (e.g. `webpack.config.js` or similar).
- there is an NPM module called [`browser-or-node`](https://www.npmjs.com/package/browser-or-node) (28.5k weekly downloads) that actually implements `isNode` in the same way as introduced in this PR.

My personal opinion here is that it's better to fix the code for users who actually reported they have problems (in #1221), and then fix it once more if the fix actually breaks someone else's code (it definitely will, obligatory [xkcd](https://xkcd.com/1172/)). Folks who really care about the extra 1.6K of minimized code will be able to configure their bundler to remove this shim.

alexander-fenster:
...but the bad thing is that it actually breaks `npm run make` (`ReferenceError: process is not defined`), so yes we'll either need to rollback or figure out a different way to fix it.

dcodeIO:
The `util.global` alias was intentionally introduced to avoid the shim bundling issues we are having now, so perhaps https://github.com/protobufjs/protobuf.js/issues/1221#issuecomment-493147824 is a better fix than this one. Like, `util.global.process` being missing appears to be more of a symptom of an underlying issue with what we alias into `util.global`. If the fix proposed within the linked comment doesn't introduce any immediate issues, I recommend to revert this PR and fix `util.global` as proposed instead.

dcodeIO:
Perhaps with an additional safeguard [there](https://github.com/protobufjs/protobuf.js/blob/b950877c86676399674821fca4cf444f046b5acb/src/util/minimal.js#L28):

```js
function checkNodeGlobal(global) {
  if (global && global.process) return global; // or whatever reasonably identifies node
  return null;
}

// global object reference
util.global = typeof global !== "undefined" && checkNodeGlobal(global)
           || typeof window !== "undefined" && window
           || typeof self   !== "undefined" && self
           || this; // eslint-disable-line no-invalid-this
```

PR #1417, opened 2020-06-08
don't overwrite old value, allow seperate pbjs compile, fix #1383

### LINKED ISSUE #1383: pbjs overwrite package so seperate compile multiple file of one package is not allowed
protobuf.js version: 6.8.8
 
we have two proto file a.proto and b.proto

```protobuf
syntax = "proto3";
package ttt;
message A{
    string RID = 1;
}
```

```protobuf
syntax = "proto3";
package ttt;
message B{
    string RID = 1;
}
```
when build with pbjs with two file together, the generated file is ok,$root.ttt.A and $root.ttt.B is set right.
```shell
pbjs -t static-module -w es6 a.proto b.proto > ttt.js
```

when build a.proto and b.proto seperately, each output has reset the $root.ttt

```js
export const ttt = $root.ttt = (() => {

    /**
     * Namespace ttt.
     * @exports ttt
     * @namespace
     */
    const ttt = {};
```
so when import one generated js file will overwrite any previous setting with the same package name.

the solution is changing cli/targets/static.js line 132 to below
```js
push((config.es6 ? "const" : "var") + " " + escapeName(ns.name) + " = {};");
```
to
```js
push((config.es6 ? "const" : "var") + " " + escapeName(ns.name) + " = $root."+escapeName(ns.name)+" || {};");
```

yangjuncode:
don't overwrite old value, allow seperate pbjs compile, fix #1383

### DIFF
--- a/cli/targets/static.js
+++ b/cli/targets/static.js
@@ -129,7 +129,7 @@ function buildNamespace(ref, ns) {
             ns.parent instanceof protobuf.Root ? "@exports " + escapeName(ns.name) : "@memberof " + exportName(ns.parent),
             "@namespace"
         ]);
-        push((config.es6 ? "const" : "var") + " " + escapeName(ns.name) + " = $root." + escapeName(ns.name) + " || {};");
+        push((config.es6 ? "const" : "var") + " " + escapeName(ns.name) + " = $root." + (ns.parent instanceof protobuf.Root ? escapeName(ns.name) : (exportName(ns.parent) + '.' + escapeName(ns.name))) + " || {};");
     }
 
     ns.nestedArray.forEach(function(nested) {
--- a/cli/targets/static.js
+++ b/cli/targets/static.js
@@ -129,7 +129,7 @@ function buildNamespace(ref, ns) {
             ns.parent instanceof protobuf.Root ? "@exports " + escapeName(ns.name) : "@memberof " + exportName(ns.parent),
             "@namespace"
         ]);
-        push((config.es6 ? "const" : "var") + " " + escapeName(ns.name) + " = {};");
+        push((config.es6 ? "const" : "var") + " " + escapeName(ns.name) + " = $root."+escapeName(ns.name)+" || {};");
     }
 
     ns.nestedArray.forEach(function(nested) {
--- a/cli/targets/static.js
+++ b/cli/targets/static.js
@@ -129,7 +129,7 @@ function buildNamespace(ref, ns) {
             ns.parent instanceof protobuf.Root ? "@exports " + escapeName(ns.name) : "@memberof " + exportName(ns.parent),
             "@namespace"
         ]);
-        push((config.es6 ? "const" : "var") + " " + escapeName(ns.name) + " = $root."+escapeName(ns.name)+" || {};");
+        push((config.es6 ? "const" : "var") + " " + escapeName(ns.name) + " = $root." + escapeName(ns.name) + " || {};");
     }
 
     ns.nestedArray.forEach(function(nested) {
--- a/cli/targets/static.js
+++ b/cli/targets/static.js
@@ -129,7 +129,7 @@ function buildNamespace(ref, ns) {
             ns.parent instanceof protobuf.Root ? "@exports " + escapeName(ns.name) : "@memberof " + exportName(ns.parent),
             "@namespace"
         ]);
-        push((config.es6 ? "const" : "var") + " " + escapeName(ns.name) + " = $root." + (ns.parent instanceof protobuf.Root ? escapeName(ns.name) : (exportName(ns.parent) + '.' + escapeName(ns.name))) + " || {};");
+        push((config.es6 ? "const" : "var") + " " + escapeName(ns.name) + " = $root." + (ns.parent instanceof protobuf.Root ? escapeName(ns.name) : exportName(ns.parent) + "." + escapeName(ns.name)) + " || {};");
     }
 
     ns.nestedArray.forEach(function(nested) {


cli/targets/static.js
```diff
@@ -129,7 +129,7 @@ function buildNamespace(ref, ns) {
             ns.parent instanceof protobuf.Root ? "@exports " + escapeName(ns.name) : "@memberof " + exportName(ns.parent),
             "@namespace"
         ]);
-        push((config.es6 ? "const" : "var") + " " + escapeName(ns.name) + " = {};");
+        push((config.es6 ? "const" : "var") + " " + escapeName(ns.name) + " = $root." + escapeName(ns.name) + " || {};");
```
alexander-fenster, line 132: I think this change is valid but it might fail for the nested package names (if a package name has multiple components, e.g. `google.cloud.something`).  Can you please add some tests to this PR that checks this behavior?

Also, just FYI, in Google Cloud libraries, we use `-r` option for `pbjs` to make sure we don't override the existing root.

yangjuncode: you are right, when package name has multiple level, the change is not correct, i have push new code to fix it.
 i test in my machine, but don't know how to write such a test. 

given two proto: a.proto
```protobuf
syntax = "proto3";
package ttt.ff.a;
message A{
    string RID = 1;
}
```
b.proto
```protobuf
syntax = "proto3";
package ttt.cc.b;
message B{
    string RID = 1;
}
```

then generated js with cmd below
```bash
pbjs --no-create --no-verify --no-convert --no-delimited --force-long -t static-module -w es6 a.proto > t.js
pbjs --no-create --no-verify --no-convert --no-delimited --force-long -t static-module -w es6 b.proto > tt.js
```

the result is t.js
```javascript
/*eslint-disable block-scoped-var, id-length, no-control-regex, no-magic-numbers, no-prototype-builtins, no-redeclare, no-shadow, no-var, sort-vars*/
import * as $protobuf from "protobufjs/minimal";

// Common aliases
const $Reader = $protobuf.Reader, $Writer = $protobuf.Writer, $util = $protobuf.util;

// Exported root namespace
const $root = $protobuf.roots["default"] || ($protobuf.roots["default"] = {});

export const ttt = $root.ttt = (() => {

    /**
     * Namespace ttt.
     * @exports ttt
     * @namespace
     */
    const ttt = $root.ttt || {};

    ttt.ff = (function() {

        /**
         * Namespace ff.
         * @memberof ttt
         * @namespace
         */
        const ff = $root.ttt.ff || {};

        ff.a = (function() {

            /**
             * Namespace a.
             * @memberof ttt.ff
             * @namespace
             */
            const a = $root.ttt.ff.a || {};

            a.A = (function() {

                /**
                 * Properties of a A.
                 * @memberof ttt.ff.a
                 * @interface IA
                 * @property {string|null} [RID] A RID
                 */

                /**
                 * Constructs a new A.
                 * @memberof ttt.ff.a
                 * @classdesc Represents a A.
                 * @implements IA
                 * @constructor
                 * @param {ttt.ff.a.IA=} [properties] Properties to set
                 */
                function A(properties) {
                    if (properties)
                        for (let keys = Object.keys(properties), i = 0; i < keys.length; ++i)
                            if (properties[keys[i]] != null)
                                this[keys[i]] = properties[keys[i]];
                }

                /**
                 * A RID.
                 * @member {string} RID
                 * @memberof ttt.ff.a.A
                 * @instance
                 */
                A.prototype.RID = "";

                /**
                 * Encodes the specified A message. Does not implicitly {@link ttt.ff.a.A.verify|verify} messages.
                 * @function encode
                 * @memberof ttt.ff.a.A
                 * @static
                 * @param {ttt.ff.a.IA} message A message or plain object to encode
                 * @param {$protobuf.Writer} [writer] Writer to encode to
                 * @returns {$protobuf.Writer} Writer
                 */
                A.encode = function encode(message, writer) {
                    if (!writer)
                        writer = $Writer.create();
                    if (message.RID != null && Object.hasOwnProperty.call(message, "RID"))
                        writer.uint32(/* id 1, wireType 2 =*/10).string(message.RID);
                    return writer;
                };

                /**
                 * Decodes a A message from the specified reader or buffer.
                 * @function decode
                 * @memberof ttt.ff.a.A
                 * @static
                 * @param {$protobuf.Reader|Uint8Array} reader Reader or buffer to decode from
                 * @param {number} [length] Message length if known beforehand
                 * @returns {ttt.ff.a.A} A
                 * @throws {Error} If the payload is not a reader or valid buffer
                 * @throws {$protobuf.util.ProtocolError} If required fields are missing
                 */
                A.decode = function decode(reader, length) {
                    if (!(reader instanceof $Reader))
                        reader = $Reader.create(reader);
                    let end = length === undefined ? reader.len : reader.pos + length, message = new $root.ttt.ff.a.A();
                    while (reader.pos < end) {
                        let tag = reader.uint32();
                        switch (tag >>> 3) {
                        case 1:
                            message.RID = reader.string();
                            break;
                        default:
                            reader.skipType(tag & 7);
                            break;
                        }
                    }
                    return message;
                };

                return A;
            })();

            return a;
        })();

        return ff;
    })();

    return ttt;
})();

export { $root as default };
```
tt.js
```javascript
/*eslint-disable block-scoped-var, id-length, no-control-regex, no-magic-numbers, no-prototype-builtins, no-redeclare, no-shadow, no-var, sort-vars*/
import * as $protobuf from "protobufjs/minimal";

// Common aliases
const $Reader = $protobuf.Reader, $Writer = $protobuf.Writer, $util = $protobuf.util;

// Exported root namespace
const $root = $protobuf.roots["default"] || ($protobuf.roots["default"] = {});

export const ttt = $root.ttt = (() => {

    /**
     * Namespace ttt.
     * @exports ttt
     * @namespace
     */
    const ttt = $root.ttt || {};

    ttt.cc = (function() {

        /**
         * Namespace cc.
         * @memberof ttt
         * @namespace
         */
        const cc = $root.ttt.cc || {};

        cc.b = (function() {

            /**
             * Namespace b.
             * @memberof ttt.cc
             * @namespace
             */
            const b = $root.ttt.cc.b || {};

            b.B = (function() {

                /**
                 * Properties of a B.
                 * @memberof ttt.cc.b
                 * @interface IB
                 * @property {string|null} [RID] B RID
                 */

                /**
                 * Constructs a new B.
                 * @memberof ttt.cc.b
                 * @classdesc Represents a B.
                 * @implements IB
                 * @constructor
                 * @param {ttt.cc.b.IB=} [properties] Properties to set
                 */
                function B(properties) {
                    if (properties)
                        for (let keys = Object.keys(properties), i = 0; i < keys.length; ++i)
                            if (properties[keys[i]] != null)
                                this[keys[i]] = properties[keys[i]];
                }

                /**
                 * B RID.
                 * @member {string} RID
                 * @memberof ttt.cc.b.B
                 * @instance
                 */
                B.prototype.RID = "";

                /**
                 * Encodes the specified B message. Does not implicitly {@link ttt.cc.b.B.verify|verify} messages.
                 * @function encode
                 * @memberof ttt.cc.b.B
                 * @static
                 * @param {ttt.cc.b.IB} message B message or plain object to encode
                 * @param {$protobuf.Writer} [writer] Writer to encode to
                 * @returns {$protobuf.Writer} Writer
                 */
                B.encode = function encode(message, writer) {
                    if (!writer)
                        writer = $Writer.create();
                    if (message.RID != null && Object.hasOwnProperty.call(message, "RID"))
                        writer.uint32(/* id 1, wireType 2 =*/10).string(message.RID);
                    return writer;
                };

                /**
                 * Decodes a B message from the specified reader or buffer.
                 * @function decode
                 * @memberof ttt.cc.b.B
                 * @static
                 * @param {$protobuf.Reader|Uint8Array} reader Reader or buffer to decode from
                 * @param {number} [length] Message length if known beforehand
                 * @returns {ttt.cc.b.B} B
                 * @throws {Error} If the payload is not a reader or valid buffer
                 * @throws {$protobuf.util.ProtocolError} If required fields are missing
                 */
                B.decode = function decode(reader, length) {
                    if (!(reader instanceof $Reader))
                        reader = $Reader.create(reader);
                    let end = length === undefined ? reader.len : reader.pos + length, message = new $root.ttt.cc.b.B();
                    while (reader.pos < end) {
                        let tag = reader.uint32();
                        switch (tag >>> 3) {
                        case 1:
                            message.RID = reader.string();
                            break;
                        default:
                            reader.skipType(tag & 7);
                            break;
                        }
                    }
                    return message;
                };

                return B;
            })();

            return b;
        })();

        return cc;
    })();

    return ttt;
})();

export { $root as default };

```

alexander-fenster:
TODO(@alexander-fenster): I'll copy the example code to the test suite and test it for the different namespaces.

PR #1418, opened 2020-06-09
Better comment parse

recih:
1. fix trailing comment in last line can be recognized as block comment for current field by mistake.
2. new `preferTrailingComment` option in `IParseOptions`, user can choose the priority of trailing comments and block commnets
3. test updated.


recih:
bug found, will create a new PR later.

PR #1419, opened 2020-06-09
feat: better comment parse

recih:
1. fix trailing comment in last line can be recognized as leading comment for current field by mistake.
2. new preferTrailingComment option in IParseOptions, user can choose the priority of trailing comments and leading commnets
3. test updated.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1046,6 +1046,9 @@ export interface IParseOptions {
 
     /** Recognize double-slash comments in addition to doc-block comments. */
     alternateCommentMode?: boolean;
+
+    /** Use trailing comment when both leading comment and trailing comment exist. */
+    preferTrailingComment?: boolean;
 }
 
 /** Options modifying the behavior of JSON serialization. */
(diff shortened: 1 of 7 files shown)


murgatroid99:
Can you explain a bit more what the incorrect previous behavior was and how your fix changes it?

recih:
> Can you explain a bit more what the incorrect previous behavior was and how your fix changes it?

```
message Example {
  /// leading comment A
  a = 1; /// trailing comment A
  b = 2; /// trailing comment B
}
```
when there is no extra new line between two fields, you will get `Example.fields['b'].comment === "trailing comment A"`, the trailing comment of field `a` is considered as the leading comment of field `b`, which makes no sense.

after my fix, you will get:
```
Example.fields['a'].comment === "leading comment A"
Example.fields['b'].comment === "trailing comment B"
```

also there is an option `preferTrailingComment`, when its true, you will get:
```
Example.fields['a'].comment === "trailing comment A"
Example.fields['b'].comment === "trailing comment B"
```

recih:
@murgatroid99 will you review my PR, please :heart:

PR #1441, opened 2020-07-13
Fix util.global

dcodeIO:
An alternative to https://github.com/protobufjs/protobuf.js/pull/1363, attempting to fix the problem at the `util.global` level as proposed in https://github.com/protobufjs/protobuf.js/issues/1221#issuecomment-493147824.

Reverts #1363 and #1440, with the latter not being needed anymore. See also: https://github.com/protobufjs/protobuf.js/pull/1363#issuecomment-657820448

### DIFF
--- a/src/util/minimal.js
+++ b/src/util/minimal.js
@@ -26,8 +26,8 @@ util.pool = require("@protobufjs/pool");
 util.LongBits = require("./longbits");
 
 // global object reference
-util.global = typeof window !== "undefined" && window
-           || typeof global !== "undefined" && global
+util.global = typeof global !== "undefined" && Object.prototype.toString.call(global) === "[object global]" && global
+           || typeof window !== "undefined" && window
            || typeof self   !== "undefined" && self
            || this; // eslint-disable-line no-invalid-this
 
@@ -52,7 +52,7 @@ util.emptyObject = Object.freeze ? Object.freeze({}) : /* istanbul ignore next *
  * @type {boolean}
  * @const
  */
-util.isNode = Boolean(process && process.versions && process.versions.node);
+util.isNode = Boolean(util.global.process && util.global.process.versions && util.global.process.versions.node);
 
 /**
  * Tests if the specified value is an integer.


alexander-fenster:
Thank you @dcodeIO, you definitely have more experience with bundlers than I do :)

dcodeIO:
Sorry for second-guessing btw :). If I have any bundler experience, then it is in preventing them from functioning as intended. As lost as everybody else otherwise.

evgenius:
`Object.prototype.toString.call(global) === "[object global]"` seems to be too strict.
Now protobufjs fails to initialize in the Jest environment since Jest modifies `global`. See https://github.com/facebook/jest/blob/d7f3427/packages/jest-environment-node/src/index.ts#L30

dcodeIO:
Hmm, more forgiving checks there could be

```
Object.prototype.toString.call(global.process) === "[object process]"
```

if `process` remains unmodified, and/or

```
global.setTimeout === setTimeout
```

to just validate that it's at least somewhat an actual global object.

dcodeIO:
A more general question would be what Jest is actually trying to test there. Does a Jest test suite automatically assume that it's testing in a node environment, or may it also test client code while filling in browser APIs? If the latter is possible, protobuf.js cannot decide on its own but will need some sort of flag to indicate, i.e. setting `util.isNode` manually.

evgenius:
How about changing the order?
```
util.isNode = typeof global !== "undefined"
           && global.process
           && global.process.versions
           && global.process.versions.node;
util.global = util.isNode && global 
           || typeof window !== "undefined" && window
           || typeof self   !== "undefined" && self
           || this;
```

dcodeIO:
That looks good as well, just needs another check `&& global` before accessing `global.process` in case it's `null`.

evgenius:
Ah, right, forgot about `null` :)

alexander-fenster:
Folks also suggested https://github.com/protobufjs/protobuf.js/pull/1444 as a possible fix to the strict `[object global]`. I must admit I lost track of all the possible environments and frameworks this code should work in :)

PR #1455, opened 2020-07-18
fix: add braces around message decoder case statements so 'let' variable declarations cannot overlap

### LINKED ISSUE #1452: pbjs generates a syntax error with es6 javascript
protobuf.js version: 6.10.1

### Expected behavior
pbjs should generate javascript free of syntax errors when using wrap = es6

### Actual behavior
If there are 2 map fields in a protobuf message, pbjs generates code that redeclares let which is a SyntaxError. This did not happen in version 6.9.0

### Code
```js
pbjs -t static-module -w es6 ./message.proto
```

```
message Test {
  map<string, string> map_1 = 1;
  map<string, string> map_2 = 2;
}

```

Resulting javascript - `end2` is declared twice, in `case 1` and `case 2` but the case statements are not wrapped in brackets
```
    Test.decode = function decode(reader, length) {
        if (!(reader instanceof $Reader))
            reader = $Reader.create(reader);
        let end = length === undefined ? reader.len : reader.pos + length, message = new $root.Test(), key, value;
        while (reader.pos < end) {
            let tag = reader.uint32();
            switch (tag >>> 3) {
            case 1:
                if (message.map_1 === $util.emptyObject)
                    message.map_1 = {};
                let end2 = reader.uint32() + reader.pos;
                key = "";
                value = "";
                while (reader.pos < end2) {
                    let tag2 = reader.uint32();
                    switch (tag2 >>> 3) {
                    case 1:
                        key = reader.string();
                        break;
                    case 2:
                        value = reader.string();
                        break;
                    default:
                        reader.skipType(tag2 & 7);
                        break;
                    }
                }
                message.map_1[key] = value;
                break;
            case 2:
                if (message.map_2 === $util.emptyObject)
                    message.map_2 = {};
                let end2 = reader.uint32() + reader.pos;
                key = "";
                value = "";
                while (reader.pos < end2) {
                    let tag2 = reader.uint32();
                    switch (tag2 >>> 3) {
                    case 1:
                        key = reader.string();
                        break;
                    case 2:
                        value = reader.string();
                        break;
                    default:
                        reader.skipType(tag2 & 7);
                        break;
                    }
                }
                message.map_2[key] = value;
                break;
            default:
                reader.skipType(tag & 7);
                break;
            }
        }
        return message;
    };
```

Javascript using protobufjs 6.9.0

```
    Test.decode = function decode(reader, length) {
        if (!(reader instanceof $Reader))
            reader = $Reader.create(reader);
        let end = length === undefined ? reader.len : reader.pos + length, message = new $root.Test(), key;
        while (reader.pos < end) {
            let tag = reader.uint32();
            switch (tag >>> 3) {
            case 1:
                reader.skip().pos++;
                if (message.map_1 === $util.emptyObject)
                    message.map_1 = {};
                key = reader.string();
                reader.pos++;
                message.map_1[key] = reader.string();
                break;
            case 2:
                reader.skip().pos++;
                if (message.map_2 === $util.emptyObject)
                    message.map_2 = {};
                key = reader.string();
                reader.pos++;
                message.map_2[key] = reader.string();
                break;
            default:
                reader.skipType(tag & 7);
                break;
            }
        }
        return message;
    };
```

AndiDog:
Closes #1452

Tested with my minimal problem example repo https://github.com/AndiDog/protobufjs-identifier-already-declared


alnr:
Bump

alexander-fenster (approved):

alexander-fenster:
Hi @AndiDog,

This looks reasonable to me. Just wondering, would it be possible to add a quick test since apparently this case is not covered by any existing tests?

Thank you!

AndiDog:
I added a test, rebased, and regenerated files via `npm run make` (those changes are in a separate commit).

alnr:
Bump. Can this be merged please? Thank you!

alexander-fenster:
@AndiDog @alnr Something in this PR - I have no idea what exactly :( - makes Node 6 and below tests fail:

image

The failure is somewhere in `chalk`, I really have no idea what's going on here with dependencies so it starts failing. I might have some time to dig into this later this week but if you folks could take a look too it would be great :)

AndiDog:
Let's please ignore and remove Node 6 support – it has been end-of-life for a long time (see [node.js release and support cycle](https://nodejs.org/en/about/releases/)).

From the error screenshot, I could guess that Node 6 doesn't support the spread syntax yet (`...x`). Since that is in a dependency, that means they have already removed support for this very old version.

alexander-fenster:
@AndiDog I surely understand it's in a dependency and understand it's because of `...`, but it does not explain why the CI does not fail in the same way on the main branch or in other PRs. Apparently, the newer version of `chalk` gets picked only in this PR, which is weird.

As for removing support for older versions, it's up to @dcodeIO - I'm not an owner here, I just merge pull requests :)  I personally don't see a problem supporting old versions until it's not too hard.

alexander-fenster:
(to add to version support policy, we at Google Cloud client libraries tend to semver major our Node libraries at least every year dropping support for older versions, but I totally understand the reasoning for keeping the code working on older versions - again, it's not my decision here)

dcodeIO:
> I personally don't see a problem supporting old versions until it's not too hard.

Typically my thinking as well. Wondering how this magically broke, though, given `"chalk": "^4.0.0"` in package.json?

AndiDog:
chalk only [supports nodejs 10+](https://github.com/chalk/chalk/blob/master/.travis.yml). The spread syntax used here was introduced [2019-07 in this commit](https://github.com/chalk/chalk/commit/c08417e88ca41f7133eb1a00e0b3fc8aefbb6cc9). It has already been in [v3.0.0](https://github.com/chalk/chalk/blob/v3.0.0/source/index.js). With this, I don't think there's still a good reason to keep nodejs 6 support, unfortunately.

alexander-fenster:
@AndiDog I'll try to figure out what's going on with the CI in this PR. As I said, the weird thing is that the CI is green in the main branch, so it's something in this PR that triggers that :)   I'll take a look.

AndiDog:
I tried on my branch with a nodejs 6 Docker image and "npm test" gave me the same error. Let's try that on the main branch. Any caching in place with outdated node_modules, maybe?

alexander-fenster:
@AndiDog It works :) 

```
$ docker run -it --rm --entrypoint /bin/bash node:6
# git clone https://github.com/protobufjs/protobuf.js
# cd protobuf.js/
# npm install
# npm test
# echo $?
0
# node --version
v6.17.1
```

(I haven't looked deep into the changes in your branch yet, but I will)

def-ai:
Hey just hit the bug that this fixes - would be fantastic to get merged if possible :). Still an issue on 6.11.2, latest as of today.

joshross12:
Is there a timeline for when we can expect to get this feature merged in? Thank you!

PR #1464, opened 2020-07-30
verify now allows string values in enums

### LINKED ISSUE #1261: Error "enum typ: enum value expected" when JSON enum Value is String
Version `protobufjs": "^6.8.8"`

I have a message like this:
```enum ReferenceType {
  Point = 0;
  Plane = 1;
  Vector = 2;
}

message Reference {
  ReferenceType type = 1;
  // Others
}
```

This as as json becomes:
```
ref_type = pbCommon.ReferenceType.Value("Vector")
reference = pbCommon.Reference(type=ref_type)
MessageToDict(reference)
```

```
{
        "type": "Vector"
},
```

If I run `MyMessage.verify(json)`, I get `type: enum value expected`
If I convert type to int `2`, validation works, but I would have expected string key enum to have worked since enums as JSON should be string:
https://developers.google.com/protocol-buffers/docs/proto3#json
image

ronkorving:
This fixes #1261

I haven't figured out the test suite yet, so that will follow. But in the mean time, I hope this addition could be considered.

### DIFF
--- a/src/verifier.js
+++ b/src/verifier.js
@@ -24,9 +24,12 @@ function genVerifyValue(gen, field, fieldIndex, ref) {
             ("switch(%s){", ref)
                 ("default:")
                     ("return%j", invalid(field, "enum value"));
-            for (var keys = Object.keys(field.resolvedType.values), j = 0; j < keys.length; ++j) gen
-                ("case %i:", field.resolvedType.values[keys[j]]);
-            gen
+            for (var keys = Object.keys(field.resolvedType.values), j = 0; j < keys.length; ++j) {
+                gen
+                    ("case %i:", field.resolvedType.values[keys[j]])
+                    ("case %j:", keys[j]);
+                }
+                gen
                     ("break")
             ("}");
         } else {


alexander-fenster:
Hi @ronkorving,

We believe this won't be the right thing to do. According to the diagram below, `verify` checks that the object can be passed to `encode` without any changes. This is not the case, since right now if you put a string for an enum value, `encode` will assume the integer 0 instead of the expected enum value. So this change to `verify` will actually make it return the wrong result.

image

You can use `.fromObject` that accepts enum values as strings.

ronkorving:
@alexander-fenster Thanks for your feedback, and apologies for the late response.

The problem is that `fromObject` doesn't really throw if the input is incorrect. I've played around with enum values being incorrect strings and being incorrect numbers. In both cases, the output of `fromObject` just cheerfully omits the property. No error is thrown. So how do I go about typechecking my object here using my Protobuf schema (and not a ton of custom type checks)?

GreenRover:
@alexander-fenster I face currently the same issue. 
`verify()` is expecting numbers. `fromObject()` is accepting string for an enum, but is totally silent if total bullshit was given. 
My requirement is to check if the given objects match the Message instance but accepting all variants that might be provided by `toObject()'

PR #1478, opened 2020-08-14
Fix ext fromDescriptor not handling map fields bug

dkishere:
tested with Descriptor proto generated by protoc gogo


alexander-fenster:
Hi @dkishere,

Thank you! Would it be possible to add a test that covers the added code?

Issue #1473, opened 2020-08-11
utf8.read function producing wrong strings


masad-frost:
protobuf.js version: 6.10.1

The `utf8.read` function seems to inserting extra unicode characters sometimes.

Here's a repro (https://repl.it/@masfrost/pbjs-bad-decode) where I check `utf8.read` against WHATWG `TextEncoder`. The repro file is not very minimal but this issue seems to be pretty common for us when decoding strings.

FYI for future readers, we monkey patched the library and forced it to use TextDecoder/TextEncoder here https://github.com/replit/crosis/blob/v5.0.3/src/fixUtf8.ts

I think maybe using the standard TextEncoder/Decoder might be the best thing to do here, encoding is just too complicated and I'm sure these standard libraries are faster. Happy to put up a PR if that's an option, otherwise, I don't really have enough time to go splunking into utf8 land.

daviderenger:
This is still not fixed with #1486. We got a very mysterious problem where extra � was added at almost even 8000 places sometimes in some of our messages. Turns out we only got it when we had emojis in the strings.

We went the way of using TextDecoder too and it works every time.

masad-frost:
@daviderenger do you have a test case?

maximelkin:
`@protobufjs/utf8` is not published in npm, maybe this is why fix not working?
@alexander-fenster can you please release this packages too?

Cubelrti:
Encountered this too, @protobufjs/utf8 is still v1.1.0 and not applying the patch.

Any update?

PR #1486, opened 2020-09-10
fix utf8 -> utf16 decoding bug on surrogate pairs

turbio:
This fixes https://github.com/protobufjs/protobuf.js/issues/1473

The custom utf8 -> utf16 decoder appears to be subtly flawed. From my reading it appears the chunking mechanism doesn't account for surrogate pairs at the end of a chunk causing variable size chunks. A larger chunk followed by a smaller chunk leaves behind garbage that'll be included in the latter chunk.

The chunking mechanism was added to prevent stack overflows when calling `formCharCode` with too many args. From some simple benchmarking it doesn't seem like putting utf16 code units in an array and spreading that into `fromCharCode` is helping performance much anyway. I've removed the chunking code which simplifies it significantly and fixes the particular issue we've encountered.

Here's a repro of the existing encoding bug in a fuzzing suite
https://repl.it/@turbio/oh-no-our-strings#decoder.js


lib/utf8/index.js
```diff
@@ -38,36 +38,27 @@ utf8.length = function utf8_length(string) {
...
-        chunk = [],
-        i = 0, // char offset
-        t;     // temporary
-    while (start < end) {
-        t = buffer[start++];
-        if (t < 128)
-            chunk[i++] = t;
-        else if (t > 191 && t < 224)
-            chunk[i++] = (t & 31) << 6 | buffer[start++] & 63;
-        else if (t > 239 && t < 365) {
```
turbio, line 54: My understanding of this function is that the passed in buffer is always an array of uint8s, each byte being a utf8 code unit. The condition `t < 365` is a bit odd under this assumption, from profiling it actually forces v8 to deopt when `t` is compared to a non uint8 value. Alternatively I may have grossly misunderstood how this is expected to work.

alexander-fenster:
Hi @turbio,

Would it be possible to have a test that shows the problem included in the PR?

Thank you!

turbio:
@alexander-fenster yup!
 
added a test which fails on the old `utf8_read` function.

alexander-fenster (approved):
Thanks @turbio!

jparser:
hi guys, the merged code seems not included in the latest release.
protobufjs@7.4.0 -> @protobufjs/utf8@1.1.0,  @protobufjs/utf8@1.1.0  is still the old code.

image
image

Issue #885, opened 2017-08-01
Empty buffer with "repeated byte" lead to exception 


niicojs:
protobuf.js version: 6.8.0

Using an empty buffer with Message.fromObject leads to exception on compiled JS.

```js
Message.fromObject({ bytesarray: [ Buffer.alloc(0) ] })
```

```
TypeError: Cannot read property 'length' of undefined
    at BufferWriter.write_bytes_buffer [as bytes] (c:\Users\nico\Documents\GitHub\sandbox\node_modules\protobufjs\src\writer_buffer.js:50:20)
    at Function.encode (c:\Users\nico\Documents\GitHub\sandbox\node_modules\node-pogo-protos-vnext\compiled.js:53810:77)
    at testEncode (c:\Users\nico\Documents\GitHub\sandbox\index.js:48:22)
    at Object.<anonymous> (c:\Users\nico\Documents\GitHub\sandbox\index.js:51:1)
    at Module._compile (module.js:566:14)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Function.Module.runMain (module.js:605:10)
    at startup (bootstrap_node.js:158:16)
    at bootstrap_node.js:575:3
```

nicoschmitt:
Actually the error stack may not be accurate. The thing is `fromObject()`convert `Buffer.alloc(0)` with `undefined`, which break another part of my code.
I don't think an empty `Buffer` should be converted with `undefined`

PR #1500, opened 2020-10-28
fix: empty buffers submitted to a `repeated bytes` field were converted to `undefined`

jawid-h:
I believe it is related to the [following issue (#885)](https://github.com/protobufjs/protobuf.js/issues/885). We're experiencing the same problem.

When we're having following `message` in `.proto` file:

```proto
message SomeMessage {
  repeated bytes objects = 1;
}
```

and trying to submit an empty `Buffer` in `objects` list field, the code compiled by `protobufjs` ignores it. I believe by having following statement:

```js
else if (objects[i].length) ...
```

This PR fixes that by including zero-length buffers in the check.

### DIFF
--- a/src/converter.js
+++ b/src/converter.js
@@ -71,7 +71,7 @@ function genValuePartial_fromObject(gen, field, fieldIndex, prop) {
             case "bytes": gen
                 ("if(typeof d%s===\"string\")", prop)
                     ("util.base64.decode(d%s,m%s=util.newBuffer(util.base64.length(d%s)),0)", prop, prop, prop)
-                ("else if(d%s.length)", prop)
+                ("else if(d%s.length >= 0)", prop)
                     ("m%s=d%s", prop, prop);
                 break;
             case "string": gen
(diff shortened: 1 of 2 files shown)


alexander-fenster:
@jawid-h The change looks good to us, could you add a test that would cover this situation?  Thank you!

vicb:
Hey @alexander-fenster

I would like to move this issue forward as it is causing some problems with the google datastore node client (googleapis/nodejs-datastore#755).

This is the test case I have tried to add to this PR:

```typescript
            test.test(test.name + " - Message.toObject with empty buffers", function(test) {
                var msg = Message.create({
                    bytesVal: protobuf.util.newBuffer(0),
                });

                test.equal(Message.toObject(msg, { bytes: String }).bytesVal, "", "bytes to base64 strings");

                if (protobuf.util.isNode) {
                    const bytesVal = Message.toObject(msg, { bytes: Buffer }).bytesVal; 
                    test.ok(Buffer.isBuffer(bytesVal), "bytes to buffers");
                    test.equal(bytesVal.length, 0, "empty buffer");
                }

                test.end();
            });            
```

However it seems to be causing unrelated troubles.

Would you have any hints on the best way to test this PR ?

Thanks

alexander-fenster:
Taken care of by #1514.

PR #1507, opened 2020-11-11
Fix type definition of parsedOptions

dorgalam:
Based on tests/actual code: https://github.com/protobufjs/protobuf.js/blob/master/tests/api_object.js#L27-L36

And actual results in my code using that field, the suggested type is the correct one

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -860,7 +860,7 @@ export abstract class ReflectionObject {
     public options?: { [k: string]: any };
 
     /** Parsed Options. */
-    public parsedOptions?: { [k: string]: any[] };
+    public parsedOptions?: ({ [k: string]: any | { [innerK: string]: any[] } })[];
 
     /** Unique name within its namespace. */
     public name: string;


index.d.ts
```diff
@@ -860,7 +860,7 @@ export abstract class ReflectionObject {
     public options?: { [k: string]: any };
 
     /** Parsed Options. */
-    public parsedOptions?: { [k: string]: any[] };
+    public parsedOptions?: ({ [k: string]: any | { [innerK: string]: any[] } })[];
```
alexander-fenster, line 863: I'm not sure that `any | .....` makes sense.  Also, the second part of the `|` does not look correct to me because according to the test, the value of the nested object does not need to be an array, e.g. `[{"opt1": {a: 1, b: 2}}]`.

dorgalam: Hmm, you're right on both counts, not sure why I wrote it as any[] 🤔 

I think that the best option here is then to get rid of the `|` part, and leave it as `({ [k: string]: any})[]`
Trying to describe every option with typescript would probably be more confusing than helpful

alexander-fenster: I would just leave it as `any`, yes. If there's a type for "any proto message" (like a nested object with values that can be arrays, objects, strings, numbers, booleans, nulls, etc. :) ) defined somewhere, we can use it here, but it won't be too helpful I guess.

dorgalam:
@alexander-fenster can we merge this?

alexander-fenster (approved):
It may or may not be considered a breaking change, but type changes hurt when they break something. 7.x is going out really soon so let's put it on hold for a week or two and then merge into v7.x.

PR #1526, opened 2020-12-27
fix: Plumb the forceNumber flag to the converter

lhchavez:
This change makes sure that the generated code also does not attempt to
auto-detect the presence of the `Long` library if the code is generated
with `--force-number`. With tests!


napodan:
I still have problem with initialisation. I guess you need to do something with src/reader.js.
In _configure, with this line "var fn = util.Long ? "toLong" : /* istanbul ignore next */ "toNumber";" I still have initialisation as Long.
Regards.

PR #1528, opened 2020-12-30
pbjs/sparse: fix namespace replacement, include transitive dependencies

### LINKED ISSUE #1436: sparse fails when referencing a nested enum
protobuf.js version: 6.9.0 7fd2e182150

Referencing an enum defined in another message from another file causes --sparse to fail.

main.proto
```
syntax = "proto3";
import "lib.proto";
message Foo {
        NestedEnum x = 1;
}
```
lib.proto
```
syntax = "proto3";
message Bar {
        enum NestedEnum {
                A = 0;
        }
}
```

error:
```
$ ./bin/pbjs --path . --target static --sparse main.proto
/tmp/protobuf.js/cli/pbjs.js:244
            throw err;
            ^

Error: duplicate name 'Bar' in Root
    at Root.add (/tmp/protobuf.js/src/namespace.js:239:23)
    at /tmp/protobuf.js/cli/pbjs.js:291:32
    at traverse (/tmp/protobuf.js/cli/util.js:47:5)
    at /tmp/protobuf.js/cli/util.js:62:13
    at Array.forEach (<anonymous>)
    at Object.traverse (/tmp/protobuf.js/cli/util.js:61:29)
    at sparsify (/tmp/protobuf.js/cli/pbjs.js:275:14)
    at Object.main (/tmp/protobuf.js/cli/pbjs.js:237:17)
    at Object.<anonymous> (/tmp/protobuf.js/bin/pbjs:4:16)
```

imirkin:
This fixes sparse mode for my use-case. Protos are spread over several
files, and may include nested messages/enums from other files.
Additionally, this fixes the case where a Type is replaced by a Namespace,
e.g. when an enum is used without its containing message type.

Fixes #1436

### DIFF
--- a/cli/pbjs.js
+++ b/cli/pbjs.js
@@ -251,10 +251,16 @@ exports.main = function main(args, callback) {
         if (tobj.fieldsArray)
             tobj.fieldsArray.forEach(function(fobj) {
                 fobj.referenced = true;
+                var type = fobj.resolvedType;
+                if (type && !type.referenced)
+                    markReferenced(type);
             });
         if (tobj.oneofsArray)
             tobj.oneofsArray.forEach(function(oobj) {
                 oobj.referenced = true;
+                var type = oobj.resolvedType;
+                if (type && !type.referenced)
+                    markReferenced(type);
             });
         // also mark an extension field's extended type, but not its (other) fields
         if (tobj.extensionField)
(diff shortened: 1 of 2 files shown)


alexander-fenster:
This looks reasonable, we'll just need some more time to review the change and we'll get back to you @imirkin. Thanks for the PR!

binsee:
Is there any progress on this please? @alexander-fenster

imirkin:
Is there any chance this will ever get in? Should I just close it? Rebase it?

I can't use `--sparse` without this patch, since the current logic ends up over-removing types which are actually in use.

PR #1555, opened 2021-03-07
Fixed codegen if exported type name ends with ".Object"

initial-wu:
Message named `Object` should be allowed when a `package`  name is specified.

```protobuffer
syntax = "proto3";
package foo;
message Object {
    int32 bar = 1;
}
```

```javascript
'Object.'.match(/((?<!\.)\bObject\b(?!\.))/g) // null
'.Object'.match(/((?<!\.)\bObject\b(?!\.))/g) // null
'Object'.match(/((?<!\.)\bObject\b(?!\.))/g) // ["Object"]
```


alexander-fenster:
@initial-wu Can you please include a regression test that fails without your change and passes with your change, and also fix the formatting so that lint passes? Thank you!

PR #1557, opened 2021-03-09
Use BigInt over Number for int64

mjgp2:
This PR removes the need for the `Long` package, and instead of Number uses BigInt as standard for all int64 types.

I realise that this would be a breaking change, but it seems like it would be the right design decision for the next major version of this library?

All unit tests pass - there needs to also be gulp updates to support the literal BigInt syntax, which I need some assistance with.


alexander-fenster:
Hi @mjgp2,

Thank you for this PR. We probably cannot include it in v7.0.0 (that is coming soon) but we'll think if we can drop `long` in the next semver major.  At least this would be a major pain for us at Google to convert all the client libraries (which reexport types from protobuf.js), and I can only imagine how disruptive it can be for others. Also, if you see how the `long` support is implemented, there's a configuration option that enables or disables `long`. It maybe makes sense to think if we can make this less breaking by doing it in steps:

(1) start accepting `BigInt` where we currently accept `long` when serializing;
(2) make it configurable to be able to choose if we dump `long` or `BigInt`.

Also, dropping support for Node.js v4-6-8 is a big decision.

Of course, Cc: @dcodeIO on this.

I promise we'll get back to reviewing this PR when we're done with releasing v7.0.0 :)

mjgp2:
Thanks for your reply.

Yes, I think that definitely you could support Number, BigInt, String and Long for int64 types. My thinking here was firstly to understand the codebase and to swap Long for BigInt being simpler than adding the extra branching logic, and then hopefully get some feedback.

I would suggest that it is considered whether support for Number for int64 should be dropped - there are silent overflows today I think within the codebase that are really data corruption.

ardatan:
Any progress on this PR?

mjgp2:
@ardatan - you can use the branch on my repo, it works, i.e. in your package.json

```
  "resolutions": {
    "**/protobufjs": "mjgp2/protobuf.js#bigint"
  },
```

I understand it's going to be targeted for version 8. I'm happy to help when the time comes - it's still my thinking that Number support for int64 should be dropped in version 8 because it will corrupt data.

hornta:
Sorry for pinging a bunch of people but me and my colleagues are really looking forward to this PR being merged and published so we are looking for a status on this. Is it stuck on something? Can we help out in any way?

mouafa:
Well done @mjgp2

Is there some sort of a timeline with a release date for v7 and v8?, I'm really looking forward to dropping the ~40KB long.js from our codebase.

image

goya:
just use this https://gist.github.com/goya/8185996e60725371bb30971bd641aad4

mjgp2:
@alexander-fenster - wondering if there is an ETA for v7 so I can start to work out how to take this forward?

alexander-fenster:
@mjgp2 Sorry, I don't have access to releases in this repository so I cannot help here. Your best bet would be to ask the owner, @dcodeIO.

src/util/longbits.js
```diff
@@ -36,23 +36,48 @@ function LongBits(lo, hi) {
...
 zero.zzEncode = zero.zzDecode = function() { return this; };
 zero.length = function() { return 1; };
 
+const TWO_32 = 4294967296n;
+
 /**
- * Zero hash.
- * @memberof util.LongBits
- * @type {string}
+ * Constructs new long bits from the specified number.
```
achingbrain, line 46:
```suggestion
 * Constructs new long bits from the specified bigint.
```

src/util/longbits.js
```diff
@@ -36,23 +36,48 @@ function LongBits(lo, hi) {
...
+ * @param {number} value Value
```
achingbrain, line 47:
```suggestion
 * @param {bigint} value Value
```

src/util/longbits.js
```diff
@@ -36,23 +36,48 @@ function LongBits(lo, hi) {
...
+ * @returns {util.LongBits} Instance
  */
-var zeroHash = LongBits.zeroHash = "\0\0\0\0\0\0\0\0";
+LongBits.fromBigInt = function fromNumber(value) {
```
achingbrain, line 50:
```suggestion
LongBits.fromBigInt = function fromBigInt(value) {
```

mahnunchik:
Any news?

mjgp2:
Well, v7 has been released (woo!), and so hopefully we can start to pick this up @dcodeIO ?

alexander-fenster:
Yes, we finally got v7 out - and also dropped support for older versions of Node.js (we start from 12 now). I should say I'm not sure how we properly release this if we make this change, this will be a _huge_ breaking change for _many_ users. I wonder if we could somehow use the configuration option to choose which type to use (and have this option for `pbts` as well, since the generated types are also affected).

src/util/longbits.js
```diff
@@ -36,23 +36,48 @@ function LongBits(lo, hi) {
...
+    value = BigInt(value);
+    if (value === 0n)
+        return zero;
+
+    var negative = value < 0;
+    if (negative) {
+        value = -value;
+    }
+    var hi = Number(value >> 32n) | 0;
+    var lo = Number(value - ( BigInt(hi) << 32n ) ) | 0;
```
achingbrain, line 60: I've been unable to correctly encode uint64 fields with the max field value of `18446744073709551615n` without this change:

```suggestion
    var hi = Number(value >> 32n);
    var lo = Number(value - ( BigInt(hi) << 32n ) );
```

hallahan:
I'm assuming bigint will perform significantly better since it's a native type. If we aren't going to merge this, would it make sense to fork this repo and have an npm package under a different name?

Having an object wrapping two Numbers uses significantly more memory as well, so wouldn't it be advantageous to remove Long? We could add a Long shim around bigint for backwards compatibility.

mahnunchik:
Any news?

Maligosus:
Any news?

src/util/longbits.js
binsee: I fixed it by https://github.com/juzibot/protobuf.js/pull/7

azizghuloum:
I'm not a maintainer so speaking for myself here.

Honestly I don't think this can be merged due to breaking compatibility.  

What needs to be done is:

First: *Add* support to bigints as an option (so people can migrate to it).

Then: Maybe in 5 years, drop support of Long (since no one would be using it by then, hopefully).

It's a great patch and I wish it were merged already.  I need bigints myself and don't want to start a new project with legacy junk from the start.  However, I can imagine that there are millions of lines of code affected by this change and that's why it's been 2 years since the patch was written and it's still not merged.  It breaks people's code.

Yes, the maintainers can merge it and force people to migrate or stay behind.  However, people will generally not like that.

@alexander-fenster @mjgp2 Is this kinda what the situation is?  Would this be a reasonable approach to make progress?

hallahan:
Why not just merge this and release a full version bump? That's what versions are for. Software that wants to keep using Long can stick with an older version.

Putting two Numbers (which are float64) in an Object to handle integers over 52 bits is a very inefficient solution.

goya:
i agree merge this. i vote we merge this 1 year after bigint has been released in stable JS versions.

“millions of lines of code” = code generated by this library.

mjgp2:
> i agree merge this. i vote we merge this 1 year after bigint has been released in stable JS versions.

Bigint is in all major browsers for over 2 years and since Node 10.

I think it's waiting for the owner(s) to decide how to proceed - there is a decision to be made about if it is necessary to support both Long and BigInt. I do not generate code with this library, just use it dynamically, so do not have much context on how it affects that.

phoenix741:
I wait for native bigint too

src/converter.js
```diff
@@ -239,11 +233,8 @@ converter.toObject = function toObject(mtype) {
             if (field.resolvedType instanceof Enum) gen
         ("d%s=o.enums===String?%j:%j", prop, field.resolvedType.valuesById[field.typeDefault], field.typeDefault);
             else if (field.long) gen
-        ("if(util.Long){")
-            ("var n=new util.Long(%i,%i,%j)", field.typeDefault.low, field.typeDefault.high, field.typeDefault.unsigned)
-            ("d%s=o.longs===String?n.toString():o.longs===Number?n.toNumber():n", prop)
-        ("}else")
-            ("d%s=o.longs===String?%j:%i", prop, field.typeDefault.toString(), field.typeDefault.toNumber());
+            ("var n=new util.LongBits(%i,%i,%j)", field.typeDefault.low, field.typeDefault.high, field.typeDefault.unsigned)
```
seo-rii, line 236: This code will define n for every long field, which cause failure on es6 build(`-w es6`).
You should check if n has declared in previous field.
Like this : https://github.com/seo-rii/protobuf.js/commit/1dac06f2521694c6cb49dead3e571dbb1933c07f#diff-444e086c8038faf674140e30c0597966b9a9102859bf04422d247118adedd08c

seo-rii (changes requested):

jtbandes:
I created a PR that's smaller in scope: #1908 which adds support for `toObject({ longs: BigInt })`.

jmondo:
the concerns about breaking changes seem like they are complicating this PR, but could we get some 👀  on @jtbandes's PR https://github.com/protobufjs/protobuf.js/pull/1908 ? it seems like a simple first step that would at least unblock users like myself who want to opt in to using BigInt 😄

shanghaikid:
2024, any progress?

hallahan:
😢

mahnunchik:
😢 😢 😢

cupen:
Lol.
It seems that `int64/uint64` is a big problem?

PR #1573, opened 2021-03-30
fix: util.global = global when this is undefined

ZhiruiLi:
`this` may be `undefined` in strict mode, which will cause error in the following code:

```javascript
util.Long = /* istanbul ignore next */ util.global.dcodeIO && /* istanbul ignore next */ util.global.dcodeIO.Long
         || /* istanbul ignore next */ util.global.Long
         || util.inquire("long");
```

Instead keeps the util.global as undefined, we assign user defined `global` to `util.global`. This can be very useful when using protobuf.js in some specifical platform (for example: game engine).


h3902340 (approved):
Tested in Unity with Puerts. It's OK.

PR #1575, opened 2021-04-02
deps: set @types/node to >=

indutny:
When using `protobuf.js` as a dependency in a project it is important
that `@types/node` package gets de-duped and has the same version as for
the rest of the modules in the project. Otherwise, typing conflicts
could happen as they do between v13 and v14 node types.


indutny:
I'm happy with changing it to 14 as well, but I think `*` is more future proof. Btw, I'd greatly appreciate if a release could be made for this change since otherwise I'll have to add my fork as a dependency.

alexander-fenster:
We discussed it and we think it might be a little bit more permissive than needed, would it be OK if we make it `>=13.7.0` instead? We can surely make a release with this change if needed.

indutny:
Totally! This sounds great!

alexander-fenster (approved):

indutny:
Awesome. Thanks a lot!

PR #1577, opened 2021-04-04
Add --no-service option for pbjs static target

mdouglass:
This option skips generation of service clients.


mdouglass:
We do not use the service clients for our projects (we have our own generator for those). It would be nice to be able to skip generation of the service clients to a) remove bulk from the generated file (in our code ~10K lines) and to avoid the confusion around the name clash with our own generated services.

This PR adds a --no-service option to static/static-module. The default behavior for the option is false to ensure that it is backward-compatible with existing users of the library.

Thanks!

alexander-fenster (approved):

mdouglass:
Thanks @alexander-fenster!

PR #1562, opened 2021-03-12
Change protobufjs-cli to protobufjs-cli-dbx

frederikhors:

PR #1589, opened 2021-04-16
Remove "protobufjs-cli" install command

16m:


16m:
Close the pull request, see reply in #1562 .

PR #1592, opened 2021-04-20
Support bigint as input

ardatan:
Until we completely migrate to `BigInt` with that other PR https://github.com/protobufjs/protobuf.js/pull/1557, this PR allows users to give `BigInt` as input for Int64 types.
So Int64 types will be able to consume `BigInt` in `LongBits`

This is needed for another project
https://github.com/Urigo/graphql-mesh/issues/1355#issuecomment-813866667


dcodeIO:
This has partly been superseded by the recent Long conversion support, where `fromObject` can be used to coerce BigInt values before encoding.

The change proposed here also affects the writer / direct encode path by accepting BigInt there implicitly. I do not think we should do that in 8.x, because it would conflict with strict generated typings and blur the boundary between runtime values and conversion input. This is better handled in 9.x, where BigInt support can become the actual runtime representation instead.

PR #1595, opened 2021-04-27
fix: wrong arg type to pushComment

ZhiruiLi:
`pushComment` should have array of string as it's arg, not string


alexander-fenster (approved):

ZhiruiLi:
@alexander-fenster Hi, I don't have permission to run workflow, could you please run it?
image

PR #1584, opened 2021-04-09
feat: proto3 optional support

alexander-fenster:
Adding support for proto3 `optional` fields as described in the [design](https://github.com/protocolbuffers/protobuf/blob/master/docs/implementing_proto3_presence.md).

Summary: if the proto has `syntax = "proto3";`, then for every `optional` field the parser will create a synthetic single member `oneof` field. Also, in the static code (produced by `pbjs`), proto3 `optional` fields will be set to `null` by default.

PR #1597, opened 2021-04-28
fix: fromObject should not initialize oneof members

### LINKED ISSUE #1596: fromObject breaks oneof content
protobuf.js v6.10.2

In a kind of special case when `oneof` contains both a primitive field and a message, `fromObject` that is run on the instance of the static object would break the `oneof` by populating the primitive field with its default value.

Example:

`test.proto`
```proto
syntax = "proto3";

message TargetSpend {
  int32 micros = 1;
}

message OneofContainer {
  oneof campaign_bidding_strategy {
    string bidding_strategy = 1;
    TargetSpend target_spend = 2;
  }
}
```

`repro.js`
```js
const protos = require('./test.js');

const root = protobuf.Root.fromJSON(require('./test.json'));
const OneofContainer = root.lookup('OneofContainer');

const request = new protos.OneofContainer();
const targetSpend = new protos.TargetSpend();
targetSpend.micros = 1234;
request.targetSpend = targetSpend;

console.log(request);

const object = OneofContainer.fromObject(request);

console.log(object);
```

Commands to reproduce:
```
echo {} > package.json
npm install protobufjs
npx pbjs -t static-module -o test.js test.proto
npx pbjs -t json -o test.json test.proto
node repro.js
```

Result:
```
OneofContainer { targetSpend: TargetSpend { micros: 1234 } }    // before .fromObject
OneofContainer {
  biddingStrategy: '',
  targetSpend: TargetSpend { micros: 1234 }
}   // after .fromObject - the oneof is broken
```

Note that it work properly when `fromObject` is applied to a plain object:
```js
const request1 = {
  targetSpend: {
    micros: 1234,
  }
};
const object1 = OneofContainer.fromObject(request1);
console.log(object1); // works
```

(I'm going to look into this, filing an issue just in case)

alexander-fenster:
Fixes #1596.

The problem happens when an instance of a statically generated class (e.g. generated with `pbjs`) is passed to a dynamically generated `.fromObject` (e.g. from a protobuf `Root` loaded in the runtime). Since the statically generated instance has all non-optional fields set to their default values in the prototype, including the `oneof` fields, the resulting message is incorrect because it has all non-primitive members of `oneof` assigned to the default value (e.g. empty string), which make serialized object deserialize incorrectly by official protobuf implementations (e.g. Python). The implementation of dynamic `.fromObject` has a safety check that does nothing if the `typeof` shows it's already an object of the correct type, but this safety check does not work since they technically have different types (static one vs dynamic one) in this particular case.

Real life scenario: someone creates an object using `pbjs` generated static class and `new`, then passes it to gRPC client based on proto files loaded dynamically (so the `.fromObject` it uses comes from a dynamic class).

To fix it, we need to make sure the `oneof` fields don't get initialized with anything other than `null`. This is the same logic I added previously (in #1584), so since proto3 `optional` fields are actually implemented as `oneof` members, I'm just extending the same logic to all `oneof` members, not just proto3 `optional` fields. I think it became clearer.

This PR consists of two parts. First, I'm adding a test to the static code generation (basically, `pbjs -t static`), because it's currently not covered by any test.  The second commit will bring an actual fix to the problem described above.

### DIFF
--- a/cli/targets/static.js
+++ b/cli/targets/static.js
@@ -394,8 +394,7 @@ function buildType(ref, type) {
         if (config.comments) {
             push("");
             var jsType = toJsType(field);
-            if (field.optional && !field.map && !field.repeated && field.resolvedType instanceof Type ||
-                    field.options && field.options.proto3_optional)
+            if (field.optional && !field.map && !field.repeated && field.resolvedType instanceof Type || field.partOf)
                 jsType = jsType + "|null|undefined";
             pushComment([
                 field.comment || type.name + " " + field.name + ".",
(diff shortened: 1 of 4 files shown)


alexander-fenster:
Expected test failure:
image

murgatroid99 (approved):

Issue #1572, opened 2021-03-27
undefined instead of default for optional fields ?


sloonz:
protobuf.js version: 6.10.2

Currently, to test if an optional field has been provided, you have to use `message.hasOwnProperty("optField")` because `message.optField` will be set to the default value (0 for numbers, `""` for strings) if not provided.

While this is in line with idiomatic protobuf, it is completely out of line with idiomatic javascript where unset optional field would just return `undefined` when accessed.

My suggestion is to add a pbjs option (for example `--no-optional-defaults`) for people like me that prefer the javascript way, that would set `Message.prototype.optField` to `undefined`  instead of the default value for the type.

alexander-fenster:
Are you talking about `proto2` or `proto3` `optional` fields here? They have completely different semantics, unfortunately. For `proto3` `optional`, I think I got it right in #1584 and #1597 (always assigning them to `null`). For `proto2` `optional`, I think having an option for `pbjs` might be a good idea.

sloonz:
I’m talking about proto2. I made a first shot at an option : https://github.com/sloonz/protobuf.js/commit/a00ee4180c38a9bf1f2149dc9eeed06ce6f5f368

Why is optional currently typed as `null|undefined` tho (https://github.com/protobufjs/protobuf.js/blob/master/cli/targets/static.js#L398) ? Since there's always a value defined on the prototype (even if it's null), it can't be undefined, no ?

alexander-fenster:
I checked `git blame` :) The `undefined` part was added 4 years ago in [this commit](https://github.com/protobufjs/protobuf.js/commit/c306d19d806eb697913ffa2b8613f650127a4c50), fixing #826. I don't think it's a good idea to revert it, since it will likely break TypeScript types for many users.

Your WIP commit looks good to me. I recently added a very minimal [test](https://github.com/protobufjs/protobuf.js/blob/master/tests/cli.js) for `pbjs`, so you could add a small proto file to `tests/data/cli` and use it to verify the logic in the test.

sloonz:
Added the unit tests : https://github.com/sloonz/protobuf.js/commit/a9fb6dda3e09a927bdb48f108a16c3408ce2e583

> I don't think it's a good idea to revert it, since it will likely break TypeScript types for many users.

[...]

PR #1611, opened 2021-05-06
feat: add null-defaults option

sloonz:
Discussed in #1572 : add an option to pbjs to allow default values of fields to be null instead of the zero value for the type (0 for number, "" for strings)

### DIFF
--- a/tests/data/cli/null-defaults.proto
+++ b/tests/data/cli/null-defaults.proto
@@ -1,0 +1,11 @@
+syntax = "proto2";
+
+message OptionalFields {
+  message SubMessage {
+    required string a = 1;
+  }
+
+  optional SubMessage a = 1;
+  optional string b = 2;
+  optional uint32 c = 3;
+}
--- a/cli/pbjs.js
+++ b/cli/pbjs.js
@@ -41,7 +41,7 @@ exports.main = function main(args, callback) {
             "force-message": "strict-message"
         },
         string: [ "target", "out", "path", "wrap", "dependency", "root", "lint" ],
-        boolean: [ "create", "encode", "decode", "verify", "convert", "delimited", "beautify", "comments", "service", "es6", "sparse", "keep-case", "force-long", "force-number", "force-enum-string", "force-message" ],
+        boolean: [ "create", "encode", "decode", "verify", "convert", "delimited", "beautify", "comments", "service", "es6", "sparse", "keep-case", "force-long", "force-number", "force-enum-string", "force-message", "null-defaults" ],
         default: {
             target: "json",
             create: true,
@@ -59,7 +59,8 @@ exports.main = function main(args, callback) {
             "force-long": false,
             "force-number": false,
             "force-enum-string": false,
-            "force-message": false
+            "force-message": false,
+            "null-defaults": false,
         }
     });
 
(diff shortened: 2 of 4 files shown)


tests/data/cli/null-defaults.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto2";
```
castarco, line 1: would it be only for `proto2`?
does it make sense to test it against `proto3` too?

castarco:
As far as I can see, this PR only targets the code generation features, but does not take into account Typescript decorators, right?

tests/data/cli/null-defaults.proto
sloonz: `--null-defaults` is already the default behavior for proto3 according to @alexander-fenster  : 

> For proto3 optional, I think I got it right in #1584 and #1597 (always assigning them to null).

sloonz:
The typings (JS decorators/TS) already declares that optional field can be null/undefined, so `--null-defaults` actually ends up making existing typings more correct.

I’m not sure what you mean by "dynamic loading features", but yes this PR only concerns pbjs, not stuff like reflection/custom classes.

tests/data/cli/null-defaults.proto
castarco: Actually it's not exactly how it works. Internally missing values might be represented by undefined or null, but the getters pervert his behavior, returning non-null defaults.

castarco:
> The typings (JS decorators/TS) already declares that optional field can be null/undefined, so `--null-defaults` actually ends up making existing typings more correct.
> 
> I’m not sure what you mean by "dynamic loading features", but yes this PR only concerns pbjs, not stuff like reflection/custom classes.

https://github.com/protobufjs/protobuf.js/pull/1611#discussion_r627737654

tests/data/cli/null-defaults.proto
alexander-fenster: Thanks for spotting this @castarco. We mostly use static objects (generated by `pbjs -t static-module`) and the default values are set to `null` there for optional fields; I haven't seen any bad consequences of the `field.js` code but I guess we just don't have this use case in our code base. I'll look into this, I believe it should not be hard to fix it.

alexander-fenster: This is what I was talking about (`proto3` optional fields are internally `oneof` members so get `null` as their default values):

https://github.com/protobufjs/protobuf.js/blob/master/cli/targets/static.js#L414

castarco: I'm maintaining a code generator ( https://github.com/join-com/protoc-gen-ts/ , quite opinionated, for the specific uses of my company, although it would be usable for others too ), and I had to implement a fix today because of this:
https://github.com/join-com/protoc-gen-ts/pull/37

alexander-fenster: _today I learned: you need to link to the specific commit SHA to make GitHub magic work :)_

castarco: Is this how it is supposed to work? To always use `oneOf` fields if we want truly nullable fields in proto3? 😢

alexander-fenster: Yes, this is how this is supposed to work in proto3. Original design: https://github.com/protocolbuffers/protobuf/blob/master/docs/implementing_proto3_presence.md - I just implemented it as is.

The thing is, message fields are always optional. The main reason of having optional primitive fields is to distinguish between integer 0 and "unset" value, same for strings and booleans. Putting them into a hidden `oneof` is apparently the easiest non-breaking solution.

alexander-fenster: `optional` fields are supported in proto3 starting from `protoc` 3.12 (under the flag) and IIRC 3.15 without the flag.

alexander-fenster (approved):

PR #1621, opened 2021-06-10
Fixes #1230, adding ES Module output

### LINKED ISSUE #1230: protobufjs library itself doesn't provide module support
protobuf.js version: 6.8.8
example: https://github.com/gertcuykens/protobuf

Generated modules looks ok using 
```
pbjs -p . task.proto -t static-module -w es6 -o task.pb.mjs
```

```js
import protobuf from 'protobufjs'
import {Task} from './task/task.pb.mjs'
```

But it seems the protobufjs library itself doesn't provide module support?
```
node --experimental-modules task_test.mjs 
(node:1831) ExperimentalWarning: The ESM module loader is experimental.
internal/modules/esm/default_resolve.js:69
  let url = moduleWrapResolve(specifier, parentURL);
            ^

Error: Cannot find module '/Users/gert/go/src/github.com/gertcuykens/protobuf/node_modules/protobufjs/minimal' imported from /Users/gert/go/src/github.com/gertcuykens/protobuf/task/task.pb.mjs
    at Loader.resolve [as _resolve] (internal/modules/esm/default_resolve.js:69:13)
    at Loader.resolve (internal/modules/esm/loader.js:70:33)
    at Loader.getModuleJob (internal/modules/esm/loader.js:143:40)
    at ModuleWrap.<anonymous> (internal/modules/esm/module_job.js:43:40)
    at link (internal/modules/esm/module_job.js:42:36)
gert@gert ~/go/src/github.com/gertcuykens/protobuf:master>
```

benjamind:
I took a stab at adding an ES Module build output using a Rollup build process similar to the existing browserify one.

I think this should work but will likely need some further testing to be sure as I haven't yet used this in a project.

While this PR does not touch any of the existing build outputs, it does add an `exports` field to the `package.json` to map the imports to either the existing CJS or the new ESM builds. This change can be a breaking change since it effectively isolates the interface to the package in modern resolvers, I believe I have however mapped all the existing intended exports of this package using this field but should probably get confirmation of this also.


index.d.ts
```diff
@@ -300,9 +300,6 @@ export class FieldBase extends ReflectionObject {
      */
     constructor(name: string, id: number, type: string, rule?: (string|{ [k: string]: any }), extend?: (string|{ [k: string]: any }), options?: { [k: string]: any }, comment?: string);
 
-    /** Field rule, if any. */
-    public rule?: string;
```
alexander-fenster, line 304: Not sure why the .d.ts file is changed in this PR, is it intended?

benjamind: Seems to be a consequence of running `build` again, I suspect its been changed in some other commit but the project hasn't been fully rebuilt since?

vasco-santos:
@bcoe @alexander-fenster what is the state of this PR? Do you intend to land this?

PR #1636, opened 2021-07-21
fix(static.js) Fix auto-generated code issue 1477

VagnerNico:
With these changes, issue 1477 is fixed, and all tests keep passing.


oliverrahner:
This is pretty much a duplicate of #1417, so would also fix #1383.
I'm not sure whether this one is more or less robust than #1417.

PR #1638, opened 2021-07-29
Add new StreamingDecoder class for resumable decoding

Azmisov:
This adds a new `MyMessage.StreamingDecoder[Delimited]` method, which creates a `StreamingDecoder` object. This can be used to decode messages in chunks, which is useful when you have very large serialized data that you want to decode and process as it comes in from network or disk.

Along with the chunked decoding, you can pass in an `onmessage` callback, allowing you to process messages yourself as they come in, in a sort of simple SAX-style manner. Via this callback, you can choose to exclude a message from being inserted into the root message's DOM tree. This is useful for large data, where you don't necessarily want to load everything into RAM before you start processing it.

Simple usage is like so:
```javascript
function onmessage(msg, stack){
    // process msg
    return true;
}
var decoder = MyMessage.StreamingDecoder(onmessage)
decoder.write(chunk1)
decoder.write(chunk2)
decoder.end()
console.log(decoder.root)
```
See `streaming-decoder.js` file for a working example.

### Implementation details

Rather than write an entirely new byte-by-byte parser that uses a state machine, I figured it would be best to try and reuse as much of the existing code as possible, and make minimal changes to the existing codebase. The new decoder is a generator function which will pause decoding when the current data has been exhausted, and resume when the user provides more data (via `write`/`end` methods). There are two new files which most of the code resides:
1. `reader_streaming.js`: Implements a `StreamingReader` class, which mimics the `Reader` class API. It replaces all the methods with generator functions, which can pause when more data is needed. Internally, it keeps `Reader` objects for the current user input, and possibly next user input. It will try to call the underlying `Reader` method when possible to do all the parsing. One complicated thing is that when transitioning between the current and next data inputs, it may create a temporary buffer so it has enough bytes available to read the next token. Additonally and unfortunately, the bytes, string, skip, and skipType methods couldn't easily be made to reuse the `Reader` class implementations, so there is a bit of code duplication between `Reader` and `StreamingReader`.
2. `decoder_streaming.js`: This is very very similar to `decoder.js`, except coded as a generator function so it can use the `SreamingReader` class. Besides some cleanup I did, it is pretty close to `decoder.js`, so I think the two functions could be merged to conditionally generate a streaming decoder vs traditional decoder from the same method. I left them separate since I wanted these changes to be non-invasive at first.

I didn't do a whole lot of testing, if anyone wants to pitch in on that front; I can help fix bugs if you have a test case. Honestly, I am not really using this library anyways 😄  so don't think I'll have a lot of time to invest to get this into a mergeable state, whatever steps that entails (if the maintainers don't reject this outright). So if anyone want to pick this up and run with it, feel free!

### DIFF
--- a/examples/streaming-decoder.js
+++ b/examples/streaming-decoder.js
@@ -1,0 +1,67 @@
+// proof of concept for streaming decoding message-by-message
+var protobuf = require("..");
+const { utf8 } = require("../src/util");
+
+// test protocol
+var proto = `
+syntax="proto3";
+message RootMessage {
+	uint32 field = 1;
+	repeated SubMessage msg_list = 2;
+}
+message SubMessage {
+	float field = 1;
+}
+`;
+
+var root = protobuf.parse(proto, {keepCase:true}).root,
+	RootMessage = root.lookupType("RootMessage");
+
+// generate a test example to decode
+var msg = {field:99, msg_list:[]};
+for (var i=0; i<30; i++)
+	msg.msg_list.push({
+		field: Math.random()
+	});
+var msg_str = RootMessage.encodeDelimited(msg).finish();
+
+// NORMAL DECODE
+console.log("One-shot decoding:")
+var msg_decoded = RootMessage.decodeDelimited(msg_str);
+console.log(msg_decoded);
+
+// STREAMING DECODE >>>>>
+console.log("\nStreamed decoding:")
+
+// simulated mesage chunks, as would come from network/disk
+function chunkify(str, num_chunks){
+	str = new Uint8Array(str);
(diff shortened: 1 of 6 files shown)


Azmisov:
May be better in the `onmessage` callback `stack` argument, to pass in the actual `Field` object for `field`, rather than the two props `{field:str, repeated:bool}`. But I suppose at runtime, it may not have access to the `Field` objects (?)

Also there's the question of what to do about map fields. Should there be more props in the stack entry `{map: bool, map_key: any}`? And I'm not super familiar with the protobuf syntax, but is there any other information about the field that should be included in that stack entry object?

PR #1651, opened 2021-09-15
Add protective block to map case in decoder, fixes multiple maps

### LINKED ISSUE #1650: multiple map fields in the same message type yield a SyntaxError in the generated decoder function
protobuf.js version: 6.11.2

If there are multiple maps defined in a message, the generated decoder gives an error. This is because the same variable is defined on the same switch statement level under multiple cases .

```
message Typeful {
  map<string, string> mapStringStringField=21;
  map<int32, string> mapInt32StringField=22;
}
```

```
SyntaxError: [deleted...]proto.js: Identifier 'end2' has already been declared. (1573:24)

      1571 |                     if (message.mapInt32StringField === $util.emptyObject)
      1572 |                         message.mapInt32StringField = {};
    > 1573 |                     let end2 = reader.uint32() + reader.pos;
           |                         ^
      1574 |                     key = 0;
      1575 |                     value = "";
      1576 |                     while (reader.pos < end2) {
```

I have a working fix for this, which I will submit as a PR.

reebalazs:
fixes #1650

### DIFF
--- a/src/decoder.js
+++ b/src/decoder.js
@@ -37,6 +37,7 @@ function decoder(mtype) {
 
         // Map fields
         if (field.map) { gen
+                ("{")
                 ("if(%s===util.emptyObject)", ref)
                     ("%s={}", ref)
                 ("var c2 = r.uint32()+r.pos");
@@ -75,7 +76,7 @@ function decoder(mtype) {
                 ("%s[typeof k===\"object\"?util.longToHash(k):k]=value", ref);
             else gen
                 ("%s[k]=value", ref);
-
+            gen ("}");
         // Repeated fields
         } else if (field.repeated) { gen
 


reebalazs:
Unfortunately I did not / cannot add a test for this. However: it should be simple - Every message type that has multiple maps in it should have this problem. Since there are multiple such proto definitions in the tests, maybe the generated encode/decode functions are never exercised via tests? If they are, the problem should have popped up and the fix could be provable through tests.

hsjoberg:
Looks like both this one and #1455 addresses this issue.

xWTF:
This fixes what I met too, pbts complaints about `Identifier 'end2' has already been declared.` when there're multiple maps and I have to add `{}` manually every time there's a change in proto files...
Hope it get merged soon, since #1455 was created TWO YEARS AGO and is still open today :(

hsjoberg:
> Hope it get merged soon, since #1455 was created TWO YEARS AGO and is still open today :(

@xWTF Very sad state of affairs, I chose to roll with #1455 and it has been working without any issues whatsoever:

https://github.com/hsjoberg/blixt-wallet/blob/99c50afb3fc2119d8e531c4b8cf582df829a9b4f/package.json#L74

reebalazs:
What I don't understand. Not having any of these fixes means that maps are plain broken. Are people not using this package? (Or maybe they are using some fork where this is fixed?) Or are users not using maps? In any case I'd expect this to be clarified by the package owners.

hsjoberg:
@reebalazs Yes, this error (or family of errors, this PR only touches one of them...) does not always appear for everyone, I was rolling fine without any patch until my upstream project updated their protobuf file which caused the issue to appear to me.

Malonrio (approved):
I met same problem

JrSchild (approved):

PR #1653, opened 2021-09-17
Fix encoding of int64/bool keys of maps in encoder

### LINKED ISSUE #1652: map field with int64 type has broken key in generated encode function
protobuf.js version: 6.11.2

The generated encode function for maps with an int64 key is broken.

The int64 writer cannot handle the hash directly, because it interprets
it as a string representation of a Long number, instead of a hash.

The result: there is no error, but the encoded message contains a map with wrong keys. 

For example, the following message type exhibits the problem:

```

message Typeful {
 map<int64, string> mapInt64StringField=23;
}

```

Consider the following field value as an example. This has long keys 1 and 2. The following  object  is displayed with the longbit hash keys.

```
{
       ​'\x01\x00\x00\x00\x00\x00\x00\x00': 'foo',
       ​'\x02\x00\x00\x00\x00\x00\x00\x00': 'bar'
}
```

Because of the error in the encoder, the keys will all be converted with an error to the same  key containing 8 zero bytes. As a result, if we encode this message and decode it back, the following value is received. As a side effect, the values are also collapsed, since all keys are zero in the case shown above.

```
{
   ​​'\x00\x00\x00\x00\x00\x00\x00\x00': 'bar'
}
```

I  am  providing a PR that fixes the issue.

reebalazs:
Fixes #1652 .

The int64 writer cannot handle the hash directly, because it interprets
it as a string representation of a Long number, instead of a hash.

This is obviously wrong, because the input at this point always contains the longbit hashes already. So the int64() writer fails inevitably.

I believe that the simplest solution is: Explicit conversion from hash to long is needed before writing.

Again (just like with the other PR I submitted) I believe that the tests are missing for the generated encode/decode functions. If there were such tests, it would be fairly straightforward to write a test case that exercises this problem, and proves that the fix, in fact, works.

### DIFF
--- a/src/encoder.js
+++ b/src/encoder.js
@@ -48,6 +48,10 @@ function encoder(mtype) {
             let keyGetter;
             switch (field.keyType) {
                 case "int64":
+                case "uint64":
+                case "sint64":
+                case "fixed64":
+                case "sfixed64":
                     keyGetter = "$util.longFromHash(ks[i], false)";
                     break;
                 case "bool":
--- a/src/encoder.js
+++ b/src/encoder.js
@@ -45,10 +45,22 @@ function encoder(mtype) {
 
         // Map fields
         if (field.map) {
+            let keyGetter;
+            switch (field.keyType) {
+                case "int64":
+                    keyGetter = "$util.longFromHash(ks[i], false)";
+                    break;
+                case "bool":
+                    keyGetter = "ks[i] === \"true\"";
+                    break;
+                default:
+                    keyGetter = "ks[i]";
+                    break;
+            }
             gen
     ("if(%s!=null&&Object.hasOwnProperty.call(m,%j)){", ref, field.name) // !== undefined && !== null
         ("for(var ks=Object.keys(%s),i=0;i<ks.length;++i){", ref)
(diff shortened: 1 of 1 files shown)


reebalazs:
Exactly same case with a bool key type. Fix extended to cover this case as well.

src/encoder.js
```diff
@@ -45,10 +45,22 @@ function encoder(mtype) {
 
         // Map fields
         if (field.map) {
+            let keyGetter;
+            switch (field.keyType) {
+                case "int64":
```
angadn, line 50: Probably also makes sense to extend this for `uint64`?

angadn (approved):

src/encoder.js
reebalazs: I think you are right. Likely also for `sint64`, `fixed64`, `sfixed64` as well. Would you like me to extend the PR with these?

angadn: That would be amazing, yeah! Thanks so much! Do you think we'd see this merged to `master` at some point? (Cc: @dcodeIO)

reebalazs: My pleasure! :) I've added this.

Imo it should be merged to master, as it is definitely broken without this fix.

Issue #1560, opened 2021-03-12
[question] How to support `@deprecated` fields?


mreiche:
protobuf.js version: 6.10.2

I have some `deprecated` fields like

```protobuf
message MyMessage {
    string my_field = 6 [deprecated=true];
}
```

But no matter what I try...

```protobuf
message MyMessage {
    string my_field = 6 [deprecated=true]; // @deprecated
}
```

```protobuf
message MyMessage {
    /**
    * @deprecated
    */
    string my_field = 6 [deprecated=true];
}
```

```protobuf
message MyMessage {
    // @deprecated
    string my_field = 6 [deprecated=true];
}
```

When I run the command:
```shell
pbjs --no-create --no-verify --no-convert --no-delimited --no-beautify --force-number -t static-module --es6 -w es6 -o bundle.js
pbts -o bundle.d.ts bundle.js
```
The field is not marked as `@deprecated` in the `bundle.ts`

```ts
/** undefined */
public myField?: (string|null);
```

but in the `bundle.js`

```js
/**
  * @deprecated
  * @member {string|null|undefined} myField
  * @memberof MyMessage
  * @instance
  */
  MyMessage.prototype.myField = null;
```

Zsar:
Just bit me again today in v7.1.2 . Would be great if something could be done against it.

PR #1666, opened 2021-11-17
keep deprecated tag

sherlock1982:
Some entities in proto file can be marked as deprecated.
Let's keep deprecated tag for users to see it.


mreiche:
May close #1560

kcirtaptrick:
What's blocking this from getting in? This would make surfacing deprecations very nice in editors that show them, and the change seems simple enough.

mreiche (approved):

sherlock1982:
Just a gentle reminder on this - was patching it for years.

PR #1687, opened 2022-01-31
fix: support for nested messages and enums within group blocks

JoshuaWise:
This fixes support for proto files that have nested messages or enums declared within groups, like this:

```proto
syntax = "proto2";

message Foo {
  repeated group Field = 1 {
    message Bar {
      optional bool my_bool = 2;
    }

    optional string my_string = 3;
    optional Bar my_bar = 4;
  }
}

```

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -443,6 +443,14 @@ function parse(source, root, options) {
                     }
                     break;
 
+                case "message":
+                    parseType(type, token);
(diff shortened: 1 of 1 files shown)


JoshuaWise:
Closed in favor of https://github.com/protobufjs/protobuf.js/pull/1790

PR #1704, opened 2022-03-31
fix(deps): patch minimatch vulnerability

richgerrard:
If I follow this, glob packages minimatch.  Minimatch released a fix, glob also has a newer build, picking this up should pick up that.
Fixes #1696
Fixes #1697
Fixes #1698

PR #1705, opened 2022-04-07
Get the package into a working state again

dcodeIO:
While proper modernization seems due, this PR for now tries to get all the things working again.

* Updates dependencies but doesn't go to lengths in regards of breaking dependency changes yet.
* Continued work towards separation of the `protobufjs` and `protobufjs-cli` packages.
* Removes `dist/` files from source control.
* Updates CI scripts, drops testing on Node.js < 12 and sets up a new automated publishing step.


Qard:
Can you do something about #1704 while you're in there?

dcodeIO:
I think this should be fixed by the dependencies update. The referenced PR doesn't do this I believe, since if package-lock is honored, there is no change as it hasn't been updated, and if it is not, the most recent `^7` will be installed anyway.

Qard:
Not updating the bottom end of a caret range in package.json means lower numbers are still considered valid. If npm tries to dedupe and flatten dependencies it may pick an older version to be in-range of all depending modules. Generally the package.json should be updated in the case of security vulnerabilities to explicitly block npm from allowing the module to depend on vulnerable versions of the dependency.

dcodeIO:
TIL, thanks :)

richgerrard (approved):

craigmarker:
Anything we can do to help push this along? I'm interested in getting access to https://github.com/protobufjs/protobuf.js/commit/af1b449602b360091e191a58abde2f246d8b0f1d which fixes an option parsing error I'm experiencing. I'd also be interested in providing a fix to parse EnumValueOptions: https://github.com/protobufjs/protobuf.js/issues/1631

Thanks!

PR #1726, opened 2022-05-16
fix: throw error on missing import (#1725)

### LINKED ISSUE #1725: Missing import with loadSync is not thrown
protobuf.js version: 6.10.2

```
syntax = "proto3";

import "no/such/file.proto";

message Something {
  ImportedType field = 1;
}
```

```js
root.loadSync("missing-import.proto")
```

This should throw an error if `"no/such/file.proto"` does not exist but it does not.

steinybot:
Fixes #1725


src/root.js
```diff
@@ -94,14 +94,21 @@ Root.prototype.load = function load(filename, options, callback) {
     var sync = callback === SYNC; // undocumented
 
     // Finishes loading by calling the callback (exactly once)
+    var throwing = null;
     function finish(err, root) {
+        // If we get here, and we are throwing then something must have
+        // incorrectly caught the error, so we need to rethrow it.
+        if (throwing)
+            throw err;
```
alexander-fenster, line 102: Just to double check, are we sure that `err` is set at this point? or maybe `throw throwing;` since it's where you actually saved the error that was thrown?

steinybot: That is a good point. It worked in the case that I tested but it seems wrong or at least prone to breaking. I'll change it.

PR #1742, opened 2022-06-11
fix: update minimal.js to evade override mistake

erights:
See https://github.com/Agoric/agoric-sdk/blob/master/patches/%40confio%2Bics23%2B%2Bprotobufjs%2B6.11.3.patch

The original code used assignment to override the `constructor` and `toString` properties inherited from Error.prototype. However, if `Error.prototype` is frozen, as it is under Hardened JS (aka SES) or under the Node frozen intrinsics flag, then this assignment fails due to the JavaScript "override mistake".

`enumerable: true` would accurately preserve the behavior of the original assignment, but I'm guessing that was not intentional. For an actual error subclass, this property would not be enumerable, so my PR currently proposes that. But either would work, so let me know if you'd like me to change it.

`configurable: false` would accurately preserve the behavior of the original, but I'm guessing that was not intentional. For an actual error subclass, this property would be configurable. But either would work, so let me know if you'd like me to change it.


alexander-fenster (approved):

erights:
Thanks!

PR #1753, opened 2022-06-22
fix: proper relative path to protobufjs in cli

maitbayev:
According to https://docs.npmjs.com/cli/v8/configuring-npm/package-json#local-paths the local path should be `file:..` instead of `file://..`.

The later version copies whole top-level file-system on a macOS and with a recent version of npm (8.12.2). As a result , it errors with message `Cannot find module 'protobufjs'`


maitbayev:
cc @bcoe @dcodeIO

bcoe (approved):

alexander-fenster:
Going to merge this as soon as I figure out the CI problem (doing it now).

PR #1762, opened 2022-07-07
feat: allow message.getTypeUrl provide custom tyepUrlPrefix

kairlec:
I noticed that in many other language implementations (such as [Java](https://github.com/protocolbuffers/protobuf/blob/f367bfb13d9ff5f097666bcdf7389aa8cef902b7/src/google/protobuf/compiler/java/java_message.cc#L1683-L1690), [CSharp](https://github.com/protocolbuffers/protobuf/blob/3f5fc4df1de8e12b2235c3006593e22d6993c3f5/csharp/src/Google.Protobuf/WellKnownTypes/AnyPartial.cs#L136-L145)), Google provides optional parameter `typeUrlPrefix` for the function `pack` or `getTypeUrl`

---

this pr is a re-request of #1665 , the original pr was mistakenly overwritten by the force push of the pull bot

### DIFF
--- a/cli/targets/static.js
+++ b/cli/targets/static.js
@@ -597,7 +597,7 @@ function buildType(ref, type) {
             "@function getTypeUrl",
             "@memberof " + exportName(type),
             "@static",
-            "@param {string|undefined} [typeUrlPrefix] your custom typeUrlPrefix(default \"type.googleapis.com\")",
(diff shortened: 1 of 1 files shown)


alexander-fenster (approved):
Thank you! Now that we have some test coverage for CLI tools, I added a test (and regenerated part of the test files which was generated, so this added some extra files to this PR).

PR #1776, opened 2022-07-18
fix(types): update type deps

thw0rted:
Remove duplicate `long` types

Long has shipped with its own TS types for a while now, and the package from DefinitelyTyped conflicts with the newest version.  TS consumers wind up with errors when trying to work with `Long` values.

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -51,7 +51,6 @@
         "@protobufjs/path": "^1.1.2",
         "@protobufjs/pool": "^1.1.0",
         "@protobufjs/utf8": "^1.1.0",
-        "@types/long": "^4.0.1",
         "@types/node": ">=13.7.0",
         "long": "^5.0.0"
       },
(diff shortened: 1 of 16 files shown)


alexander-fenster (approved):

alexander-fenster:
@thw0rted Would you mind fixing the failing tests in your branch? Thank you!

thw0rted:
I think it's fixed now.  I updated the test-runner script to do the whole directory at once, rather than 3 separate invocations of `tsc`.  That should be enough to close out this PR.

I did also try to take the "minimal" definition of a `Long` interface out of `util/common.js` but vanilla JSDoc still doesn't have a way to import a type definition (https://github.com/jsdoc/jsdoc/issues/1537, https://github.com/jsdoc/jsdoc/issues/1645) so I think we're stuck with it for now.

Just as an aside, this is like the 4th issue in the last week where trying to stick with vanilla JSDoc, instead of using TypeScript's JSDoc parser, has made my life difficult.  JSDoc is a dead standard, they haven't published a meaningful update since 2017, and it's really holding the whole ecosystem back.  Specifically, ESM support has come a long way in 5 years, and JSDoc still assumes a CJS / global-pollution model for types... and we have no reason to expect them to rectify that situation.  Not sure what PBJS can do about that, I guess, but it's something to think about.

PR #1777, opened 2022-07-19
Fix blocker jsdoc bugs

buu700:
These are major happy path bugs that have been in the stable release for a number of years now. The root cause seems to be on jsdoc's end, but the ticket remains unaddressed.

At this point, I think it makes sense to just stick with 3.5.5 indefinitely, at least until jsdoc is fixed and updated to provide a cleaner interface.


buu700:
I'm not really sure. This is the fix that worked for me, but I'm not familiar enough with protobufjs or jsdoc to say what it would take to work around the issue without pinning.

This has been broken for several years now, but the jsdoc team hasn't been responding to the issue I'd submitted (https://github.com/jsdoc/jsdoc/issues/1644).

buu700:
Closed because it's superseded by #1844.

PR #1785, opened 2022-08-10
fix: remove unused `@types/long`

SimenB:
#1751 updated to v5 which ships with types. However, the `Long` type used here seems to be custom and not imported, so this was never used anyways?

https://github.com/protobufjs/protobuf.js/blob/6f0806ddef408647fbaed049dbd8929ad9f5c10c/index.d.ts#L1848-L1862

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -51,7 +51,6 @@
         "@protobufjs/path": "^1.1.2",
         "@protobufjs/pool": "^1.1.0",
         "@protobufjs/utf8": "^1.1.0",
-        "@types/long": "^4.0.1",
         "@types/node": ">=13.7.0",
         "long": "^5.0.0"
       },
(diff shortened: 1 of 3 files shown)


zmwangx:
Just encountered the same problem. `@types/long@4.0.1` is causing an additional `Long` type that's incompatible with the builtin `Long` type from `long@5.0.0`. Had to add `import Long from "long";` to generated `proto.d.ts` to prevent it from picking up the wrong `Long` from `@types/long`.

alexander-fenster (approved):

alexander-fenster:
Well, this happens to be harder than I thought - adding this actually breaks some code that depended on `import * as Long from 'long'` being possible, so it's a breaking change. I will need to roll it back now and think how we could proceed here.

alexander-fenster:
Maybe not rollback, since we actually have more packages doing the same thing with `long`, but I'll take some time before releasing that to make sure other code I care about does not break because of this. I expect to release this some time next week.

PR #1800, opened 2022-09-08
fix: fix the unexpected case when listener be triggered

C-Dao:
Fix the unexpected case when listener be triggered. When I add two listeners of the same type to the event-emitter, and then I remove the first listener in the first listener's handler,  the second listener cannot be executed correctly.

### DIFF
--- a/lib/eventemitter/index.js
+++ b/lib/eventemitter/index.js
@@ -74,11 +74,12 @@ EventEmitter.prototype.emit = function emit(evt) {
     var listeners = this._listeners[evt];
     if (listeners) {
         var args = [],
-            i = 1;
-        for (; i < arguments.length;)
-            args.push(arguments[i++]);
-        for (i = 0; i < listeners.length;)
-            listeners[i].fn.apply(listeners[i++].ctx, args);
+          i = 1,
+          _listeners = listeners.slice();
+
+        for (; i < arguments.length; ) args.push(arguments[i++]);
+        for (i = 0; i < _listeners.length; )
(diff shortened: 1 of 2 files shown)


alexander-fenster:
If I'm reading the code correctly, it requires copying the list of listeners every time an event is emitted, which is probably a lot of extra work for just a simple use case. I'm guessing the workaround could be to postpone the invocation of `.off()` to the next event loop iteration:

```js
ee.on("event", () => {
  ...
  setImmediate(() => { ee.off(func); });
});
```

C-Dao:
> If I'm reading the code correctly, it requires copying the list of listeners every time an event is emitted, which is probably a lot of extra work for just a simple use case. I'm guessing the workaround could be to postpone the invocation of `.off()` to the next event loop iteration:
> 
> ```js
> ee.on("event", () => {
>   ...
>   setImmediate(() => { ee.off(func); });
> });
> ```

Hi，thanks you take time to review my PR 😊.

Using `setImediate` is effective, but I think it is not a good way. Because sometimes we forget to use `setImmediate` to avoid this situation.

`slice()` method is shallow copy the array, the cost is acceptable 😃.

Moreover, the invocation of `emit`, `off` and `on`, they should not affect the execution of the previous invocation. In another case, if we listen to it self again in each `on` handler, the listener added later will also be affected by the last emit and triggered, which is inappropriate and will cause the infinite loop. 

Show code like below
```
  let fn = function () {
    /* ... do something */
    ee.on("b", fn);
  };
  ee.on("b", fn, ctx);
```

PR #1815, opened 2022-09-23
pbjs/sparse: fix namespace replacement, include transitive dependencies

binsee:
This fixes sparse mode for my use-case. Protos are spread over several files, and may include nested messages/enums from other files. Additionally, this fixes the case where a Type is replaced by a Namespace, e.g. when an enum is used without its containing message type.

from https://github.com/protobufjs/protobuf.js/pull/1528


binsee:
Sorry, this is a bad submission, wrong operation.

Issue #564, opened 2016-12-16
Have an option to use snake_case for message properties


vgoloviznin:
protobuf.js version: 6.1.1

I'm running pbjs utility to generate JSON for my proto files, but the issue is that it generates camelCase property names  (proto files have snake_case). 

Is there already some configuration for this (found a broken link to wiki article) or is it possible to include one?

dcodeIO:
Currently, v6 always converts to camel case for simplicity. Google's JS implementation also does this.

Do you have a specific use case where this fails or snake_case is absolutely required?

vgoloviznin:
Can't say that I have a hard requirement for snake_case :) 
We have all documentation for communication between services where all field names are in snake case, so it's just easier to blindly follow the documentation (or reading the proto file) without thinking about camel case conversion

dcodeIO:
pbjs now supports the option `--keep-case` that skips the camel case conversion. I haven't tested this for edge cases in detail, yet, so let me know if there are any issues.

vgoloviznin:
@dcodeIO cool! When will it get released on npm?

dcodeIO:
Maybe tonight as 6.2.1 if there are no more pending issues.

vgoloviznin:
Awesome, thanks for the quick fix! 

I will check during the weekend for our definition

dcodeIO:
It's on npm now!

thirstyfish:
Hi!
Apologies if it's a silly question, I'm just familiarising myself with your library, but haven't found an answer to this in the docs, this issue is as close as I got.
I have the same problem as the OP, but I'm not using the CLI to generate a json file.
Is there an equivalent option that can be set in the proto file or elsewhere that will have the same effect (leave the snake case intact) in case I'd like to load the raw proto file directly with protobuf.load()?
Thanks!

dcodeIO:
Yes: [Root#load](http://dcode.io/protobuf.js/Root.html#load3)

```js
var root = new protobuf.Root();
root.load(..., { keepCase: true }).then(...);
```

thirstyfish:
Great, thanks for your help!

samkeke:
Thank you help,

I finally found the answer

rahulverma:
Is there a way to --keep-case for static modules. Its a big problem when using typescript.

PR #1825, opened 2022-10-12
Honor json_name option

### LINKED ISSUE #992: Using field names based on json_name annotation in generated static code
protobuf.js version: 6.8.0

I've run into an issue a few times when parsing JSON data into the statically generated classes, when the fields in the JSON data are named using using the json_name annotation for the field names instead of the normal field names.

For example, if a proto file looks like:
```protobuf
import "google/api/annotations.proto";
import "otherstuff.proto";
message MyClass {
    repeated Thing thing = 1 [json_name="my_things"];
}
```
And you are parsing in the object like
```ts
const myObject = MyClass.fromObject(someObjectParsedFromJsonData);
```
the object field currently needs to be named `thing` instead of `my_things`.

It would be nice if the alternate naming scheme could be used directly, either through a command line switch or using another method like fromJson() in place of fromObject().

### LINKED ISSUE #1245: toJSON() does not use json_name
protobuf.js version: "protobufjs": "^6.8.8",

decode message to JSON should use json_name as key. 

pbjs -t json-module ...

Generates the following

```javascript         
fields: {
              timestamp: {
                type: "uint64",
                id: 1,
                options: {
                  json_name: "t"
                }
              },
              latitude: {
                type: "double",
                id: 2,
                options: {
                  json_name: "lat"
                }
              },
              longitude: {
                type: "double",
                id: 3,
                options: {
                  json_name: "lon"
                }
              },
              altitude: {
                type: "double",
                id: 4,
                options: {
                  json_name: "alt"
                }
              },
              fields: {
                type: "uint32",
                id: 15,
                options: {
                  json_name: "f"
                }
              }
            }
```

However whenever I call toJSON() on a message that has been decoded I get strings that are not json_name values. I cannot find any options to create the object with keys reflecting these json_names.

```javascript
  const test = root.lookupType("com.<>.<>");
  const bin = readBinaryFile("./<>.bin");
  const mess = test.decodeDelimited(bin);
  console.log(mess.toJSON)
```
Logs
```javascript
{
  "timestamp": "1557957939176",
  "latitude": 32.305504145820564,
  "longitude": -64.8585479141154,
  "altitude": 65.5
}
```

### LINKED ISSUE #1775: Can't properly convert structure to proto message with fromObject method.
protobuf.js version: 7.0.0

Hello, all! 
Try to convert JS structure in proto message.
Problem is that nested structure is always empty list.
In example expected that output 3 will be like
TestCollection { collections: [{name: 'Test Sub Collection', collections: []}], name: 'Test Collection' }

```proto
message TestCollection {
  string name = 1;                                                                                         // Collection Name
  repeated TestCollection collections = 2 [json_name = "collectionsList"];  // List of nested collections
}
```

```js
const model_path = path.join(process.cwd(), 'components/integration-test/model/test_result.proto');
var root = protobuf.loadSync(model_path);
const TestCol = root.lookupType("TestCollection");

let col = new result.TestCollection()
col.setName("Test Collection")
let subCol = new result.TestCollection()
subColEx.setName("Test Sub Collection")
col.addCollections(subCol)

console.log(col)                                                     // 1
console.log(col.toObject())                                    // 2

message = TestCol.fromObject(col.toObject());
console.log(message)                                            // 3
```

```
// 1
{
  wrappers_: { '2': [ [Object] ] },
  messageId_: undefined,
  arrayIndexOffset_: -1,
  array: [ 'Test Name 1', [ [Array] ] ],
  pivot_: 1.7976931348623157e+308,
  convertedPrimitiveFields_: {}
}

// 2
{
  name: 'Test Collection',
  collectionsList: [ { name: 'Test Sub Collection', collectionsList: [] } ]
}

// 3
TestCollection { collections: [], name: 'Test Collection' }
```

michaelbgreen:
Honor the json_name option.  This is useful when a .proto file is compiled for multiple languages, to conform to whatever random casing rules were chosen for the project:
- The .proto convention is to use lower_snake_case for field names, and UPPER_SNAKE_CASE for enum names
- C++ (via protoc) simply uses lower_snake_case for field names
- C# (via protobuf-net) uses PascalCase for both field and enum names, and has [options for customizing the names](https://github.com/protobuf-net/protobuf-net/blob/main/src/Tools/protogen.proto)
- JavaScript (via protobuf.js) uses camelCase for field names, and the json_name attribute can now be used to customize the name

Example:
```
message NetworkAdapter {
  string get_ip_address = 1 [json_name = "getIPAddress", (.protobuf_net.fieldopt).name ="GetIPAddress"];
}
```

Default names without the options:
- C++: get_ip_address (unchanged)
- C#: GetIpAddress
- JS: getIpAddress

Fixes #992, fixes #1245, fixes #1775

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "protobufjs",
-  "version": "7.1.2",
+  "version": "7.1.3",
   "versionScheme": "~",
   "description": "Protocol Buffers for JavaScript (& TypeScript).",
   "author": "Daniel Wirtz <dcode+protobufjs@dcode.io>",
@@ -36,7 +36,7 @@
     "prof": "node bench/prof",
     "test": "npm run test:sources && npm run test:types",
     "test:sources": "tape -r ./lib/tape-adapter tests/*.js tests/node/*.js",
-    "test:types": "tsc tests/comp_typescript.ts --lib es2015 --esModuleInterop --strictNullChecks --experimentalDecorators --emitDecoratorMetadata && tsc tests/data/test.js.ts --lib es2015 --esModuleInterop --noEmit --strictNullChecks && tsc tests/data/*.ts --lib es2015 --esModuleInterop --noEmit --strictNullChecks",
+    "test:types": "tsc tests/comp_typescript.ts --lib es2015 --esModuleInterop --strictNullChecks --experimentalDecorators --emitDecoratorMetadata && tsc tests/data/test.js.ts --lib es2015 --esModuleInterop --noEmit --strictNullChecks && tsc --project tests/data/all_files.tsconfig.json --lib es2015 --esModuleInterop --noEmit --strictNullChecks",
     "make": "npm run lint:sources && npm run build && npm run lint:types && node ./scripts/gentests.js && npm test"
   },
   "dependencies": {
--- a/src/field.js
+++ b/src/field.js
@@ -221,6 +221,16 @@ Field.prototype.setOption = function setOption(name, value, ifNotSet) {
     return ReflectionObject.prototype.setOption.call(this, name, value, ifNotSet);
 };
 
+Field.prototype.applyParsedOptions = function applyParsedOptions() {
+    if(this.parsedOptions !== null) {
+        var opt = this.parsedOptions.find(function (opt) {
+            return Object.prototype.hasOwnProperty.call(opt, "json_name");
+        });
+        if(opt)
+            this.name = opt["json_name"];
+    }
+};
+
 /**
  * Field descriptor.
  * @interface IField
--- a/src/parse.js
+++ b/src/parse.js
@@ -388,6 +388,8 @@ function parse(source, root, options) {
             parseInlineOptions(field);
         });
 
+        field.applyParsedOptions();
+
         if (rule === "proto3_optional") {
             // for proto3 optional fields, we create a single-member Oneof to mimic "optional" behavior
             var oneof = new OneOf("_" + name);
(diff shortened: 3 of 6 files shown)


aneeshgarg:
Can we prioritize this review? We need it as well.

aneeshgarg:
Will this change also use json_name for files generated statically?

tomquist:
Any update on this? Would really like to see this land.

YashJainSC:
it seems we have a solution here for this
https://github.com/protobufjs/protobuf.js/issues/564#issuecomment-280923387

tomquist:
The linked comment is not exactly a solution as it doesn't respect the explicitly defined json_name.

mbohal:
Is there any way I could help this PR move forward and get merged?

I just did a quick test against `v7.2.5`  and it seems to be working.  I really need this functionality and I would like to to avoid having to maintain my own fork.

alexander-fenster (approved):
The code looks good to me, my main concern here is that it's technically a breaking change. The code that used the original field names will stop working if the proto defines `json_name` for those fields. I feel it's a little bit more serious than https://xkcd.com/1172/ so I would rather stay on the safe side and call it a major version.

What do you folks think if we mark it as a breaking feature, and combine it with dropping Node.js v12 support to release it as 8.0.0? I'll wait for a few days here for any feedback from the community.

aneeshgarg:
+1 to releasing it as a major version. This change will definitely break my current use case. 

Can this be behind a build flag/option?

tomquist:
Alternatively this could be a runtime option which is off by default.

aneeshgarg:
Runtime option might not work with folks like us using the statically generated files.

alexander-fenster:
I think it can technically go to `IParseOptions` (right near the `keepCase`) and to the CLI flag of `pbjs` for static proto generation. That way we can make it false by default and avoid making a semver major version. Not sure if @michaelbgreen has time to implement it that way though :)  (I might have but later this month).

PR #1840, opened 2022-12-23
feat(cli): generate static files at the granularity of proto messages

vpluschen:
I encountered a problem in the project I was working on: Due to the large size of my project and the complexity of the proto, the resulting js static code file was very large, nearly 100,000 lines before uncompression, about 5mb, and nearly 1mb even after compression , which causes my pages to load slower.

Later I found out the cause of the problem: my page only uses a few messages, but I still need to compile the entire file and all the messages, enums, etc. of all dependent files, and there is a huge waste here.

Finally, I found a solution: by adding a filter parameter to the cli, to filter those structures that we don't need at all, so that our files are greatly reduced, which will be very useful in a huge project. Eventually I cut it from 100,000 lines to 10,000.

example：
pbjs --target=static-module  --filter=/home/filter.json  --out=appsvr-source.js proto/appsvr.proto

/home/filter.json:
{ "messageName": ["mypackage.message1", "message2"]}

then, only  mypackage.message1 and message2  and their dependencies will gen in code.  

Hope this solution can help others.

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -35,7 +35,8 @@
       }
     },
     "..": {
-      "version": "7.1.0",
+      "name": "protobufjs",
+      "version": "7.1.2",
       "dev": true,
       "hasInstallScript": true,
       "license": "BSD-3-Clause",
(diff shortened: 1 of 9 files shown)


alexander-fenster (approved):

alexander-fenster:
Thank you for contributing the feature with the test! I updated the jsdoc a little bit to make the linter happy. The feature seems pretty well isolated with a CLI flag, so I hope it should be pretty safe and we can fix the issues as they arise.

Issue #1839, opened 2022-12-19
Pbjs does not use "import * as ..." for importing js files


AVKurbatov:
Versions:
protobufjs: 7.1.2,
protobufjs-cli: 1.0.2,
pbjs: 0.0.14,
ptjs: 0.2.4.

Hello!

We have a problem using protobuf-js-cli (pbjs, pbts) and protobuf-js.
In our project we use more than 100 proto files and some of them use more than 20 imports for other proto files.
Protobuf-js-cli generates all the code for importing protofiles into one.
As a result, one of the js files we generated has a size of 60 MB and 300,000 lines of code.
The pbts plugin can't handle this. It just crashes without any error messages and no d.ts file is created.
For other cases, it would be generally convenient for us to use imports in the generated js files instead of
large js files with all the imported code in them.

Could you give some advice? Can we use protobuf-js-cli to generate files without imported code in them?

AVKurbatov:
Let me explain it with the example.
I have two files: sub_module.proto:

syntax = "proto3";
package Codegen;

message HelloRequest {
  string name = 1;
}
message HelloReply {
  string message = 1;
}

and module.proto:
syntax = "proto3";
package Codegen;

import "sub_module.proto";
message SomeComposition {
  HelloRequest title = 1;
  HelloReply assets = 2;
}
service Greeter {
  rpc SayHello (HelloRequest) returns (HelloReply) {}
}

If I generate js file for this two proto files, I get file module.js with definitions for HelloRequest and HelloReply.
I use command line "pbjs -t static-module -p protoFolder -w commonjs -o module.js module.proto"
But I want to import sub_module into module.
I see it in this way. 
1) We need to add imports to the head of the generated file:
![image](https://user-images.githubusercontent.com/31739697/208418231-c0f52566-d0ae-4387-951e-df2a76a69b11.png)
2) We need to add the namespace to names for all classes described in the sub_module:
![image](https://user-images.githubusercontent.com/31739697/208418462-2b1ebcd0-959c-4ce1-88cc-2d7f03d0310e.png)
3) And finally we need to remove definitions of all classes that are defined in the sub_module:
![image](https://user-images.githubusercontent.com/31739697/208418632-d63ec9b4-e4cd-407c-a1b1-22603e3417e7.png)


I'm going to update pbjs and pbts code according to this ideas.
What do you think about this plan? Are there any critical flaws in it?
Do we have any easier ways to solve this problem?

[...]

PR #1841, opened 2022-12-23
Generate js and d.ts files with imports. Fix creation of nested namespaces with the same names.

AVKurbatov:
Fixed two issues.
1) If protofile uses some Namespace (for example, Test) and it contains some Service with the same name, nested Namespace Test.Test will be generated.
This Namespace will contain classes for in and out arguments for the Service (for example Test.Test.Request_in).
They can't be used from the same Namespace becouse of names conflict.
In this fix names of nested Namespaces are updated. The argument class now will be named Test.TestService.Request_in.
To use this mode run pbjs with the flag "-u", "--unify-names".
2) Proto-file can contain imports like
import "submodule.proto";
Currently all code from them is included into the result of the protobuf-js generated js-file.
So, if we have a lot of imports, the result of generation can be enormous. 
This can be a problem and it was deeply discussed in the issue https://github.com/protobufjs/protobuf.js/issues/1839#issue-1502670536.
With this fix only content of the requested proto file will be generated. All content from imported protofiles will be imported in 
the generated js and d.ts files. 
To use this mode run pbjs and pbts with the flag "--use-imports".


cli/targets/static.js
```diff
@@ -19,9 +19,24 @@ var config = {};
...
+        return;
+    }
+    root.imports.forEach(function(oneImport) {
+        let name = oneImport;
+        const lastSlashIndex = name.lastIndexOf('/');
+        if (lastSlashIndex != -1) {
+            name = name.substring(lastSlashIndex + 1);
+        }
+        push("const " + name + "_pb = require(\"" + oneImport + "\");\n");
+    });
```
avmaxim, line 33: 1. Please use either single, or double quotes everywhere in strings for consistency
2. Please use template strings instead of using an old syntax with string concatenation 
```suggestion
    imports.forEach((import) => {
        const name = import.replace(/\/$/, '');
        push(`const ${name}_pb = require("${import}");\n`);
    });
```

cli/targets/static.js
```diff
@@ -690,13 +730,32 @@ function buildService(ref, service) {
...
-            "@param {" + exportName(method.resolvedRequestType, !config.forceMessage) + "} request " + method.resolvedRequestType.name + " message or plain object",
-            "@returns {Promise<" + exportName(method.resolvedResponseType) + ">} Promise",
+            "@param {" +  exportTypeWithProtoFromIfNeeded(method.resolvedRequestType, !config.forceMessage) + "} request " + method.resolvedRequestType.name + " message or plain object",
+            "@returns {Promise<" + exportTypeWithProtoFromIfNeeded(method.resolvedResponseType) + ">} Promise",
             "@variation 2"
         ]);
     });
 }
 
+function exportTypeWithProtoFromIfNeeded(type, asInterface) {
```
avmaxim, line 740: I recommend keeping function names more succinct. `ifNeeded` looks excess and can be casted aside
```suggestion
function exportTypeWithProtoFrom(type, asInterface) {
```

cli/targets/static.js
```diff
@@ -690,13 +730,32 @@ function buildService(ref, service) {
...
+
+
+function exportResolvedType(type, asInterface) {
+    let exportedType = exportName(type, asInterface);
+    if (type.undefinedInMainProto) {
+        exportedType = type.protoFrom + "." + exportedType;
+    } else {
+        exportedType = "$root." + exportedType;
+    }
+    return exportedType;
```
avmaxim, line 756:
```suggestion
    if (type.undefinedInMainProto) {
        return type.protoFrom + "." + exportedType;
    }
    return "$root." + exportedType;    
```
or even better:
```suggestion
    return `${ type.undefinedInMainProto ? type.protoFrom : '$root'}.${exportedType}`;
```
Pay attention to the `.` sign between `${}` and `${}` above

cli/targets/static.js
```diff
@@ -690,13 +730,32 @@ function buildService(ref, service) {
...
         ]);
     });
 }
 
+function exportTypeWithProtoFromIfNeeded(type, asInterface) {
+    let exportedType = exportName(type, asInterface);
+    if (type.undefinedInMainProto) {
+        exportedType = type.protoFrom + "." + exportedType;
+    }
+    return exportedType;
```
avmaxim, line 745:
```suggestion
    let exportedType = exportName(type, asInterface);
    return type.undefinedInMainProto ? `${type.protoFrom}.` : '' + exportedType
```

avmaxim (changes requested):
@AVKurbatov Please see my comments regarding your improvements

AVKurbatov:
> @AVKurbatov Please see my comments regarding your improvements

Fixed all

avmaxim (approved):

bochenev:
Any updates?

Issue #1856, opened 2023-01-30
snake_case to camelCase fails to match _[number]


horacimacias:
protobuf.js version: 7.2.0

A similar issue was raised https://github.com/protobufjs/protobuf.js/issues/714 but the difference here is having just a number after the underscore.

camelCase('testing_1') -> 'testing1' (expected)
camelCase('testing_1') -> 'testing_1' (actual)

copy this snippet into Chrome dev tools:

```js
var camelCaseRe = /_([a-z])/g;

function camelCase(str) {
    return str.substring(0,1)
         + str.substring(1)
               .replace(camelCaseRe, function($0, $1) { return $1.toUpperCase(); });
}
```
Then evaluate:

```js
camelCase('testing_1')
```

I believe the regular expression should be instead:

```js
var camelCaseRe = /_([a-z0-9])/g;
```

jcready:
https://github.com/protocolbuffers/protobuf/blob/7f3e23707122f31ebfa58b7280bd56cbe77cb44e/src/google/protobuf/util/field_mask_util.cc#L68

>The character after a "_" must be a lowercase letter.

horacimacias:
thanks, I wasn't aware of this.
So far every page I check for "convert to camelCase" is not behaving like that.
But you're absolutely right; searching a bit more (sorry I should have seen this earlier) I see this:

https://developers.google.com/protocol-buffers/docs/style#message_and_field_names

> If your field name contains a number, the number should appear after the letter instead of after the underscore. For example, use song_name1 instead of song_name_1


alright, thanks again. nothing to fix here. sorry for the confusion.

PR #1857, opened 2023-01-30
snake_case to camelCase fails to match _[number]

horacimacias:
Proposed fix for https://github.com/protobufjs/protobuf.js/issues/1856


horacimacias:
nothing broken here.

Issue #1863, opened 2023-02-02
Treat absolute paths as relative


ddimitrioglo:
protobuf.js version: 7.2.0

We have a package `@packaly/core-sdk` package where we store all the proto files and generated files.
After the last dependencies reinstallation, seems that the path resolution got broken.
After investigation, I could drill down to `protobufjs` which is an indirect dependency for our package.

The structure:
```bash
├── package.json
└── packaly
    ├── backend
    │   └── v1
    │        ...
    │       ├── store.pb.js
    │       ├── store.pb.ts
    │       └── store.proto
    ├── common
    │    ...
    │   ├── geo.pb.js
    │   ├── geo.pb.ts
    │   └── geo.proto
    ...

```

The `store.proto`
```js
syntax = "proto3";

package packaly.backend.v1;

import "packaly/common/geo.proto"; // <== THIS import fails

service StoreService {
  // some rpcs here
}
```

with version 7.2.0 (which is installed automatically via the dependency tree) I get the following error:
```bash
ENOENT: no such file or directory, open '/MY-PROJECT/node_modules/@packaly/core-sdk/packaly/backend/v1/packaly/common/geo.proto'
Error: The invalid .proto definition (file not found)
```

when I update my package.json with `"protobufjs": "7.1.2"` everything works as expected.

alexander-fenster:
I also see this problem, I'll see what changes breaks the world and will fix / revert.

alexander-fenster:
v7.2.1 should fix it.

alexander-fenster:
I actually suspect that if your code failed on v7.2.0, it might mean that the imports in your protos might actually be wrong, and the fix I reverted just made the loading code fail loudly where it previously accepted the wrong inputs. It does not justify the breaking change – hence I reverted #1817 – but it might signal that your protos are non compliant. Mine definitely were: I had an incorrect relative path in the `import` directive. So, if you were affected by this, it's worth checking your imports :)

ddimitrioglo:
@alexander-fenster I've posted the structure and import above, so if you see any issues I'd be more than happy to fix them if you provide any hints on how to do that...

P.S. Currently, I'm using [buf lint](https://docs.buf.build/tour/lint-your-api) to validate protos and generate the code and it passes validation with my existing structure.

PR #1864, opened 2023-02-02
fix: Revert "fix: 🐛 error should be thrown"

alexander-fenster:
Reverts protobufjs/protobuf.js#1817

Cc: @binsee 

This breaks the well established logic of proto loading (see #1863, I also see the problem in `google-gax` package). I _suspect_ that the places that got broken are actually incorrect, but we cannot just break things like that even if we are making them work properly. So rolling back for now, we'll need to figure out the safer way of fixing the original problem.

### DIFF
--- a/src/root.js
+++ b/src/root.js
@@ -98,10 +98,10 @@ Root.prototype.load = function load(filename, options, callback) {
         /* istanbul ignore if */
         if (!callback)
             return;
-        if (sync)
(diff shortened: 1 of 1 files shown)


bcoe (approved):

binsee:
Okay, it does look like it's a problem. If I have time, I will investigate this problem again.

PR #1877, opened 2023-03-22
fix: type names can be split into multiple tokens

alexander-fenster:
Some proto packages are pretty long, and some proto formatters might split them into multiple lines in a weird way, like here:

https://github.com/googleapis/googleapis/blob/master/google/ads/googleads/v13/common/asset_set_types.proto#L176-L177
```
  google.ads.googleads.v13.enums.LocationStringFilterTypeEnum
      .LocationStringFilterType filter_type = 2;
```

The parser expects the whole type name to be just one token, and fails to parse this. I'm fixing the parser to accept multiple tokens as the type name: keep glueing tokens together while there are periods that separate parts of the type name.

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -360,6 +360,16 @@ function parse(source, root, options) {
             parseGroup(parent, rule);
             return;
         }
+        // Type names can consume multiple tokens, in multiple variants:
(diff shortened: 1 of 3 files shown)


bcoe (approved):

alexander-fenster:
Parser changes are scary, so if anyone from the community could review this, I would be grateful :)  I will hold off merging for a day or two in hope someone else takes a look.

PR #1256, opened 2019-06-24
feat: parsed options

eyalpost:
Properly parse option values into objects in addition to the regular option parsing. 
Repeated options are preserved as well as repeated fields within nested options.
Parsed options are kept in a `parsedOptions` field on every level (message, field etc.)

PR #1899, opened 2023-06-23
fix: do not let setProperty change the prototype

alexander-fenster:
Follow up to #1731, fixes a bug introduced in #1256.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -176,7 +176,7 @@ util.decorateEnum = function decorateEnum(object) {
 util.setProperty = function setProperty(dst, path, value) {
     function setProp(dst, path, value) {
         var part = path.shift();
-        if (part === "__proto__") {
+        if (part === "__proto__" || part === "prototype") {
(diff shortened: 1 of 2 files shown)


src/util.js
```diff
@@ -176,7 +176,7 @@ util.decorateEnum = function decorateEnum(object) {
 util.setProperty = function setProperty(dst, path, value) {
     function setProp(dst, path, value) {
         var part = path.shift();
-        if (part === "__proto__") {
+        if (part === "__proto__" || part === "prototype") {
```
bcoe, line 179: are there any tests for `util.js`?

bcoe (approved):
LGTM, but ideally we'd add a test here:

https://github.com/protobufjs/protobuf.js/blob/master/tests/api_util.js

src/util.js
alexander-fenster: There are! added :)

Issue #1411, opened 2020-05-21
Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.


Wonder233:
protobuf.js version: 6.9.0

The above deprecation error is logged from new version of Node when using this package.

```
internal/process/warning.js:18 (node:61713) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
```
how to solve it?

iairsalem:
`DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.`
present on version 6.11.2 (apr 30)

Eneris:
Ping

maksgithub:
Same problem

panther7:
+1

panther7:
#1905

panther7:
Done via #1905.

dcodeIO:
Closing this issue because it has been inactive for more than two years and the project has changed substantially.

If this is still relevant, please open a fresh issue with current reproduction steps, version details, and any updated context.

Please also consider that maintaining protobuf.js takes time and effort. If the project is valuable to you or your organization, sponsorship is appreciated.

Thanks!

PR #1905, opened 2023-06-27
Fixed deprecation warning for new Buffer

panther7:
Used `Buffer.alloc` if it's available.

#1411

### DIFF
--- a/src/reader.js
+++ b/src/reader.js
@@ -312,9 +312,14 @@ Reader.prototype.bytes = function read_bytes() {
     this.pos += length;
     if (Array.isArray(this.buf)) // plain array
         return this.buf.slice(start, end);
-    return start === end // fix for IE 10/Win8 and others' subarray returning array of size 1
-        ? new this.buf.constructor(0)
-        : this._slice.call(this.buf, start, end);
+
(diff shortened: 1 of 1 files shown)


panther7:
^ @alexander-fenster Please, can you check it? Thanks

src/reader.js
```diff
@@ -313,7 +313,9 @@ Reader.prototype.bytes = function read_bytes() {
     if (Array.isArray(this.buf)) // plain array
         return this.buf.slice(start, end);
     return start === end // fix for IE 10/Win8 and others' subarray returning array of size 1
-        ? new this.buf.constructor(0)
+        ? (typeof Buffer.alloc === "function")
```
alexander-fenster, line 316: Will this work if there's no `Buffer` at all (in non-Node environments)? I'm afraid it might fail with `ReferenceError`.

Maybe try something like `typeof Buffer !== "undefined" && this.buf.constructor === Buffer && typeof Buffer.alloc === "function"`?

panther7: Ok, new commit apended.

alexander-fenster (approved):

panther7:
Fixed eslint

alexander-fenster:
Thank you!

Issue #216, opened 2015-01-06
Question/Feature request:  Equality check of messages


dtracers:
Is there a way to have a message check if it is equal (in terms of contained data) to another message?

If not what would be the best way to check this?
right now I compile it to base64 and compare the base64 values.

due to the fact that I dynamically add new proto files the messages themselves may from different instances but the data they contain could be exactly the same.  So doing naive field-value checking could fail.

dcodeIO:
Closing this for now.

[protobuf.js 6.0.0](https://github.com/dcodeIO/protobuf.js/releases/tag/6.0.0)

Feel free to send a pull request if this is still a requirement.

PR #1906, opened 2023-06-27
implement Message#equals

bblack:
Is there any appetite to add a function like this? I've seen this was requested in https://github.com/protobufjs/protobuf.js/issues/216 but it doesn't look like anything like this exists. Just want to see if there's any desire for it before I write tests.

Intended to fill the same role as e.g. [Message#equals](https://www.javadoc.io/static/com.google.protobuf/protobuf-java/2.3.0/com/google/protobuf/Message.html#equals(java.lang.Object)) in the Java classes generated by protoc.

### DIFF
--- a/src/message.js
+++ b/src/message.js
@@ -117,6 +117,18 @@ Message.fromObject = function fromObject(object) {
 };
 
 /**
+ * Returns whether one Message's fields have values equal to those of another Message
+ * @param {T} b Message instance
+ * @returns {Boolean}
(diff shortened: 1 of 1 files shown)


dcodeIO:
This won't work with nested messages it seems. To support nesting as well, a codegen branch or reflection walk would be necessary, and the added surface area seems not worth it compared to the alternative: compare the objects the JS-way instead. Hence closing.

PR #1907, opened 2023-06-29
fix: doc link to TS definitions

iMarv:
Seems like this has been moved in the past, fixed the link


bitcoinfinancier (approved):

dcodeIO:
The README has changed substantially since. Closing since this change no longer has a target in it.

PR #1912, opened 2023-07-12
feat: support $unknownFields

indutny-signal:
This add `$unknownFields` property to generated code. It is populated on `.decode()` when decoder encounters currently unknown field, and similarly used by encoder to serialize unknown fields as they are into a binary data.

If user stores unknown fields and then upgrades their protobufs to the latest spec - there is a chance that some of the previously unknown fields will become known. This is handled in the encoder by attempting to decode the fields first.

The name of the field starts with `$` because it is not a valid character for a field name according to protobuf v3 specification.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,5 @@
 **/node_modules/*
+bench/data/static_pbjs.js
 bin/*
 cli/wrappers/*
 coverage/*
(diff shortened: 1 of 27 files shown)


bitcoinfinancier (approved):

bitcoinfinancier (approved):

alexander-fenster:
Thanks @indutny-signal! We'll need some time to review this one but we'll try to do it soon.

indutny-signal:
Thanks a lot! FWIW, most of the changes are just generated code in tests. I can probably rip it out of the Pull Request if you'd like!

PR #1950, opened 2023-11-29
Use import instead of require for Long

baileywickham:
When using `moduleResolution: bundler` with typescript, the complier doesn't properly resolve conditional import the `Long` package causing this module to fail to build. This PR fixes this issue, using a traditional ESM import statement.

### DIFF
--- a/cli/pbts.js
+++ b/cli/pbts.js
@@ -170,7 +170,7 @@ exports.main = function(args, callback) {
                     output.push("import * as " + key + " from \"" + imports[key] + "\";");
                 });
 
-                output.push("import Long = require(\"long\");");
+                output.push("import type Long from \"long\";");
             }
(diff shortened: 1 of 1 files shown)


jenthethird:
I second this! Is there an update on this issue?

dcodeIO:
Thanks for the PR. I tried to reproduce against current master with current `Long` and TypeScript 5 using `moduleResolution: "bundler"`. There, the current `import Long = require("long")` output type-checks successfully.

Also, the proposed replacement with `import type Long from "long"` appears to regress CommonJS/non-`esModuleInterop` consumers, so I do not think this is safe as a general change.

Closing as no longer reproducible. If there is still a failing current setup, please open a fresh issue with a minimal repro.

PR #1960, opened 2024-01-13
fix: report missing import properly in loadSync

### LINKED ISSUE #1954: order of imports matters?
[filing this mostly for myself, who else would go fix it? :) ]

v7.2.5

Twice during the last two weeks I noticed a proto that failed `pbjs` being unable to locate a type (even though all dependencies were provided) but passed after I reordered proto imports.

One of these protos is https://github.com/googleapis/googleapis/blob/master/google/cloud/compute/v1/compute.proto (warning, big file!) which required `import "google/protobuf/any.proto";` to be listed first to make it work.

In another case, the sequence of 

```
import "google/api/field_behavior.proto";
import "google/api/field_info.proto";
import "google/api/resource.proto";
import "google/protobuf/duration.proto";
import "google/protobuf/timestamp.proto";
import "google/protobuf/wrappers.proto";
import "google/type/dayofweek.proto";
import "google/type/timeofday.proto";
```

did not work (saying that type of enum `google.type.TimeOfDay` cannot be found), while

```
import "google/protobuf/duration.proto";
import "google/protobuf/timestamp.proto";
import "google/protobuf/wrappers.proto";
import "google/type/dayofweek.proto";
import "google/type/timeofday.proto";
import "google/api/field_behavior.proto";
import "google/api/field_info.proto";
import "google/api/resource.proto";
```

worked.

Something dodgy is going on in the parser, I will need to take a look at it.

alexander-fenster:
Fixes #1954.

The `finish` function can be called in two modes, sync (when it throws an exception), or async (when it calls the callback). It makes sure the callback won't be called twice, but it makes one extra step and clean up the callback even it was called in sync mode. It causes problems such as in #1954 where instead of reporting "file not found" when parsing a proto, it jumps over that error and only reports the next error–or no error if the parsing happens to succeed without a bunch of imports being skipped as a result of the first exception.

Let's avoid clearing the callback if we haven't used it yet.

The problem actually occurs when loading a proto file in sync mode (e.g. by running `root.loadSync(filename)`, which is what `pbjs` does), hence the PR title.

### DIFF
--- a/src/root.js
+++ b/src/root.js
@@ -98,10 +98,10 @@ Root.prototype.load = function load(filename, options, callback) {
         /* istanbul ignore if */
         if (!callback)
             return;
-        var cb = callback;
-        callback = null;
         if (sync)
             throw err;
+        var cb = callback;
+        callback = null;
         cb(err, root);
     }
 
--- a/tests/api_root.js
+++ b/tests/api_root.js
@@ -67,6 +67,29 @@ tape.test("reflected roots", function(test) {
         });
     });
 
+    test.test(test.name + " - missing import", function(test) {
+        var root = new Root();
+        test.plan(2);
+        root.load("tests/data/badimport.proto", function(err) {
+            test.ok(err, "should return an error when an imported file does not exist");
+            test.match(err.toString(), /nonexistent\.proto/, "should mention the file name which was not found");
+            test.end();
+        });
+    });
+
+    test.test(test.name + " - missing import, sync load", function(test) {
+        var root = new Root();
+        test.plan(2);
+        try {
+            root.loadSync("tests/data/badimport.proto");
+            root.resolveAll();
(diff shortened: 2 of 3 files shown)


src/root.js
```diff
@@ -98,10 +98,10 @@ Root.prototype.load = function load(filename, options, callback) {
         /* istanbul ignore if */
         if (!callback)
             return;
-        var cb = callback;
-        callback = null;
         if (sync)
             throw err;
+        var cb = callback;
```
bcoe, line 103: The style is internally consistent, so LGTM. But, it would be an improvement to use `let` / `const` to protect us from weird scope issues.

bcoe (approved):
I'm a little confused, is `callback` called elsewhere when using `loadSync`, rather than line `105`?

Issue #1961, opened 2024-01-13
Missing valuesOptions in fromJSON parser function


dobrac:
Hi! 👋 
      
I've found out the "valuesOptions" option in Enum class is missing when parsed from the JSON.

Reproduction:
- Create .proto file containing an enum with the valuesOptions
- Generate JSON from this .proto file
- Parse the generated JSON back
- Try to access the valuesOptions field in that enum

The following code change should fix it:
```diff
diff --git a/node_modules/protobufjs/src/enum.js b/node_modules/protobufjs/src/enum.js
index 1c01620..730b8e4 100644
--- a/node_modules/protobufjs/src/enum.js
+++ b/node_modules/protobufjs/src/enum.js
@@ -87,7 +87,7 @@ function Enum(name, values, options, comment, comments, valuesOptions) {
  * @throws {TypeError} If arguments are invalid
  */
 Enum.fromJSON = function fromJSON(name, json) {
-    var enm = new Enum(name, json.values, json.options, json.comment, json.comments);
+    var enm = new Enum(name, json.values, json.options, json.comment, json.comments, json.valuesOptions);
     enm.reserved = json.reserved;
     return enm;
 };
```
<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

PR #1979, opened 2024-04-08
fix: missing valuesOptions in the fromJSON parser function (#1961)

dobrac:
I've found out that the "valuesOptions" option in the Enum class is missing when parsed from the JSON.

This PR should fix the issue: https://github.com/protobufjs/protobuf.js/issues/1961


idiomic:
I reproduced this issue on 7.4, and verified that this fix works as intended.

PR #2000, opened 2024-06-03
Update parser to support editions syntax

dcodeIO:
* Recognizes the `edition` keyword. Falls back to proto3 for now.
* Updates the parser to support more option identifiers with parenthesized extension references, e.g. `features.(pb.java).legacy_closed_enum`. Retains backwards compatibility with assumptions made by prior options logic.
* Adds dummy schemas for Cpp/Go/Java features, in case these are imported by the CLI.
* Adds a test case based on an example editions schema.

cc @protobufjs/maintainers


google/protobuf/cpp_features.proto
```diff
@@ -0,0 +1 @@
+syntax = "proto3";
```
mkruskal-google, line 1: What's the purpose of these dummies?  It seems like this could conflict with the actual versions.  Also, it doesn't actually define any custom features, so anyone referencing those would have problems

src/parse.js
```diff
@@ -128,8 +180,8 @@ function parse(source, root, options) {
         } catch (e) {
 
             /* istanbul ignore else */
-            if (acceptTypeRef && typeRefRe.test(token))
-                return token;
+            if (acceptIdentifier && nameRe.test(token))
+                return readIdentifier(token); // `ENUM_VALUE`
 
             /* istanbul ignore next */
             throw illegal(token, "value");
```
mkruskal-google, line 187: I think readRanges needs to be updated as well.  In edition 2023 we changed the grammar of reserved field/enum value names to be identifiers instead of strings.  This avoids accidental concatenation

tests/comp_edition.js
```diff
@@ -0,0 +1,42 @@
...
+option features.field_presence = EXPLICIT;\
+option features.enum_type = CLOSED;\
+option features.repeated_field_encoding = EXPANDED;\
+option features.json_format = LEGACY_BEST_EFFORT;\
+option features.utf8_validation = NONE;\
+option features.(pb.cpp).legacy_closed_enum = true;\
+option features.(pb.go).legacy_unmarshal_json_enum = true;\
+option features.(pb.java).legacy_closed_enum = true;\
+message A {\
+    repeated int32 b = 1 [features.repeated_field_encoding = EXPANDED];\
```
mkruskal-google, line 18: Adding reserved field identifiers would be good to test the grammar changes

src/parse.js
```diff
@@ -602,30 +647,26 @@ function parse(source, root, options) {
...
-
-        /* istanbul ignore if */
-        if (!typeRefRe.test(token = next()))
-            throw illegal(token, "name");
-
-        var name = token;
-        var option = name;
+        var identifier = readOptionIdentifier();
+
+        // Historically, `(some.option).prop` has been interpreted as a property
```
mkruskal-google, line 652: This sounds like exactly what it is today.  What's the change in behavior here?

src/parse.js
```diff
@@ -260,12 +310,26 @@ function parse(source, root, options) {
...
+        skip("=");
+        edition = readString();
+        isProto3 = true;
+
+        /* istanbul ignore if */
+        if (syntax)
+            throw illegal(syntax, "edition");
+
+        syntax = "proto3";
+        isProto3 = true;
```
mkruskal-google, line 329: This seems strange.  Why treat editions files like proto3?

google/protobuf/cpp_features.proto
dcodeIO: The purpose I had in mind was that an import of these files may be present in a schema originating in another language or toolchain, so having dummies prevents "file not found" errors in pbjs. Left them blank because the definitions within are not of relevance to protobuf.js anyway. But now that you say it, given that the dummies might leak out to other toolchains, it seems like a good idea to include their full contents.

src/parse.js
dcodeIO: This is mostly a comment I found useful to explain why the somewhat strange code exists in the first place. Say, someone stumbles over this part, wonders why it's there, and decides to make a PR to remove it because it doesn't seem useful to them. Other than that, there's indeed no change in behavior other than also applying the fallback to syntax previously not supported, e.g. `a.(b.c).d`.

src/parse.js
dcodeIO: Only added the fallback to produce valid schemas that can be tested, including during resolve. I guess another option is to test via the parser only, and for now produce an "editions are not supported" error during resolve?

google/protobuf/cpp_features.proto
mkruskal-google: Yea, also usually people would import them *and* reference the features defined in them.  So I think you'd still hit errors from missing options definitions?

OTOH copying them verbatim here seems like a good way to get version skew issues.  How do you deal with descriptor.proto?  We've been modifying that pretty rapidly in the last few years

src/parse.js
mkruskal-google: Ah I see, this is just a temporary test hook?  That makes sense, but I definitely like the failure approach better.  That way nobody starts depending on this behavior

google/protobuf/cpp_features.proto
dcodeIO: The missing definitions aren't an issue for protobuf.js itself. It currently parses any options literally without looking up schemas. Hence the historic comment in the parser on how properties are assumed. Similarly, descriptor.proto [is optional](https://github.com/protobufjs/protobuf.js/tree/master/ext/descriptor) and only updated sporadically, as it is not needed internally for protobuf.js to function. Perhaps a data point that descriptor.proto hasn't leaked out so far.

src/parse.js
mkruskal-google: Ah I think I misunderstood what was happening here.  FWIW it looks like you still don't support *multiple* extensions (e.g. `a.(b).c.(d).e`), but we hopefully won't ever need that

galaxycores (approved):

PR #2001, opened 2024-06-04
fix(types): reserved field in IType can contain reserved names

mkruskal-google:
This is causing downstream problems with some recent descriptor.proto changes that add reserved names.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1698,7 +1698,7 @@ export interface IType extends INamespace {
     extensions?: number[][];
 
     /** Reserved ranges */
-    reserved?: number[][];
+    reserved?: (number[]|string)[];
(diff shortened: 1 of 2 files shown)


dcodeIO (commented):
Looks like this change should be applied to the source file, as the changes to index.d.ts would be lost when this file is regenerated. 

Source location:

https://github.com/protobufjs/protobuf.js/blob/11393eaa1dc95480c0a446bca8e5a50b3095ec91/src/type.js#L227

Command to regenerate index.d.ts:

```
npm run build:types
```

mkruskal-google:
Thanks, done!

dcodeIO (approved):

PR #2006, opened 2024-06-26
Fix the issue with int64 value type in a map

bobhfut:
As an employee from ByteDance, we have discovered a bug in our codebase. The issue occurs when a property in a protocol buffer (PB) structure is of type map<int32, int64>. When the value in the map is 0, the corresponding JavaScript value is represented as the number 0 instead of Long.ZERO.

We are working to fix this bug.

### DIFF
--- a/src/decoder.js
+++ b/src/decoder.js
@@ -45,9 +45,13 @@ function decoder(mtype) {
                 ("k=%j", types.defaults[field.keyType]);
             else gen
                 ("k=null");
-
-            if (types.defaults[type] !== undefined) gen
-                ("value=%j", types.defaults[type]);
+            if (types.defaults[type] !== undefined) {
+                if (types.isLongType(type)) {
+                    gen("value= $util.Long ? $util.Long.fromValue(0) : 0");
+                } else {
+                    gen("value=%j", types.defaults[type]);
+                }
+            }
             else gen
                 ("value=null");
 
--- a/src/types.js
+++ b/src/types.js
@@ -26,6 +26,17 @@ var s = [
     "bytes"     // 14
 ];
 
+/**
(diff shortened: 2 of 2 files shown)


src/decoder.js
```diff
@@ -45,9 +45,13 @@ function decoder(mtype) {
                 ("k=%j", types.defaults[field.keyType]);
             else gen
                 ("k=null");
-
-            if (types.defaults[type] !== undefined) gen
-                ("value=%j", types.defaults[type]);
+            if (types.defaults[type] !== undefined) {
+                if (types.isLongType(type)) {
+                    gen("value= $util.Long ? $util.Long.fromValue(0) : 0");
```
dcodeIO, line 50: Seems to always set 0 but should set the default value

dcodeIO:
Isn't the issue rather the value in, say, `types.defaults["int64"]`? Already does a lookup there, whereas the new `isLongType` seems redundant on this basis.

bobhfut:
> Isn't the issue rather the value in, say, `types.defaults["int64"]`? Already does a lookup there, whereas the new `isLongType` seems redundant on this basis.

The value stored in types.defaults["int64"] is the numerical value 0, but in reality we need to generate a string like "$util.Long ? $util.Long.fromValue(0) : 0" for all xx64 related types.

bobhfut:
> > Isn't the issue rather the value in, say, `types.defaults["int64"]`? Already does a lookup there, whereas the new `isLongType` seems redundant on this basis.
> 
> maybe we just set types.defaults["xxx64"] to string "$util.Long ? $util.Long.fromValue(0) : 0" ?

The problem is that there are places where default values are being used to perform numeric comparisons.

PR #2010, opened 2024-07-24
fix: multi-line typedefs generating "let" statements

### LINKED ISSUE #2009: `pbts` generating syntax error on types when using multi-line `@typedef`s in static code
protobuf.js version: 7.3.0

Using `@typedef` definitions on static code generates invalid type syntax.
Any property after the first declared property will be prefixed with `let`

```js
/**
 * @typedef {Object} MyType
 * @property {string} prop1
 * @property {number} prop2
 * @property {number] prop3
 */
```
When running `pbts`, this will be converted to:

```ts
type MyType = {
  prop1: string;
  let prop2: number;
  let prop3: number;
}
```
This isn't valid type syntax.

The problem occurs here:

https://github.com/protobufjs/protobuf.js/blob/0a0cdb6b0c635d86a9bb96a382b4ff73ee848109/cli/lib/tsd-jsdoc/publish.js#L394-L396

`forEach` passes `(value, index)`, and `writeProperty`s second argument `declare` will then be populated with `index`.

https://github.com/protobufjs/protobuf.js/blob/0a0cdb6b0c635d86a9bb96a382b4ff73ee848109/cli/lib/tsd-jsdoc/publish.js#L400-L404

Thus, the `declare` condition will be true for every element past array index 0.

LavissaWoW:
Multi-line typedefs now generates valid type syntax

Fixes #2009

### DIFF
--- a/cli/lib/tsd-jsdoc/publish.js
+++ b/cli/lib/tsd-jsdoc/publish.js
@@ -392,7 +392,7 @@ function writeInterfaceBody(element) {
     if (element.tsType)
         writeln(element.tsType.replace(/\r?\n|\r/g, "\n"));
     else if (element.properties && element.properties.length)
-        element.properties.forEach(writeProperty);
+        element.properties.forEach((property) => writeProperty(property));
     --indent;
     write("}");
 }


danielbankhead (approved):

danielbankhead:
Thanks for your contribution!

PR #1780, opened 2022-07-27
Fix nullability of non-optional fields in TS interfaces and class-level JS comments

martin-traverse:

Issue #1779, opened 2022-07-27
Optional types nullability - not working correctly for TS


martin-traverse:
protobuf.js version: 7.0.0

I don't think optional types are working correctly. My understanding is that optional types (including proto3 optionals) should have a type signature of "<type>|null|undefined", while non-optional types should just have the base type signature. Currently, this is being applied to member definitions in JS, but not to the property type hints set at the class level. This cascades through in the TS, where optional types are respected in the concrete type definitions, but not the interface definitions where everything is nullable.

Is this diference in the type definitions intentional, or not? Even though concrete classes are returned from API calls, the nested types they are built from use interfaces. So, the net result is that our web developers still have to treat everything as nullable, regardless of what is defined as optional in the original proto files. The behviour we would like is that optional fields show as nullable, while non-optional fields do not and are supplied a default value if they are missing. I have include some example outputs below.

The root cause seems to be in static.js in the CLI, where nullability for class property comments is defined differently than for members. I will link a PR to change this, which is fairly simple. However, I realise this is a change to API and may not be desirable for everyone! So my questions are:

1) Does this all make sense and some sort of patch is needed, or have I missed something obvious?!

and

2) Assuming a fix is helpful in some cases, should it be guarded behind a CLI option? If so what that be?

**Sample proto file**

    syntax = "proto3";
    package tracdap.api;
    ...

    message FileWriteRequest {
    
        /// Tenant code for the requested operation, always required
        string tenant = 1;
    
        /// Prior object/tag version to use for update operations
        optional metadata.TagSelector priorVersion = 2;
        ...

**Sample output which shows the problem**

JS file has "tenant" as nullable in the object properties, but not the member definition

            api.FileWriteRequest = (function() {
    
                /**
                 * Properties of a FileWriteRequest.
                 * @memberof tracdap.api
                 * @interface IFileWriteRequest
                 * @property {string|null} [tenant] Tenant code for the requested operation, always required
                 * @property {tracdap.metadata.ITagSelector|null} [priorVersion] Prior object/tag version to use for update operations
                
                ...
                
                /**
                 * Tenant code for the requested operation, always required
                 * @member {string} tenant
                 * @memberof tracdap.api.FileWriteRequest
                 * @instance
                 */
                FileWriteRequest.prototype.tenant = "";
    
                /**
                 * Prior object/tag version to use for update operations
                 * @member {tracdap.metadata.ITagSelector|null|undefined} priorVersion
[...]

PR #2011, opened 2024-07-24
Fix / Handle nullability for optional fields

martin-traverse:


martin-traverse:
This PR is to address #1779. I believe it is a more complete solution than my old PR, #1780.

The expected behaviour IMO is that fields marked as optional (or fields in a one-of) should have nullable type hints and a null default, while fields marked as required (proto2) or with no annotation (proto3) should be marked not-nullable and have a zero default. In proto3, although technically everything is optional, the protobuf system should provide a default value making it not-null in practice if the optional flag is not set, which is consistent with other implementations.

After getting more into the code, the real culprit here is the Field rule. A rule of "proto3_optional" or "optional" can get set in the parser, but these both get converted to "optional" and then ultimately discarded, before reaching the generator. To avoid losing this information in proto3, I added a proto3Optional member to the Field class which gets set before the rule is discarded. I also made the parser store the proto syntax as an option on the Root object during the parse. With these two bits of information together, there is enough info to determine what is optional in the static generator.

The new functionality is behind a --force-optional flag, which sets nullability based on the optional keyword. The default is false for now, but it could be changed later. I added a couple of tests to show this is working as expected. I chose that name to match the existing flags but happy to change to whatever you think will be most clear / consistent.

Lastly, I have a question about JSON descriptors. Are these always created by running the parser and storing the result? If so they should load back in with the syntax option and work as expected. If there is another way of creating these objects that doesn't go through the parser, the syntax option will be missing and the static generator would fall back on proto2 syntax.

Please do let me know your thoughts on this when you get time. I'm still very keen to get this fix into my project, if there are things you want tweaking / adding I'm happy to do it.

martin-traverse:
Thank you to whoever approved the CI workflows. There were a few lint errors, I have fixed and re-pushed, could you approve the workflows to run again?

cli/targets/static.js
```diff
@@ -354,8 +354,48 @@ function toJsType(field) {
     return type;
 }
 
+function syntaxForType(type) {
+
+    var syntax = null;
+    var namespace = type;
+
+    while (syntax === null && namespace !== null) {
+        if (namespace.options != null && "syntax" in namespace.options)
```
sofisl, line 363: For readability, please wrap if/else statements in curly brackets

cli/targets/static.js
```diff
@@ -354,8 +354,48 @@ function toJsType(field) {
...
+            syntax = namespace.options["syntax"];
+        else
+            namespace = namespace.parent;
+    }
+
+    return (syntax === "proto3") ? "proto3" : "proto2";
```
sofisl, line 369: With editions this will likely be more options, could we make this `return syntax ?? proto2`

cli/targets/static.js
```diff
@@ -354,8 +354,48 @@ function toJsType(field) {
...
+}
+
+function isOptional(field, syntax) {
+
+    // In proto3, optional fields are explicit
+    if (syntax === "proto3")
+        return field.options != null && field.options["proto3_optional"] === true;
+
+    // In proto2, fields are explicitly optional if they are not part of a map, array or oneOf group
+    return field.optional && !(field.partOf || field.repeated || field.map);
```
sofisl, line 379: Thinking ahead to editions syntax, should we also gate this `if syntax === proto2`?

cli/targets/static.js
```diff
@@ -394,8 +449,19 @@ function buildType(ref, type) {
         if (config.comments) {
             push("");
             var jsType = toJsType(field);
-            if (field.optional && !field.map && !field.repeated && (field.resolvedType instanceof Type || config["null-defaults"]) || field.partOf)
-                jsType = jsType + "|null|undefined";
+
+            // New behaviour - fields explicitly marked as optional and members of a one-of are nullable
```
sofisl, line 453: New/old behavior is vague (there will be a lot of changes coming!) Can you specify syntax, etc., in comments instead?

cli/targets/static.js
martin-traverse: Done, changes pushed

cli/targets/static.js
martin-traverse: Done, changes pushed

cli/targets/static.js
martin-traverse: Done, changes pushed. I'm throwing an error for unknown proto syntax if it is not either "proto2" or "proto3".

cli/targets/static.js
martin-traverse: After thinking about this comment, I changed the guard flag to --null-semantics. The inline comments are updated accordingly and changes pushed. See notes below.

martin-traverse:
Thanks for all the feedback. I believe I have addressed all the comments, changes are pushed. I believe all the tests / linting should be up to date now if you want to re-run CI.

After thinking about the "new behaviour" / "old behaviour" comment, I changed the guard flag to --null-semantics. I think this is more descriptive. With this flag, the nullability of fields in messages / interfaces reflects the semantics described in the proto file. This allows bringing in various fixes under the one flag. In particular, I have folded in the change from PR #2013 (issue #2012) regarding nullability when message types are composed. It gets away from having multiple flags, keeps backwards compatibility and keeps open the option to make semantic nulls the default in future if people think that is the right thing to do. If you're happy to take this as one change for Semantic Nulls, I can close issue #2012 and PR #2013 as these would no longer be needed.

Lastly on the coming changes for edition syntax. These methods will for sure need updating. Happy to help at that point if that is useful. Also I left an issue on the protobuf GitHub page requesting the optional keyword be maintained as syntactic sugar for [features.field_presence = EXPLICIT]. After spending some time working with this, I believe nullability is a core part of the semantic type definition, not a technical annotation for how a type should be handled. Time will tell, whether anyone on the protobuf team agrees! Here is the link: https://github.com/protocolbuffers/protobuf/issues/17624

cli/targets/static.js
```diff
@@ -311,9 +311,15 @@ function buildFunction(type, functionName, gen, scope) {
         push("};");
 }
 
-function toJsType(field) {
+function toJsType(field, parentIsInterface) {
```
danielbankhead, line 314: Should we default this to `false`? This will also make the type clear.

```suggestion
function toJsType(field, parentIsInterface = false) {
```

danielbankhead (commented):
LGTM, I'd like a second opinion from @dcodeIO, if possible. I'm still familiarizing myself with the library.

martin-traverse:
Thanks @danielbankhead - I committed the update you suggested.

martin-traverse:
One more update. I've let implicit optional fields (proto3), repeated fields and maps be omitted from the properties. However, if specified, these fields must have a non-null value. I think this is more consistent with protobuf semantics, that assigns default values to these fields if they are missing. This only affects calls to craete() / new, implicit fields are non-null once they are constructed.

cli/targets/static.js
```diff
@@ -354,8 +360,66 @@ function toJsType(field) {
     return type;
 }
 
+function syntaxForType(type) {
```
sofisl, line 363: Nit, please add curly brackets around if/else for readability

sofisl (approved):
LGTM, with a nit about curly brackets. Also @dcodeIO, would love a second pair of eyes on this as well. TY!

martin-traverse:
Thanks @sofisl, I have updated all the code touched by this change to use braces for every if/else statement

cli/targets/static.js
martin-traverse: Done, changes pushed

martin-traverse:
Hello - I have closed PR #2013 as that change is now fully covered by this PR. Also I have marked all the conversations resolved since those changes are all pushed.

I believe we just need a final review now? @dcodeIO - Please let me know if there is anything you would like added / removed / changed, very happy to do the work and get this closed off!

danielbankhead (approved):

martin-traverse:
Thank you @danielbankhead for merging this! Was very happy yesterday when I saw it had gone in.

Do you have any idea yet what the schedule is for the next release?

danielbankhead:
@martin-traverse we're working on cutting a release on Monday:
- https://github.com/protobufjs/protobuf.js/pull/2017

martin-traverse:
Hi @danielbankhead - it looks like this release didn't get to NPM yet, is there another step that needs to happen?

I saw the [Github release](https://github.com/protobufjs/protobuf.js/releases/tag/protobufjs-v7.3.3) went out on Friday, but NPM is still showing v7.3.2 as the [latest version](https://www.npmjs.com/package/protobufjs/v/7.3.2?activeTab=versions).

martin-traverse:
Thank you so much! I have updated our mainline with the new package.

Issue #2012, opened 2024-07-26
Unexpected null behaviour (TS, --force-message, properties and members)


martin-traverse:
protobuf.js version: 7.3.2

I believe there is an issue with the way null is handled in the JSDoc / TS type definitions. Specifically, the construction of properties / members in JSDoc (and equivalent interfaces / messages in TS definitions). The --force-message flag affects this but doesn't behave as I would expect / like.

With --force-message=true, interfaces get built with messages. So when you call .create() or new MessageType(), the top level type is an interface, but any sub-types are declared as messages which makes declaring them in code overly verbose. OTOH, with --force-message=false messages get built with interfaces, so for message returned from API calls (and messages already built using create() in code) almost everything appears nullable - particularly repeated fields and map types, which are never null after the message has been built. So whichever way you go there is an issue with nullability, either when you are building objects or when you are using them.

A more natural approach to my mind would be that interfaces are built using other interfaces, and messages are built using other messages. Then the --force-message flag only affects what is needed in the method calls, not the construction of the types themselves.

I've made a PR for this change - it's fairly small / simple. If you agree this would be a helpful change, the question is how to release it:

a) Treat it as a fix and a breaking change, so make a suitable version bump.
b) Treat it as a fix but put it behind a feature flag, e.g. --null-compat, with the intention to eventually remove the old behaviour
c) Treat is as a feature, put it behind a feature flag and preserve both options indefinitely

This also relates to the proto3 optional semantics (see issue #1779, PR #2011), there could end up being a lot of flags controlling this behaviour which might be confusing and lead to unexpected affects as code is maintained over time. With that in mind, my vote would be for options (a) or (b). With option (b) the fix for proto3 optional support could go behind the same flag.

*Examples*

Here is an excerpt from our metadata definitions:

```proto
syntax = "proto3"

enum SchemaType { ... };
enum PartType { ... };

message SchemaDefinition {
   SchemaType schemaType = 1;
   PartType partType = 2;
   oneof schemaTypeDefinition {
      TableSchema table = 3;
      ...
   }
}

message TableSchema {
   repeated FieldSchema fields = 1;
}

message FieldSchema {
   string fieldName = 1;
   ...
   optional string formatCode = 7;
}

```

To create one of these definitions to send to our API, we'd use something like this:

```js
const fields = fieldInfo.map(f => { ... });

const schema = tracdap.metadata.SchemaDefinition.create({
   schemaType: tracdap.SchemaType.TABLE,
[...]

PR #2013, opened 2024-07-26
Feature / Updated null behaviour (TS, --force-message, properties and members)

martin-traverse:
This relates to #2012

See also discussion around proto3 optional fields, issue #1779 PR #2011


martin-traverse:
See #2011, it may be simpler to bring these changes into that one PR using a single guard flag, --null-semantics.

martin-traverse:
Closing as this is now fully covered by PR #2011

PR #2018, opened 2024-08-18
Fix line accounting for `alternateCommentMode`

shicks:
When `alternateCommentMode` is enabled, the tokenizer treats a block of multiple end-of-line (`//`) comments as a single line when keeping track of line numbers, making error messages very difficult to understand, since they refer to the wrong line number.

This change increments the line number in the inner loop, so that each line is accounted correctly.

To reproduce the bug:
```
$ npm install protobufjs

$ cat > broken.proto <<EOF
syntax = "proto3";

// Multiple lines of
// end-of-line comments
// seem to mess up
// the tokenizer's
// bookkeeping.

message Foo {};a
EOF

$ cat > load.js <<EOF
const {Root} = require('protobufjs');
new Root().load('broken.proto', {alternateCommentMode: true});
EOF

$ node load.js
```
This reports an error at broken.js line 5, rather than line 9 where the error actually is.

### DIFF
--- a/src/tokenize.js
+++ b/src/tokenize.js
@@ -274,6 +274,7 @@ function tokenize(source, alternateCommentMode) {
                                     break;
                                 }
                                 offset++;
+                                line++;
                                 if (!isLeadingComment) {
                                     // Trailing comment cannot not be multi-line
                                     break;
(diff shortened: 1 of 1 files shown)


shicks:
Is anyone maintaining this project?  It would be good to get this PR merged so that errors are reported on the correct line.  Without this, it's extremely difficult to debug parse errors.

Issue #2025, opened 2024-08-23
"length" is outside of buffer bounds


phaistonian:
protobuf.js version: 7.4.0

It appears that the latest nodejs version (22.7.0) triggers this error:

`
RangeError [ERR_BUFFER_OUT_OF_BOUNDS]: "length" is outside of buffer bounds
`

at

![CleanShot 2024-08-23 at 09 21 15@2x](https://github.com/user-attachments/assets/15d0f50e-bab0-4fbc-b506-b106fb3d43a4)

likezero:
I have  the same error
![image](https://github.com/user-attachments/assets/eabd0f99-adc7-4fbb-8361-d2a7101e810a)

markkkkas:
We are observing same issue
```
RangeError [ERR_BUFFER_OUT_OF_BOUNDS]: "length" is outside of buffer bounds
    at proto.utf8Write (node:internal/buffer:1066:13)
    at Op.writeStringBuffer [as fn] (/app/node_modules/.pnpm/protobufjs@7.2.4/node_modules/protobufjs/src/writer_buffer.js:61:13)
    at BufferWriter.finish (/app/node_modules/.pnpm/protobufjs@7.2.4/node_modules/protobufjs/src/writer.js:453:14)
    at /app/node_modules/.pnpm/@grpc+proto-loader@0.7.6/node_modules/@grpc/proto-loader/build/src/index.js:152:109
    at Array.map (<anonymous>)
    at createPackageDefinition (/app/node_modules/.pnpm/@grpc+proto-loader@0.7.6/node_modules/@grpc/proto-loader/build/src/index.js:152:39)
    at Object.fromJSON (/app/node_modules/.pnpm/@grpc+proto-loader@0.7.6/node_modules/@grpc/proto-loader/build/src/index.js:205:12)
    at GrpcClient.loadProtoJSON (/app/node_modules/.pnpm/google-gax@4.0.3_encoding@0.1.13/node_modules/google-gax/build/src/grpc.js:228:51)
    at new TranslationServiceClient (/app/node_modules/.pnpm/@google-cloud+translate@8.3.0_encoding@0.1.13/node_modules/@google-cloud/translate/build/src/v3/translation_service_client.js:146:38)
    at new GoogleTranslationService (/app/apps/dist/lib/google-translation.js:55:32)
```

cecilulysess:
Same issue, not necessary to protobufjs@7.4. The error thrown on my 7.3.3

[...]

PR #2026, opened 2024-08-25
fix: RangeError [ERR_BUFFER_OUT_OF_BOUNDS] "length" is outside of buffer bounds

waitingsong:
fix: RangeError [ERR_BUFFER_OUT_OF_BOUNDS] "length" is outside of buffer bounds since nodejs 22.7

call buf.utf8Write() conditionally,  it will be called in buf.write()

ref: https://github.com/protobufjs/protobuf.js/issues/2025

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node_version: ["12", "14", "16", "18"]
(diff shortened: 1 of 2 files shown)


dcodeIO:
Does not reproduce for me on Node.js 24.x. As far as I can tell, this was a Node.js Buffer regression rather than a protobuf.js bug, and it appears to have been fixed upstream in https://github.com/nodejs/node/pull/54524.

PR #2029, opened 2024-09-10
feat: add feature resolution for protobuf editions

sofisl:


tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,106 @@
+/*
+1. Defaults
```
mkruskal-google, line 2: While I agree that this monolithic test proto is ideal for initial development, can we split it up into fine-grained protos?  That would make the tests more readable and not require copy/pasting the entire thing into comments :P. Giving more meaningful feature names might help too (e.g. FILE, MESSAGE, FIELD, etc).

I'm not 100% clear on how protobufjs works, but I think you could also define the proto inline in the test instead of a separate file.  That would make each test much more self-contained and easier to read

src/parse.js
```diff
@@ -25,7 +25,8 @@ var base10Re    = /^[1-9][0-9]*$/,
     numberRe    = /^(?![eE])[0-9]*(?:\.[0-9]*)?(?:[eE][+-]?[0-9]+)?$/,
     nameRe      = /^[a-zA-Z_][a-zA-Z_0-9]*$/,
     typeRefRe   = /^(?:\.?[a-zA-Z_][a-zA-Z_0-9]*)(?:\.[a-zA-Z_][a-zA-Z_0-9]*)*$/,
-    fqTypeRefRe = /^(?:\.[a-zA-Z_][a-zA-Z_0-9]*)+$/;
+    fqTypeRefRe = /^(?:\.[a-zA-Z_][a-zA-Z_0-9]*)+$/,
+    featuresRefRe = /features\.([a-zA-Z_]*)/;
```
mkruskal-google, line 29: features can also have extensions (e.g. features.(pb.cpp).legacy_enum).  Do we need to special-case feature parsing like this though?  They're implemented as just options with special meaning, so the parser shouldn't really need to treat them differently (until it does feature resolution)

src/parse.js
```diff
@@ -364,7 +381,7 @@ function parse(source, root, options) {
```
mkruskal-google, line 381: "optional" and "required" should actually be error conditions under editions, idk if we need that validation layer tho (being *more* permissive than protoc might be ok?)

src/parse.js
```diff
@@ -602,34 +624,44 @@ function parse(source, root, options) {
...
-            skip(")");
-            name = "(" + name + ")";
-            option = name;
-            token = peek();
-            if (fqTypeRefRe.test(token)) {
-                propName = token.slice(1); //remove '.' before property name
-                name += token;
-                next();
+        // console.log(featuresRefRe.test(token = next()))
+        if (featuresRefRe.test(peek())) {
```
mkruskal-google, line 632: This is kindof an unfortunate place to put the branch, since features should parse as just ordinary options.  I think a safer pattern might be to pull the features out after parsing the option by just looking for ones that start with "features."

src/object.js
```diff
@@ -41,6 +41,11 @@ function ReflectionObject(name, options) {
      */
     this.name = name;
 
+    /**
+     * Resolved Features.
+     */
+    this.features = null;
```
mkruskal-google, line 47: features should be private, since nobody but the runtime should be inspecting them

src/object.js
```diff
@@ -175,6 +180,17 @@ ReflectionObject.prototype.setOption = function setOption(name, value, ifNotSet)
     return this;
 };
 
+/**
+ * Sets a feature.
+ * @param {string} name Feature name
+ * @param {*} value Feature value
+ * @returns {ReflectionObject} `this`
+ */
+ReflectionObject.prototype.setFeature = function setFeature(name, value) {
```
mkruskal-google, line 189: This should also be private

src/parse.js
mkruskal-google: Similarly, we might want to keep unresolved features in the options object, since they are technically options set in the proto file.  Nobody *should* be using them and we generally prefer to hide them where possible, but there are a few cases where they're needed (e.g. producing the same .proto file you were given)

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,106 @@
...
+*/
+
+var tape = require("tape");
+
+var protobuf = require("..");
+
+
+tape.test("feature resolution editions", function(test) {
+
+    protobuf.load("tests/data/feature-resolution.proto", function(err, root) {
```
mkruskal-google, line 88: This tests an important piece of feature resolution, but it isn't the whole story.  Specifically, this tests that a feature set on a descriptor will end up in the resolved features of that descriptor.  other notable test cases:

- features specified on a parent descriptor show up in the child descriptor if it isn't overridden
- features specified on a distant ancestor also propagate
- all features get initialized to the defaults for the edition if they aren't overridden
- multiple features can be specified at different levels, and they all get merged into the resolved feature set

src/parse.js
sofisl: I *think* so, because there's slightly different syntax for features than for other options, and because parseOption is I think the final layer before feature resolution (as it is for option resolution as well). If we don't do this special regex-matching, we'd basically be looking for open parentheses to determine whether it's a custom option, which we don't have for feature-matching. I'll expand the regex to include special options.

src/parse.js
sofisl: sg, I'll add the restrictions.

src/parse.js
mkruskal-google: the syntax for features is exactly the same as that for message-typed options.  There's nothing special about them except in how we use them

tests/feature_resolution_editions.js
sofisl: Added tests.

tests/comp_options.js
```diff
@@ -30,7 +30,7 @@ message Message {
 tape.test("complex options", function (test) {
   var root = protobuf.parse(proto).root;
 
-  test.deepEqual(root.parsedOptions[0], {
+  test.deepEqual(root.parsedOptions[1], {
```
sofisl, line 33: Since we now have defaults as the first parsed option, this test needs to look for the second parsed option.

tests/api_object.js
```diff
@@ -10,7 +10,6 @@ tape.test("reflection objects", function(test) {
     var obj = new protobuf.ReflectionObject("Test");
 
     obj.resolve();
-    test.equal(obj.resolved, false, "should not resolve when not part of a root");
```
sofisl, line 13: removing this test as we decided that `resolved` should be called on every level after it's been resolved, this will make future resolutions more efficient.

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,89 @@
...
+    // Should inherit from default, and Message, only change field_presence and the custom extension
+    test.same(rootEditionsOverriden.lookup("Message").lookup("Nested")._features, 
+    { enum_type: 'OPEN', field_presence: 'IMPLICIT', json_format: 'LEGACY_BEST_EFFORT', message_encoding: 'LENGTH_PREFIXED', repeated_field_encoding: 'PACKED', utf8_validation: 'VERIFY', '(abc)': { d_e: 'deeply_nested_true' } })
+
+    test.end();
+})
+// Tests precedence for different levels of feature resolution
+tape.test("feature resolution editions precedence", function(test) {
+
+    protobuf.load("tests/data/feature-resolution.proto", function(err, root) {
```
mkruskal-google, line 71: I think it might be helpful to split up this proto into smaller ones, and define them inline per-test.  I regret using monolithic protos in our tests, as it's much harder to understand what's being tests

tests/comp_options.js
sofisl: `parsedOptions` are just options that have been properly parsed - in other words, a step after options. Since features are just options, and they need to be parsed just like options, they do need to go through this step. If you want I can delete them from options after they've been copied over into _features.

src/parse.js
```diff
@@ -267,9 +272,37 @@ function parse(source, root, options) {
...
+    function parseEdition() {
+        skip("=");
+        edition = readString();
+        const supportedEditions = ["2023"];
+
+        /* istanbul ignore if */
+        if (!supportedEditions.includes(edition))
+            throw illegal(edition, "edition");
+
+        root.setOption("edition", edition);
```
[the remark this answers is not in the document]
sofisl: It's just following the pattern for setting syntax. I assume it will be useful later on when we have to decide how to treat things based on proto2, proto3 and editions.

src/object.js
```diff
@@ -195,12 +217,13 @@ ReflectionObject.prototype.setParsedOption = function setParsedOption(name, valu
         });
         if (opt) {
             // If we found an existing option - just merge the property value
+            // (If it's a feature, will just write over)
```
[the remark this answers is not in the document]
sofisl: This feature really only gets called when parsing Root, since root gets initialized with defaults, and then can be overwritten if there's options specified at the root level. For example, if we start off with editions syntax, and then immediately add the following at the root level:

```
option features.json_format = LEGACY_BEST_EFFORT;
```

Since the previous experience is to merge at each child, then the features-specific behavior gets lost, since we'd end up with an array of values (`features.json_format = [LEGACY_BEST_EFFORT, ALLOW]`);

One option is to simply initialize defaults at the time that we set root-level options, so there's no overwriting, just 'filling in the blanks'. However this would require much of the same logic that `setParsedOption` already does (i.e., searching for a property within the`_proto_features` object), and it seems pretty un-DRY to rewrite it for features only because the setProperty value is different.

Instead, I'll change the `isFeature` parameter to `overwrite` to make it more general.

tests/feature_resolution_editions.js
sofisl: I hear you, but I think in this case it helps the test reader understand the overall precedence of options - at least it does for me. It's clear by the alphabetical order what has precedence in the grand scheme of things, rather than having to piece it together in each test. As long as it doesn't get used for other development, I think it's helpful to get the greater overall picture of precendence.

src/parse.js
```diff
@@ -338,13 +371,17 @@ function parse(source, root, options) {
 
                 case "required":
                 case "repeated":
+                    if (edition)
+                        throw illegal(token);
                     parseField(type, token);
                     break;
 
                 case "optional":
```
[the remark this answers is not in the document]
sofisl: Added tests for the grammar changes. As for the second point, I believe this is already covered by feature resolution tests since there are no quotes.

tests/comp_options.js
mkruskal-google: My confusion is why *defaults* end up here.  Parsed features make sense to me, but you're explicitly adding edition defaults to `parsedOptions`

src/object.js
```diff
@@ -146,11 +156,31 @@ ReflectionObject.prototype.onRemove = function onRemove(parent) {
 ReflectionObject.prototype.resolve = function resolve() {
     if (this.resolved)
         return this;
-    if (this.root instanceof Root)
-        this.resolved = true; // only if part of a root
+    this._resolveFeatures();
+    if (this.root instanceof Root || this.parent)
```
mkruskal-google, line 160: Is this sufficient?

- We probably don't want to call `_resolveFeatures()` unless this condition is true
- What if the parent is itself orphaned?

src/object.js
```diff
@@ -186,6 +207,7 @@ ReflectionObject.prototype.setParsedOption = function setParsedOption(name, valu
     if (!this.parsedOptions) {
         this.parsedOptions = [];
     }
+    var isFeature = /features/.test(name);
```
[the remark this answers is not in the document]
mkruskal-google: I think you're still missing an `.`, this would pass for an option named `featuresFoo`

src/object.js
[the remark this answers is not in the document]
mkruskal-google: This sounds related to my confusion about calling `setParsedOption` with defaults.  Alternatively, you could just initialize `_features` of the root to the relevant defaults during feature resolution right?  i.e. if `_parent` is null, initialize with the edition defaults

src/parse.js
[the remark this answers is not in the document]
mkruskal-google: I don't think that second case is covered, since you have no reserved fields.  What I'm talking about is that for a proto2 message:
```
message Foo {
  reserved "bar", "baz";
}
```

becomes this in edition 2023:
```
message Foo {
  reserved bar, baz;
}
```

tests/feature_resolution_editions.js
mkruskal-google: The problem is that to understand this test you need to constantly be switching back and forth between two different files (as a reader, I find this test difficult to read).  If you had much smaller protos defined inline in each test, it would be much clearer what the expectation is

src/object.js
sofisl: - Moved _resolveFeatures, I think you're right.
- If the parent itself is orphaned, then this wouldn't get called recursively since this usually only gets called under `resolveAll`.

src/object.js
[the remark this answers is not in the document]
sofisl: Right, but what if `parent` isn't null? In that case we're rewriting a lot of the code that is in `setParsedOption` to confirm whether or not a value exists and if to rewrite it, or to create a new property, etc.

tests/comp_options.js
sofisl: Gotcha - yeah it seems it is related. I think the point is that we're taking advantage of the parsing `setParsedOptions` offers, which ultimately is needed in feature resolution. If we move it outside of setting parsedOptions into some other feature resolution-specific method, we're essentially copying over reusable code. Same reason why we're caling [setParsedOption here](https://github.com/protobufjs/protobuf.js/pull/2029/files/087a33e526f61cbf38af9b1c76342bcd06ba394b#diff-42f95033dbb9c5068c9cc85ce7c08cf1c22f68199a26bfab67817c63ff804253R638) too, since it takes advantage of all the parsing already offered in that method.

src/parse.js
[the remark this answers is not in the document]
sofisl: Added tests.

tests/feature_resolution_editions.js
sofisl: OK, switched it.

src/object.js
```diff
@@ -146,11 +160,43 @@ ReflectionObject.prototype.onRemove = function onRemove(parent) {
 ReflectionObject.prototype.resolve = function resolve() {
     if (this.resolved)
         return this;
-    if (this.root instanceof Root)
-        this.resolved = true; // only if part of a root
+    if (this.root instanceof Root || this.parent) {
```
mkruskal-google, line 163: I think we also need to make sure that parent.resolved is true.  If it's not, we can't resolve features yet

src/object.js
```diff
@@ -146,11 +160,43 @@ ReflectionObject.prototype.onRemove = function onRemove(parent) {
...
+        // (Breaks the bundler and eslint)
+        // If we don't create a shallow copy, we end up also altering the parent's
+        // features
+        var parentFeaturesMerged = Object.assign(defaults, this.parent._proto_features);
+        this._features = Object.assign(parentFeaturesMerged, this._proto_features || {});
+        this.parent._resolveFeatures();
+    } else {
+        this._features = Object.assign(defaults, this._proto_features || {});
+    }
+    this._proto_features = this._features;
```
mkruskal-google, line 196: O, my earlier comments may be related to some confusion here.  My understanding is that _proto_features contains the original specification in the proto file.  Why is it being replaced with duplicate data here?

src/parse.js
```diff
@@ -141,8 +141,15 @@ function parse(source, root, options) {
...
                 target.push(readString());
-            else
-                target.push([ start = parseId(next()), skip("to", true) ? parseId(next()) : start ]);
+            else {
+                try {
+                    target.push([ start = parseId(next()), skip("to", true) ? parseId(next()) : start ]);
+                } catch (err) {
+                    if (typeRefRe.test(token) && edition) {
+                        target.push(token);
+                    }
```
mkruskal-google, line 150: Should we rethrow the error otherwise?

src/enum.js
```diff
@@ -56,6 +56,12 @@ function Enum(name, values, options, comment, comments, valuesOptions) {
      */
     this.valuesOptions = valuesOptions;
 
+    /**
+     * Values features, if any
+     * @type {Object<string, Object<string, *>>|undefined}
+     */
+    this._valuesFeatures = {};
```
mkruskal-google, line 63: Do we need protoFeatures as well?

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,263 @@
+/*
+edition = "2023";
```
mkruskal-google, line 2: I think we can delete this one now right?

src/object.js
```diff
@@ -195,12 +242,13 @@ ReflectionObject.prototype.setParsedOption = function setParsedOption(name, valu
         });
         if (opt) {
             // If we found an existing option - just merge the property value
+            // (If it's a feature, will just write over)
             var newValue = opt[name];
-            util.setProperty(newValue, propName, value);
+            util.setProperty(newValue, propName, value, isFeature);
```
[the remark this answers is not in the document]
sofisl: Name is just `features`. Why do you think the regex is wrong?

src/parse.js
```diff
@@ -602,34 +640,42 @@ function parse(source, root, options) {
         }, function parseEnumValue_line() {
             parseInlineOptions(dummy); // skip
         });
-        parent.add(token, value, dummy.comment, dummy.options);
+        parent.add(token, value, dummy.comment, dummy.parsedOptions || dummy.options);
```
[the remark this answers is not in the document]
sofisl: `options` is the legacy path. `parsedOptions` is the new path that we're using for all the other features. `parsedOptions` will only be used if there are features (that's why it's gated up [here](https://github.com/protobufjs/protobuf.js/pull/2029/files#diff-42f95033dbb9c5068c9cc85ce7c08cf1c22f68199a26bfab67817c63ff804253R627))

tests/feature_resolution_editions.js
sofisl: I think it's still helpful for me to better understand the precedence of the fields. I've also added inline tests so it's easier for you to read.

src/parse.js
```diff
@@ -141,8 +141,15 @@ function parse(source, root, options) {
         do {
             if (acceptStrings && ((token = peek()) === "\"" || token === "'"))
                 target.push(readString());
-            else
-                target.push([ start = parseId(next()), skip("to", true) ? parseId(next()) : start ]);
+            else {
+                try {
+                    target.push([ start = parseId(next()), skip("to", true) ? parseId(next()) : start ]);
```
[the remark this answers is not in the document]
sofisl: Yes, it's ok because parseId itself doesn't call `next()`, so we just pass the same token that was passed into `parseId`

tests/feature_grammar.js
```diff
@@ -0,0 +1,40 @@
...
+    }, Error, "Error: illegal token 'optional'");
+
+    test.throws(function() {
+        protobuf.parse(protoEditionsGroup);
+    }, Error, "Error: illegal token 'group'");
+
+    test.ok(protobuf.parse(noQuote));
+    
+    test.end();
+});
```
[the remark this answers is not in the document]
sofisl: I don't know how to split it up any further. What do you have in mind?

src/object.js
sofisl: Fixed per our conversation.

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,263 @@
...
+
+    // Should inherit from default, and Message, only change field_presence and the custom extension
+    test.same(rootEditionsOverriden.lookup("Message").lookup("Nested")._features, 
+    { enum_type: 'OPEN', field_presence: 'IMPLICIT', json_format: 'LEGACY_BEST_EFFORT', message_encoding: 'LENGTH_PREFIXED', repeated_field_encoding: 'PACKED', utf8_validation: 'VERIFY', '(abc)': { d_e: 'deeply_nested_true' } })
+
+    test.end();
+})
+// Tests precedence for different levels of feature resolution
+tape.test("feature resolution editions precedence", function(test) {
+    var root1 = protobuf.parse(test1).root.resolveAll()
```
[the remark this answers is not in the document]
sofisl: set it inline.

tests/feature_grammar.js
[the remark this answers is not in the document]
mkruskal-google: e.g.
```
tape.test("editions required keyword", function(test) {
    test.throws(function() {
        protobuf.parse(`edition = "2023";
            message A {\
                required uint32 a = 1;\
            }`;);
    }, Error, "Error: illegal token 'required'");
});
```

I think ideally this would be 4 separate tests, since it's testing 4 distinct behaviors.

src/object.js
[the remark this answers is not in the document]
mkruskal-google: typo, the regex is only wrong if `name` is the fully feature path.  If it's always just `features`, this will overwrite the entire features object for each specified feature right?  It'll only keep the last one?

src/object.js
```diff
@@ -146,11 +161,40 @@ ReflectionObject.prototype.onRemove = function onRemove(parent) {
 ReflectionObject.prototype.resolve = function resolve() {
     if (this.resolved)
         return this;
+    if (this instanceof Root || this.parent && this.parent.resolved)
+        this._resolveFeatures();
     if (this.root instanceof Root)
-        this.resolved = true; // only if part of a root
+        this.resolved = true;
```
mkruskal-google, line 167: Did this change at some point?  I thought we were setting resolved to true for every type of descriptor so that we don't keep re-resolving features?

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,300 @@
...
+
+var proto3 = `syntax = "proto3";`;
+
+var editions2023Defaults = {enum_type: 'OPEN', field_presence: 'EXPLICIT', json_format: 'ALLOW', message_encoding: 'LENGTH_PREFIXED', repeated_field_encoding: 'PACKED', utf8_validation: 'VERIFY'}
+var proto2Defaults = {enum_type: 'CLOSED', field_presence: 'EXPLICIT', json_format: 'LEGACY_BEST_EFFORT', message_encoding: 'LENGTH_PREFIXED', repeated_field_encoding: 'EXPANDED', utf8_validation: 'NONE'}
+var proto3Defaults = {enum_type: 'OPEN', field_presence: 'IMPLICIT', json_format: 'ALLOW', message_encoding: 'LENGTH_PREFIXED', repeated_field_encoding: 'PACKED', utf8_validation: 'VERIFY'}
+
+tape.test("feature resolution defaults", function(test) {
+    var rootEditions = protobuf.parse(protoEditions2023).root;
+    rootEditions.resolveAll();
```
mkruskal-google, line 81: Do we need to call resolveAll in all these tests?  This worries me a bit, because it means any place where we *haven't* called resolveAll yet will see the wrong features

src/object.js
```diff
@@ -146,11 +161,40 @@ ReflectionObject.prototype.onRemove = function onRemove(parent) {
 ReflectionObject.prototype.resolve = function resolve() {
     if (this.resolved)
         return this;
+    if (this instanceof Root || this.parent && this.parent.resolved)
```
mkruskal-google, line 164: How does this work?  It looks to me like this.parent.resolved will only be true for the root.  So for any nested objects `_resolveFeatures` will never get called?

src/object.js
```diff
@@ -146,11 +161,40 @@ ReflectionObject.prototype.onRemove = function onRemove(parent) {
...
+    var defaults = {};
+
+    if (this instanceof Root) {
+        if (this.root.getOption("syntax") === "proto2") {
+            defaults = Object.assign({}, proto2Defaults);
+        } else if (this.root.getOption("syntax") === "proto3") {
+            defaults = Object.assign({}, proto3Defaults);
+        } else if (this.root.getOption("edition") === "2023") {
+            defaults = Object.assign({}, editions2023Defaults);
+        }
```
mkruskal-google, line 185: We should probably throw for unknown editions to clarify they aren't supported yet

src/object.js
```diff
@@ -146,11 +161,40 @@ ReflectionObject.prototype.onRemove = function onRemove(parent) {
...
 
+/**
+ * Resolves child features from parent features
+ * @returns {undefined}
+ */
+ReflectionObject.prototype._resolveFeatures = function _resolveFeatures() {
+    var defaults = {};
+
+    if (this instanceof Root) {
+        if (this.root.getOption("syntax") === "proto2") {
```
mkruskal-google, line 179: What happens if there's no syntax or edition specification in the .proto file?  That's valid in OSS and results in proto2 behavior

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,300 @@
...
+            option features.(abc).d_e = deeply_nested_true;
+            option features.field_presence = IMPLICIT;
+            int64 count = 9;
+        }
+        
+        enum SomeEnum {
+            ONE = 1 [features.repeated_field_encoding = EXPANDED];
+            TWO = 2;
+        }
+    }`).root
```
mkruskal-google, line 115: I think this is missing some cases, but it might be useful to split it up even further.  For example, a test covering file->message inheritance in isolation, instead of trying to bundle them all in 1 monolithic test.  The problem with this pattern is that it's still really difficult to verify the behavior by inspection.

Missing cases: oneof, extensions, top-level enum, service, method

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,300 @@
...
+    test.end();
+})
+// Tests precedence for different levels of feature resolution
+tape.test("feature resolution editions precedence", function(test) {
+    var root1 = protobuf.parse(`edition = "2023";
+    option features.amazing_feature = A;`).root.resolveAll()
+    
+    test.same(root1._features.amazing_feature, 'A');
+    
+    var root2 = protobuf.parse(`edition = "2023";
```
mkruskal-google, line 180: Similar comment here, instead of bundling these all into one test with a bunch of rootN variables, we can keep them more isolated in their own test

tests/feature_grammar.js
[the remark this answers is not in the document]
sofisl: Sure.

src/enum.js
```diff
@@ -56,6 +56,18 @@ function Enum(name, values, options, comment, comments, valuesOptions) {
...
+     * Resolved values features, if any
+     * @type {Object<string, Object<string, *>>|undefined}
+     */
+    this._valuesFeatures = {};
+
+    /**
+     * Unresolved values features, if any
+     * @type {Object<string, Object<string, *>>|undefined}
+     */
+    this._proto_valuesFeatures = {};
```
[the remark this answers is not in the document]
sofisl: Sure.

src/object.js
```diff
@@ -41,6 +46,16 @@ function ReflectionObject(name, options) {
...
 
+    /**
+     * Resolved Features.
+     */
+    this._features = {};
+
+    /**
+     * Unresolved Features.
+     */
+    this._proto_features = null;
```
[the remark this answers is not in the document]
sofisl: Sure.

src/object.js
sofisl: I noticed that `this.root` does that, since it's basically setting it true as long as the object isn't orphaned (vs. `this`)

tests/feature_resolution_editions.js
sofisl: Isn't that the behavior we want? Since features shouldn't be resolved while parsing? Not we're only calling `protobuf.parse` beforehand.

src/object.js
sofisl: We throw in `parseEdition` parse.js line 289

src/object.js
sofisl: Got it.

tests/feature_resolution_editions.js
sofisl: I don't think it's worth further splitting up this file into different tests given the amount of time I've already spent breaking this test up in each iteration of these comments (already 4 revisions to this test structure alone). It seems like you have a very specific idea for what you have these tests to look like, I'd encourage a commit if you have something very specific in mind.

tests/feature_resolution_editions.js
sofisl: Added tests.

tests/feature_resolution_editions.js
mkruskal-google: They shouldn't be resolved *while* parsing, but if this is a public API it seems like a footgun for users

tests/feature_resolution_editions.js
mkruskal-google: Your latest commit seems to be going in exactly the direction I was thinking.  "feature resolution inheritance file to oneofs" tests all of (defaults, inheritance from parent, overriding parent) for file -> oneof.  You *could* take it one step further and split that into 3 separate tests, but that seems excessive.  The key point I was trying to make was that the .proto file being tested should be minimal to make the expected behavior obvious (which you've accomplished above)

mkruskal-google: Correction: that one actually should set `LEGACY_BEST_EFFORT` at the *message* level, not the file level.  Oneofs should inherit directly from their containing message not the file (file->message tests that part, no need to test file->oneof)

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,425 @@
...
+        message_encoding: 'LENGTH_PREFIXED',
+        repeated_field_encoding: 'PACKED',
+        utf8_validation: 'VERIFY',
+        '(abc)': { d_e: 'deeply_nested_false' }
+    })
+    
+    test.end();
+});
+
+tape.test("feature resolution inheritance file to nested message", function(test) {
```
mkruskal-google, line 120: This should test message -> nested message

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,425 @@
...
+    var rootEditionsOverriden = protobuf.parse(`edition = "2023";
+    option features.json_format = LEGACY_BEST_EFFORT;
+    option features.(abc).d_e = deeply_nested_false;
+    
+    message Message {
+        string string_val = 1;
+        string string_repeated = 2 [features.enum_type = CLOSED];
+
+    message Nested {
```
mkruskal-google, line 129: nit: indentation

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,425 @@
...
+
+tape.test("feature resolution inheritance file to extensions", function(test) {
+
+    var rootEditionsOverriden = protobuf.parse(`
+    edition = "2023";
+    option features.json_format = LEGACY_BEST_EFFORT;
+    option features.(abc).d_e = deeply_nested_false;
+    message Message {
+        extensions 10 to 100;
+        extend Message {
```
mkruskal-google, line 219: file->extension should test a top level extension, not one inside a message

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,425 @@
...
+        message_encoding: 'LENGTH_PREFIXED',
+        repeated_field_encoding: 'PACKED',
+        utf8_validation: 'NONE',
+        '(abc)': { d_e: 'deeply_nested_false' } 
+    })
+
+    test.end();
+});
+
+tape.test("feature resolution inheritance file to service and method", function(test) {
```
mkruskal-google, line 262: nit: "service to method", the test itself is fine (adding a few file features for extra coverage is reasonable, but not the primary behavior under test)

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,425 @@
...
+        json_format: 'ALLOW',
+        message_encoding: 'LENGTH_PREFIXED',
+        repeated_field_encoding: 'PACKED',
+        utf8_validation: 'NONE',
+        '(abc)': { d_e: 'deeply_nested_false' } 
+    })
+
+    test.end();
+});
+
```
mkruskal-google, line 297: I think you're missing:

- message -> field
- oneof -> field
- message -> extension
- message -> enum
- file -> service

tests/feature_resolution_editions.js
sofisl: Ok, what exactly do you think the behavior should be? Since we're resolving features during `resolve`, we do need to call it right now. What do you recommend doing instead?

tests/feature_resolution_editions.js
sofisl: Added tests.

tests/feature_resolution_editions.js
sofisl: This test doesn't use LEGACY_BEST_EFFORT.

tests/feature_resolution_editions.js
mkruskal-google: Let's take this offline, this is just a design question

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,524 @@
...
+        json_format: 'LEGACY_BEST_EFFORT',
+        message_encoding: 'LENGTH_PREFIXED',
+        repeated_field_encoding: 'PACKED',
+        utf8_validation: 'VERIFY',
+        '(abc)': { d_e: 'deeply_nested_true' } 
+    });
+    test.end();
+});
+
+tape.test("feature resolution inheritance file to enums and enum values", function(test) {
```
mkruskal-google, line 172: nit: I think this one should be enum -> enum value, with features on the enum to inherit.  You already have file->enum below

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,524 @@
...
+
+tape.test("feature resolution inheritance message to oneofs", function(test) {
+
+    var rootEditionsOverriden = protobuf.parse(`
+    edition = "2023";
+    option features.(abc).d_e = deeply_nested_false;
+    message Message {
+        option features.json_format = LEGACY_BEST_EFFORT;
+        oneof SomeOneOf {
+            int32 a = 13;
```
mkruskal-google, line 215: you can put features in oneofs (same syntax as for message), do those work?

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,524 @@
...
+            option features.json_format = ALLOW;
+            int32 a = 13;
+            string b = 14;
+        }
+    }`).root.resolveAll();
+
+    test.same(rootEditionsOverriden.lookup("SomeOneOf").fieldsArray.find(x => x.name === 'b')._features, {
+        enum_type: 'OPEN',
+        field_presence: 'EXPLICIT',
+        json_format: 'LEGACY_BEST_EFFORT',
```
mkruskal-google, line 249: This looks wrong, the field is inheriting from the message and not the oneof

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,524 @@
...
+
+    var rootEditionsOverriden = protobuf.parse(`
+    edition = "2023";
+    option features.json_format = LEGACY_BEST_EFFORT;
+    option features.(abc).d_e = deeply_nested_false;
+
+    message Message {
+        option features.utf8_validation = NONE;
+        extend Message {
+            int32 bar = 10 [features.utf8_validation = VERIFY];
```
mkruskal-google, line 294: This doesn't verify that the extension inherits message features.  I think you need to use a different one at the message level too

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,524 @@
...
+
+tape.test("feature resolution inheritance message to enum", function(test) {
+
+    var rootEditionsOverriden = protobuf.parse(`edition = "2023";
+    option features.json_format = LEGACY_BEST_EFFORT;
+    option features.(abc).d_e = deeply_nested_false;
+    message Message {
+        option features.utf8_validation = NONE;
+        enum SomeEnum {
+            ONE = 1;
```
mkruskal-google, line 319: nit: add an enum feature for completeness?

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,524 @@
...
+
+    test.end();
+});
+
+tape.test("feature resolution inheritance file to service and service to method", function(test) {
+    var rootEditionsOverriden = protobuf.parse(`edition = "2023";
+    option features.json_format = LEGACY_BEST_EFFORT;
+    option features.(abc).d_e = deeply_nested_false;
+    service MyService {
+        option features.json_format = ALLOW;
```
mkruskal-google, line 366: This is also the 2023 default, so this test isn't quite locking down all the behaviors

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,524 @@
...
+        repeated_field_encoding: 'PACKED',
+        utf8_validation: 'NONE',
+        '(abc)': { d_e: 'deeply_nested_false' } 
+    })
+
+    test.end();
+});
+
+// Tests precedence for different levels of feature resolution
+tape.test("feature resolution editions precedence", function(test) {
```
mkruskal-google, line 398: I think we can delete this one now right?

tests/feature_resolution_editions.js
sofisl: Fixed!

tests/feature_resolution_editions.js
mkruskal-google: Following the pattern of your other tests, I'd expect these two tests to be collapsed into one that tests both inheritance and overriding of message -> oneof

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,452 @@
...
+
+tape.test("feature resolution inheritance oneofs to field", function(test) {
+    var rootEditionsOverriden = protobuf.parse(`
+    edition = "2023";
+    option features.(abc).d_e = deeply_nested_false;
+    message Message {
+        oneof SomeOneOf {
+            option features.json_format = LEGACY_BEST_EFFORT;
+            int32 a = 13;
+            string b = 14;
```
mkruskal-google, line 270: We should make sure overriding works as well by setting features on this field

tests/feature_resolution_editions.js
```diff
@@ -0,0 +1,452 @@
...
+
+    var rootEditionsOverriden = protobuf.parse(`
+    edition = "2023";
+    option features.json_format = LEGACY_BEST_EFFORT;
+    option features.(abc).d_e = deeply_nested_false;
+
+    extend Message {
+        int32 bar = 10 [features.utf8_validation = NONE];
+    }
+    message Message {}`).root.resolveAll();
```
mkruskal-google, line 297: We probably want to add a feature to this message to make sure the extension *doesn't* inherit from it

mkruskal-google (approved):

tests/feature_resolution_editions.js
sofisl: isn't that tested in the test below? (`feature resolution inheritance oneofs`)

mkruskal-google: Kindof, it's not clear why this one is split into two when all of the other ones test both things in a single test

sofisl: ok I merged it into one test!

Issue #1958, opened 2024-01-09
Cannot read properties of undefined (reading 'indexOf')


justaguyin2k23:
protobuf.js version: 7.2.5
gprc-js version 1.9.13
proto-loader version 0.7.10

line #488 of [protobuf.js/ext/descriptor/index.js] 

    if ((descriptor.oneofIndex = this.parent.oneofsArray.indexOf(this.partOf)) < 0)
            throw Error("missing oneof");

causes the error **Cannot read properties of undefined (reading 'indexOf')** when calling proto-loader's  loadPackageDefinition function. 

```js
const proto = grpc.loadPackageDefinition(packageDefinition);
```

This specifically happens when the latest version of buf/validate.proto is among the proto files being used.

Changing line #488 to below solves the issue, 

if ((**this.parent.oneofsArray &&** descriptor.oneofIndex = this.parent.oneofsArray.indexOf(this.partOf)) < 0)

Stacktrace
```
From Node.js Internals:
  TypeError: Cannot read properties of undefined (reading 'indexOf')
      at Field.toDescriptor (/Users/tiredcoder/dev/cncp/cncp-apis/cypress-tests/node_modules/protobufjs/ext/descriptor/index.js:488:62)
      at Root_toDescriptorRecursive (/Users/tiredcoder/dev/cncp/cncp-apis/cypress-tests/node_modules/protobufjs/ext/descriptor/index.js:142:40)
      at Root_toDescriptorRecursive (/Users/tiredcoder/dev/cncp/cncp-apis/cypress-tests/node_modules/protobufjs/ext/descriptor/index.js:146:13)
      at Root_toDescriptorRecursive (/Users/tiredcoder/dev/cncp/cncp-apis/cypress-tests/node_modules/protobufjs/ext/descriptor/index.js:146:13)
      at Root.toDescriptor (/Users/tiredcoder/dev/cncp/cncp-apis/cypress-tests/node_modules/protobufjs/ext/descriptor/index.js:121:5)
      at createPackageDefinition (/Users/tiredcoder/dev/cncp/cncp-apis/cypress-tests/node_modules/@grpc/proto-loader/src/index.ts:327:66)
      at Object.loadSync (/Users/tiredcoder/dev/cncp/cncp-apis/cypress-tests/node_modules/@grpc/proto-loader/src/index.ts:392:10)
 
     at createService (/Users/tiredcoder/dev/cncp/cncp-apis/cypress-tests/cypress.config.ts:38:49)
[...]

PR #2042, opened 2025-01-24
fix: Add check for existing array before reading

marsv024:
When parsing a proto and handling oneofs there is no check if the array actually exists before reading it, causing the code to throw errors some times.

This change add a check if the oneoffsArray exist before trying to read it.

Seems to be the samme issue as described in: https://github.com/protobufjs/protobuf.js/issues/1958

### DIFF
--- a/ext/descriptor/index.js
+++ b/ext/descriptor/index.js
@@ -485,7 +485,7 @@ Field.prototype.toDescriptor = function toDescriptor(syntax) {
 
     // Handle part of oneof
     if (this.partOf)
-        if ((descriptor.oneofIndex = this.parent.oneofsArray.indexOf(this.partOf)) < 0)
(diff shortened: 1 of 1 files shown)


josephharrington:
@sofisl Is this something you could help get merged and released? I found this open PR after we started running into this error as well, but I'm not sure who exactly to ping.

It seems to be much more likely to hit the `indexOf` error when using `@grpc/proto-loader` since that library uses the `toDescriptor` methods.

I added a [minimal repro](https://github.com/protobufjs/protobuf.js/issues/1958#issuecomment-2755625303) of the issue to #1958. I'm happy to add a regression test as well if it will help this get merged and released. Thanks!

josephharrington:
@dcodeIO Is this something you could help get merged and released? Or who would be the best maintainer to take a look? Thanks!

dcodeIO:
Addressed by #2122. The underlying issue #1958 was closed by that PR.

PR #2051, opened 2025-03-21
feat: add Edition 2023 Support

mkruskal-google:
Protobuf.js was audited for the following features:

*  Field presence - Full support of all 3 presence modes, although the behavior appears to be non-conformant in general.  The feature values were hooked up to the existing behaviors, with new tests added for edition 2023.
* Packed encoding - Both packed and expanded encoding are supported and conformant.  Some opportunistic cleanup was done, but was behavior-preserving.
* Delimited encoding - Proto2 groups are supported but mark a property on the *message* rather than the field.  This information was moved to the field and the more general delimited logic was added (with tests).
* Enum type - All enums are open, this was left untouched.  CLOSED enum_type will be ignored in edition 2023 (which is consistent with our long-term goal anyway).
* UTF8 validation - There doesn't seem to be any explicit validation of utf8, beyond what javascript may provide by default.  This would treat proto2/proto3 and edition 2023 the same regardless of features.

The proto target CLI appeared to be broken in a number of edge cases related to editions features, so these were fixed and tested were added.  It now supports proto2 and proto3 roundtrips, and conversions to proto2/proto3 (even *from* edition 2023) as long as they're valid.  No extra validation layer was added, so impossible conversions will just produce invalid protos.  A new target for edition 2023 wasn't added, but could be pretty easily if necessary in the future.

One major design flaw in protobuf.js, that's common in proto3-based code, is that the concept of files was largely dropped.  The in-memory and JSON representation of descriptors puts all of the file metadata into the most-nested package namespace object, but this object may be shared by multiple files.  This results in possible clobbering and merging, leading to malformed file metadata.  In proto2/proto3, this largely doesn't matter outside of custom file options, but the effects can be seen in the hack that forces `packed = false` for proto2 descriptors to get the repeated field encoding correct.  This problem was made worse in editions, where we store important metadata at the file level (edition + features).  The solution implemented here collapses all of this information into the top-level objects instead of putting it on the namespace.  This results in some minimal duplication, but has the ability to properly support proto2, proto3, and editions in addition to fixing some pre-existing bugs (demonstrated by some newly added tests).

Other changes:
* Group decoding will now look for the proper matching end-group tag instead of *any* end-group tag
* A bug in feature resolution for extensions was fixed
* A bug in how unresolved features showed up in the options struct was fixed
* Standard reflection helpers were added that are compatible with editions and proto2/proto3
* Proto2 and proto3 are now treated as special editions (with full feature resolution).  This substantially increases test coverage of all of the new editions logic
* The JSON spec now fully supports proto2 descriptors, instead of the previous hack of forcing `packed=false` to make them behave properly

**Migration Guide**
* Code must always ensure `.resolveAll()` is called on dynamic built protos.  This is needed when building from constructors directly or individual fromJSON/addJSON methods.  Top-level builders like `Root.fromJSON`, `Root.load`, and `protobuf.parse` do not need to explicitly call this

### DIFF
--- a/cli/targets/proto.js
+++ b/cli/targets/proto.js
@@ -102,7 +102,7 @@ function buildRoot(root) {
     if (pkg.length)
         out.push("", "package " + pkg.join(".") + ";");
 
-    buildOptions(ptr, ["edition", "syntax"]);
+    buildOptions(ptr, ["edition"]);
     ptr.nestedArray.forEach(build);
 }
 
--- a/cli/targets/proto.js
+++ b/cli/targets/proto.js
@@ -102,7 +102,7 @@ function buildRoot(root) {
     if (pkg.length)
         out.push("", "package " + pkg.join(".") + ";");
 
-    buildOptions(ptr, ["edition"]);
+    buildOptions(ptr);
     ptr.nestedArray.forEach(build);
 }
 
@@ -235,7 +235,7 @@ function buildFieldOptions(field) {
     }
     var sb = [];
     keys.forEach(function(key) {
-        if (key === "proto3_optional" || key === "packed" || key.startsWith("features.")) return;
+        if (key === "proto3_optional" || key === "packed" || key === "features") return;
 
         var val = field.options[key];
         switch (key) {
@@ -330,7 +330,7 @@ function buildOptions(object, ignore = []) {
         return;
     first = true;
     Object.keys(object.options).forEach(function(key) {
-        if (ignore.includes(key) || key.startsWith("features.") || key === "edition") return;
+        if (ignore.includes(key) || key === "features") return;
         if (first) {
             first = false;
             push("");
--- a/cli/targets/proto.js
+++ b/cli/targets/proto.js
@@ -257,9 +257,9 @@ function buildFieldOptions(field) {
     });
     var packable = types.packed[field.resolvedType instanceof Enum ? "int32" : field.type];
     if (packable !== undefined) {
-        if (field.packed && syntax == 2) {
+        if (field.packed && syntax === 2) {
             sb.push("packed=true");
-        } else if(!field.packed && syntax == 3) {
+        } else if(!field.packed && syntax === 3) {
             sb.push("packed=false");
         }
     }
--- a/cli/targets/proto.js
+++ b/cli/targets/proto.js
@@ -330,7 +330,7 @@ function buildOptions(object, ignore = []) {
         return;
     first = true;
     Object.keys(object.options).forEach(function(key) {
-        if (ignore.includes(key) || key.startsWith("features.")) return;
+        if (ignore.includes(key) || key.startsWith("features.") || key === "edition") return;
         if (first) {
             first = false;
             push("");
--- a/cli/targets/proto.js
+++ b/cli/targets/proto.js
@@ -102,7 +102,7 @@ function buildRoot(root) {
     if (pkg.length)
         out.push("", "package " + pkg.join(".") + ";");
 
-    buildOptions(ptr);
+    buildOptions(ptr, ["edition", "syntax"]);
     ptr.nestedArray.forEach(build);
 }
 
@@ -184,8 +184,12 @@ function buildType(type) {
 }
 
 function buildField(field, passExtend) {
-    if (field.partOf || field.declaringField || field.extend !== undefined && !passExtend)
+    if (field.partOf && !field.partOf.isProto3Optional) {
         return;
+    }
+    if (field.declaringField || field.extend !== undefined && !passExtend) {
+        return;
+    }
     if (first) {
         first = false;
         push("");
@@ -199,8 +203,10 @@ function buildField(field, passExtend) {
         sb.push("map<" + field.keyType + ", " + field.type + ">");
     else if (field.repeated)
         sb.push("repeated", field.type);
-    else if (syntax === 2 || field.parent.group)
+    else if (syntax === 2)
         sb.push(field.required ? "required" : "optional", field.type);
+    else if (syntax === 3 && field.hasPresence)
+        sb.push("optional", field.type);
     else
         sb.push(field.type);
     sb.push(underScore(field.name), "=", field.id);
@@ -211,7 +217,7 @@ function buildField(field, passExtend) {
 }
 
 function buildGroup(field) {
-    push(field.rule + " group " + field.resolvedType.name + " = " + field.id + " {");
+    push((field.rule || "optional") + " group " + field.resolvedType.name + " = " + field.id + " {");
     ++indent;
     buildOptions(field.resolvedType);
     first = true;
@@ -223,20 +229,16 @@ function buildGroup(field) {
 }
 
 function buildFieldOptions(field) {
-    var keys;
-    if (!field.options || !(keys = Object.keys(field.options)).length)
-        return null;
+    var keys = [];
+    if (field.options) {
+        keys = Object.keys(field.options);
+    }
     var sb = [];
     keys.forEach(function(key) {
+        if (key === "proto3_optional" || key === "packed" || key.startsWith("features.")) return;
+
         var val = field.options[key];
-        var wireType = types.packed[field.resolvedType instanceof Enum ? "int32" : field.type];
         switch (key) {
-            case "packed":
-                val = Boolean(val);
-                // skip when not packable or syntax default
-                if (wireType === undefined || syntax === 3 === val)
-                    return;
-                break;
             case "default":
                 if (syntax === 3)
                     return;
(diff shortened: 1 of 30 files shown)


sofisl (approved):
Mostly questions to determine whether something is breaking but seems mostly not.

tests/feature_resolution_editions.js
```diff
@@ -423,16 +472,245 @@ tape.test("feature resolution editions precedence", function(test) {
...
-        enum_type: 'CLOSED',
-        field_presence: 'EXPLICIT',
-        json_format: 'LEGACY_BEST_EFFORT',
-        message_encoding: 'LENGTH_PREFIXED',
-        repeated_field_encoding: 'EXPANDED',
-        utf8_validation: 'NONE',
-        amazing_feature: 'G'
-      })
-    test.end();    
+        test.same(root.lookup("Message").lookupEnum("SomeEnumInMessage")._features,
```
sofisl, line 475: this part seems like the only other obvious breaking change but since we hadn't released features earlier seems ok. But just confirming the previous test wasn't right?

sofisl:
@dcodeIO can you take a look at this PR? Debating whether this should be a breaking change or not - would love your input, thanks!

cli/targets/proto.js
```diff
@@ -199,8 +203,10 @@ function buildField(field, passExtend) {
         sb.push("map<" + field.keyType + ", " + field.type + ">");
     else if (field.repeated)
         sb.push("repeated", field.type);
-    else if (syntax === 2 || field.parent.group)
+    else if (syntax === 2)
```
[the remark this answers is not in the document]
mkruskal-google: No, this was just producing invalid protos before :).  Groups don't make required/optional labels valid in proto3 or editions.  This was just an opportunistic bug fix.

FWIW this file seemed to have *a lot* of bugs and no tests.  I fixed what I could and added a decent amount of coverage.  I don't know if anyone is actually using it

cli/targets/static.js
```diff
@@ -362,72 +362,12 @@ function toJsType(field, parentIsInterface = false) {
...
-    }
-
-    if (oneof.fieldsArray == null || oneof.fieldsArray.length !== 1) {
-        return false;
-    }
-
-    var field = oneof.fieldsArray[0];
-
-    return field.options != null && field.options["proto3_optional"] === true;
+function isNullable(field) {
```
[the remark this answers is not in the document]
mkruskal-google: It's really just a refactor and simplification.  All of the syntax logic has been encapsulated into feature helpers like hasPresence and required (see https://protobuf.dev/editions/implementation/#syntax_reflection).

This should be logically equivalent for proto2/proto3, but is now compatible with editions also.

src/enum.js
```diff
@@ -85,21 +79,20 @@ function Enum(name, values, options, comment, comments, valuesOptions) {
...
+    edition = this._edition || edition;
+    ReflectionObject.prototype._resolveFeatures.call(this, edition);
 
-    for (var key of Object.keys(this._valuesProtoFeatures)) {
+    Object.keys(this.values).forEach(key => {
         var parentFeaturesCopy = Object.assign({}, this._features);
-        this._valuesFeatures[key] = Object.assign(parentFeaturesCopy, this._valuesProtoFeatures[key] || {});
-    }
+        this._valuesFeatures[key] = Object.assign(parentFeaturesCopy, this.valuesOptions && this.valuesOptions[key] && this.valuesOptions[key].features);
+    });
```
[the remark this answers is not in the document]
mkruskal-google: How so?  this is just using similar short-circuit logic with `&&` instead of `||`.  Basically, if there are features for the value already, merge them, otherwise assign null (i.e. copy parentFeaturesCopy)

src/enum.js
```diff
@@ -176,21 +173,6 @@ Enum.prototype.add = function add(name, id, comment, options) {
         if (this.valuesOptions === undefined)
             this.valuesOptions = {};
         this.valuesOptions[name] = options || null;
-
```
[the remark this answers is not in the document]
mkruskal-google: Yea I collapsed all protoFeatures directly into options in a cleanup commit to reduce the footprint of the PR.  So the unresolved features get stored like regular options now, and resolved features get tracked separately

tests/feature_resolution_editions.js
mkruskal-google: Yea this was actually a bug fix :). The edition 2023 defaults weren't getting set correctly here, and proto2 defaults were being used

Tofandel:
This broke presence detection for proto3 syntax, proto3 syntax only allows `optional` keyword, this means that anything not explicitely optional is semantically required, but with this change everything is now considered optional and the keyword has no effect

tests/feature_resolution_editions.js
```diff
@@ -423,16 +472,245 @@ tape.test("feature resolution editions precedence", function(test) {
...
+    var proto2 = root.lookup("Message2");
+    test.ok(proto2.fields.default.hasPresence, "proto2 uses explicit presence");
+    test.ok(proto2.fields.required.required, "proto2 has required fields");
+    test.notOk(proto2.fields.unpacked.packed, "proto2 is expanded by default");
+
+    var proto3 = root.lookup("Message3");
+    test.ok(proto3.fields.explicit.hasPresence, "proto3 optional has explicit presence");
+    test.notOk(proto3.fields.implicit.hasPresence, "proto3 is implicit presence by default");
+    test.ok(proto3.fields.packed.packed, "proto3 is packed by default");
+
```
Tofandel, line 713: `test.ok(proto3.fields.implicit.required, "proto3 is required by default");`

cli/targets/static.js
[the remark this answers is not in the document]
Tofandel: It was in fact a breaking change

src/field.js
```diff
@@ -198,26 +183,76 @@ function Field(name, id, type, rule, extend, options, comment) {
...
 /**
- * Determines whether this field is packed. Only relevant when repeated and working with proto2.
+ * Determines whether this field is required.
+ * @name Field#required
+ * @type {boolean}
+ * @readonly
+ */
+Object.defineProperty(Field.prototype, "required", {
+    get: function() {
+        return this._features.field_presence === "LEGACY_REQUIRED";
```
Tofandel, line 193: Seems that this is the root of the issue

tests/feature_resolution_editions.js
mkruskal-google: That first test *should* fail.  Proto3 never had "required" fields in the proto2 sense.  Proto3 non-optional fields and proto2 "required" fields are two *very* different things.  Proto3 non-optional fields are "implicit" presence, where if they're unspecified, default values get used.

See https://protobuf.dev/programming-guides/field_presence/ and https://protobuf.dev/best-practices/dos-donts/#dont-add-a-required-field for more information.

Note: this change didn't change proto3 behavior at all.  It only clarified the code and added support for edition 2023.  Your test would have failed before this PR as well, because `required` was simply set `rule === "required"` which was only possible in proto2.

src/field.js
mkruskal-google: See above comment, this is behavior preserving

Tofandel: And yet the protobuf output between the previous version and this version is different. Now all fields are required which was NOT the case before

cli/targets/proto.js
```diff
@@ -184,8 +184,12 @@ function buildType(type) {
 }
 
 function buildField(field, passExtend) {
-    if (field.partOf || field.declaringField || field.extend !== undefined && !passExtend)
+    if (field.partOf && !field.partOf.isProto3Optional) {
```
Tofandel, line 187: See this isProto3Optional? That's what's not preserved in the legacy required detection

cli/targets/static.js
```diff
@@ -362,72 +362,12 @@ function toJsType(field, parentIsInterface = false) {
...
-    }
-
-    return syntax !== null ? syntax : "proto2";
-}
-
-function isExplicitPresence(field, syntax) {
-
-    // In proto3, optional fields are explicit
-    if (syntax === "proto3") {
-        return field.options != null && field.options["proto3_optional"] === true;
```
Tofandel, line 386: See here as well

tests/feature_resolution_editions.js
Tofandel: It breaks what this PR https://github.com/protobufjs/protobuf.js/pull/2011 introduced

Tofandel: https://github.com/Tofandel/protobuf-regression

I'll open an issue as well

mkruskal-google: Responded on the issue, the root problem was that by using the CLI and only upgrading one of your two dependencies, you essentially split this PR in half.  It would be nice if npm had a way of specifying this kind of "optional" dependency, where the latest version of protobufjs should bump your protobufjs-cli dependency automatically

PR #2056, opened 2025-04-04
Fix: Allow absolute URLs as paths, e.g. for load()

tchebb:
When using protobuf.js in a browser, it's useful to be able to load proto files from other origins. But currently, `path.normalize()` mangles URLs by removing the second slash after the scheme (e.g. `https://example.com/file.js` becomes `https:/example.com/file.js`).

Fix that issue, and better integrate the existing UNC support, by introducing a new `path.absolutePrefix()` function that checks for all the various prefixes that indicate a path is absolute and returns the one it finds (if any). By including both slashes in the return value for URLs, we exclude them from the normalization logic.

Note that this almost certainly still isn't perfect, as URLs, UNC paths, and even Windows drive letters are all quite complex. But I don't have the time or project context to make a major change here.

### DIFF
--- a/lib/path/index.d.ts
+++ b/lib/path/index.d.ts
@@ -1,4 +1,11 @@
 /**
+ * Gets the prefix of an absolute Windows, UNC, or Unix path or a URL that indicates it's absolute.
+ * @param {string} path Path to test
+ * @returns {string|null} the prefix, or null if path is relative
+ */
+export function absolutePrefix(path: string): (string|null);
+
+/**
  * Tests if the specified path is absolute.
  * @param {string} path Path to test
  * @returns {boolean} `true` if path is absolute
--- a/index.d.ts
+++ b/index.d.ts
@@ -2424,6 +2424,13 @@ export namespace util {
     namespace path {
 
         /**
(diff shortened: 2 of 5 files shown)


lib/path/index.d.ts
```diff
@@ -1,3 +1,10 @@
+/**
+ * Gets the prefix of an absolute Windows, UNC, or Unix path or a URL that indicates it's absolute.
+ * @param {string} path Path to test
+ * @returns {string|null} the prefix, or null if path is relative
+ */
+export function absolutePrefix(path: string): (string|null);
+
```
tchebb, line 7: Is there a script to update these `index.d.ts` files? I saw that `npm run build:types` updates the one at the root, but I didn't see anything similar for these, so I just made the change manually.

tchebb (commented):
I'm not sure what your workflow is for updates inside `lib/`. I didn't bump the version in `lib/path/package.json`, but I assume that'll be needed for protobuf.js to see these changes. Let me know if I'm missing something.

tchebb:
Note that this fixes the same problem that #1820 does. I didn't notice that PR before making this one, although this one has the added benefit of supporting additional URL schemes and simplifying the UNC code too.

PR #2058, opened 2025-04-11
fix: reserved field inside group blocks fail parsing

jonathan-ioffe:
Add handler to "reserved" field when parsing tokens for parseGroup.
This fixes issue https://github.com/protobufjs/protobuf.js/issues/2059


mkruskal-google:
This lgtm, but can you add a test case?  Maybe just a bit in tests/data/uncommon.proto to lock down that this continues to work?

mkruskal-google (approved):

PR #2092, opened 2025-08-14
fix: invalid syntax in descriptor.proto

josephharrington:
A recent change (#2075) edited descriptor.proto to include a `reserved` statement that combines a field number and field name in the same line. This is specifically called out in the proto language guide as invalid.

https://protobuf.dev/programming-guides/editions/#reserved

> Note that you can’t mix field names and numeric values in the same
> `reserved` statement.

Protobuf.js itself seems to handle this syntax just fine, but other tools like protoc will choke on it.

### DIFF
--- a/google/protobuf/descriptor.json
+++ b/google/protobuf/descriptor.json
@@ -588,11 +588,11 @@
                   42,
                   42
                 ],
-                "php_generic_services",
                 [
                   38,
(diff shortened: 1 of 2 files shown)


waclock (approved):
Thanks, this issue is currently breaking our apps internally

Issue #1229, opened 2019-05-26
Is the example rpcImpl in the README.md correct?


GeorgeBills:
The example given is:

```
const rpcImpl = function(method, requestData, callback) {
  client.makeUnaryRequest(
    method.name,
    arg => arg,
    arg => arg,
    requestData,
    callback
  )
}
```

That gives a method call to the path "SayHello" which won't actually work against the [helloworld example server](https://github.com/grpc/grpc/blob/v1.6.x/examples/node/static_codegen/greeter_server.js) - it needs to include the namespace and service in the path: "/helloworld.Greeter/SayHello".

The below seems to work for me:

```
const rpcImpl = function (method, requestData, callback) {
    const service = method.parent, namespace = service.parent;
    client.makeUnaryRequest(
        "/" + namespace.name + "." + service.name + "/" + method.name,
        arg => arg,
        arg => arg,
        requestData,
        callback
    )
}
```

longility:
I spent about a day's worth of work before figuring this out on my own. I searched here to find that I'm not the only one. Is this the correct way? If so, we should update the example.

longility:
`method.parent` do not exist anymore. I had to hardcode for now.

Mohammad-Khalid23:
@longility 
`package parentName;

syntax = "proto3";`

if you will use this in your proto file then you will get method.parent

longility:
I do have that @Mohammad-Khalid23 . I'm generating my code like this, which works, but maybe I'm missing something:

```
pbjs -t static-module -w commonjs -o src/junk-domain/__grpc-stubs__/profile.js ../profile/*.Api/Protos/*.proto
pbts -o src/junk-domain/__grpc-stubs__/profile.d.ts src/junk-domain/__grpc-stubs__/profile.js
```

I followed this: https://www.npmjs.com/package/protobufjs#pbts-for-typescript

morcam:
Yeah, I ran into this today as well. I don't have any parent value either. Assuming that there are indeed some cases where that parent value isn't present, isn't this an actual problem with the library, rather than just with the example? Or is the intended usage pattern to capture the namespace & service class in the RPC implementation? That would seem a bit odd to me, given that the name is passed in already.

Raylin51:
I have the same problem. Is it resolved?

hermanbanken:
Correct, this example is no longer accurate. Even the gRPC library is updated and now has more fields on `makeUnaryRequest`.

[...]

PR #2139, opened 2026-04-13
replace method with serviceName, methodName,methodIndex  in function rpcImpl

lili2012:
fix issue #1229

### DIFF
--- a/cli/targets/static.js
+++ b/cli/targets/static.js
@@ -710,7 +710,7 @@ function buildService(ref, service) {
         push("};");
     }
 
-    service.methodsArray.forEach(function(method) {
+    service.methodsArray.forEach(function(method, methodIndex) {
         method.resolve();
         var lcName = protobuf.util.lcFirst(method.name),
             cbName = escapeName(method.name + "Callback");
@@ -737,7 +737,7 @@ function buildService(ref, service) {
         ]);
         push("Object.defineProperty(" + escapeName(service.name) + ".prototype" + util.safeProp(lcName) + " = function " + escapeName(lcName) + "(request, callback) {");
             ++indent;
-            push("return this.rpcCall(" + escapeName(lcName) + ", $root." + exportName(method.resolvedRequestType) + ", $root." + exportName(method.resolvedResponseType) + ", request, callback);");
+            push(`return this.rpcCall(\"${service.name}\", \"${method.name}\", ${methodIndex},  $root.` + exportName(method.resolvedRequestType) + ", $root." + exportName(method.resolvedResponseType) + ", request, callback);");
             --indent;
         push("}, \"name\", { value: " + JSON.stringify(method.name) + " });");
         if (config.comments)
--- a/index.d.ts
+++ b/index.d.ts
@@ -1436,11 +1436,13 @@ export namespace rpc {
 
 /**
  * RPC implementation passed to {@link Service#create} performing a service request on network level, i.e. by utilizing http requests or websockets.
- * @param method Reflected or static method being called
+ * @param serviceName serviceName
+ * @param methodName methodName
+ * @param methodIndex methodIndex
  * @param requestData Request data
  * @param callback Callback function
  */
-type RPCImpl = (method: (Method|rpc.ServiceMethod<Message<{}>, Message<{}>>), requestData: Uint8Array, callback: RPCImplCallback) => void;
+type RPCImpl = (serviceName: string, methodName: string, methodIndex: number, requestData: Uint8Array, callback: RPCImplCallback) => void;
 
 /**
  * Node-style callback as used by {@link RPCImpl}.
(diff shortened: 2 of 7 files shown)


lili2012:
My frontend is js in browser and backend is c++ protobuf. I use websocket to stream. without service name my c++ backend have difficulty to choose the rpc method. This pull request have pass my work requirement and all you tests. @dcodeIO Please review it and explain my change here [#1229 ](https://github.com/protobufjs/protobuf.js/issues/1229) .  Feel free to ask question. Thanks!

alexander-fenster:
It seems like this is a breaking change to the public API, right?

Besides that, I'm unsure why you need `rpcImpl` to accept `serviceName`, can't it be obtained from the `method` directly? I must admit I haven't wrote any code with this for quite some time but if I remember things correctly, if you have a`method`, you have everything:

`test.proto`
```
syntax = "proto3";

package test;

service Foo {
  message Empty {}

  rpc Bar(Empty) returns (Empty);
}
```

`test.js`
```

const root = protobuf.loadSync("test.proto");
const method = root.lookup('test.Foo.Bar');
const parts = [];
let p = method;
while (p) {
  if (p.constructor.name === 'Service') {
    console.log('service:', p.name);
  }
  parts.unshift(p.name);
  p = p.parent;
}
console.log(parts.join('.'));
```

```
$ node test.js
service: Foo
.test.Foo.Bar
```

This should be enough to build any network RPC you want, right?

lili2012:
@alexander-fenster 

> It seems like this is a breaking change to the public API, right?

Yes this is a breaking change to rpcImpl function. I question whether rpc is widely used by protobuf.js users as you could see rpcImpl lack essential serviceName.

 `const root = protobuf.loadSync("test.proto");`
You are right your implementation works when you load proto files with relection. 
But I use pbjs compile test.proto to js  static code without reflection as a module

`npx pbjs -t static-module -w es6 --no-verify  --no-convert -p /protobuf -o /protobufjs/test_module_min.js protobuf\test.proto --no-create --no-delimited --force-number --force-message`

then I import test_module_min.js to my frontend project.

1. `p.parent` p doesn't have parent property in static code

2. `p.constructor.name === 'Service'` p.constructor.name will return 'e' or other gibberish letters after minification. When you deploy your frontend code minification is a must. 
3. previous implementation: `const rpcImpl = function(method, requestData, callback)` method.name could servive after minification because you generate code like below:
```
    Object.defineProperty(UploadService.prototype.uploadDwg = function uploadDwg(request, callback) {
        return this.rpcCall("UploadService", "uploadDwg", 0,  $root.DwgGzDrawing, $root.DwgPbDrawing, request, callback);
    }, "name", { value: "uploadDwg" });
```
method don't have parent property I can only get service name in rpcImpl like below:
`this.constructor.name`
but this can't survice after minification. That's all. Thanks for you patience!

I would like to hear from you! @alexander-fenster

alexander-fenster:
@lili2012 In the static code, indeed, there is no reflection. The way I did that a few years ago for Google Cloud client libraries, which were supposed to work both in Node and in browser, was to use `pbjs` to generate JSON representation of the proto (`pbjs -t json -o test.json test.proto`), and then load that JSON using `const root = protobuf.Root.fromJSON(jsonObj)`:

```
const root = protobuf.Root.fromJSON(require('./test.json')); // webpack would handle this
const method = root.lookup('test.Foo.Bar'); // and everything works
```

As for comparing the constructor name with `Service`, that was just for demo purposes; normally, you just need to know fully qualified RPC name (`test.Foo.Bar` in my example) to make a request, and you can get it from the `method` instance.

lili2012:
@alexander-fenster Thanks for you relpy.

I need to get Foo.Bar in rpcImpl function. 

> load that JSON using const root = protobuf.Root.fromJSON(jsonObj)

load the JSON won't help in rpcImpl function. In rpcImpl function I need to know which rpc call rpcImpl, serviceName and methodName are unknowns there. For now I could only have methodName in rpcImpl:
```
export const rpcImpl: RPCImpl = async function (method, requestData, callback) {
  const methodName = method.name
  const serviceName = this.constructor.name //in production mode service name is not available as terser mangle names
  websocket.send("serviceName/methodName", requestData)
}
```
As you can see **method** is no use here.  What we need is **serviceName** and **methodName**

alexander-fenster:
`method` is an instance of `protobuf.Method` which has all the information you need, if it comes from a resolved `protobuf.Root`. I don't think it will work with a static stub generated by `pbjs -t static`, but try loading JSON as I suggest and your puzzle pieces will all come together.

lili2012:
I compile to JSON which is incredibly small. I think I understand your point. you mean with JSON function rpcImpl method could get serviceName and methodName. Great!. But I still want to use the static code.  Do my pull request break protobuf.js JSON rpcImpl implementation? What could I do to let you accept my pull request?

alexander-fenster:
Unfortunately, as written, the code breaks all current users of `rpcImpl` because it changes the signature of the function they are supposed to implement. So, whenever this change is released, everyone who uses the current `rpcImpl` will need to go update their code. It might be possible to add new optional parameters to the existing functions, but you must not reorder parameters or introduce new parameters in the middle.

PR #2145, opened 2026-04-18
fix: add serviceName, methodName, methodIndex to rpcImpl (#1229)

lili2012:
fix issue https://github.com/protobufjs/protobuf.js/issues/1229
@alexander-fenster

### DIFF
--- a/cli/targets/static.js
+++ b/cli/targets/static.js
@@ -767,7 +767,7 @@ function buildService(ref, service) {
         push("};");
     }
 
-    service.methodsArray.forEach(function(method) {
+    service.methodsArray.forEach(function(method, methodIndex) {
         method.resolve();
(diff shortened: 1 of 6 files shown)


lili2012:
> Unfortunately, as written, the code breaks all current users of rpcImpl because it changes the signature of the function they are supposed to implement. So, whenever this change is released, everyone who uses the current rpcImpl will need to go update their code. It might be possible to add new optional parameters to the existing functions, but you must not reorder parameters or introduce new parameters in the middle.

I put serviceName, methodName, methodIndex at the end of rpcImpl function parameters. and no breaks to any users of rpcImpl. Hope you could review it and accept my pull request. Thanks!

@alexander-fenster

dcodeIO:
Likely superseded by #2286

PR #2156, opened 2026-04-24
chore(ci): Test against minimum supported and rolling LTS Node.js

dcodeIO:
Simplifies CI to test against a configurable minimum Node.js version and the rolling LTS release.

Also unblocks #2153, which needs a more recent Node.js version for the benchmark job.


alexander-fenster (approved):
That indeed simplifies things! You will probably need to edit branch protection settings to let this in.

upd. looks like everything is green and good :)

Issue #2123, opened 2026-02-19
Performance regression back in 8.1.5+ (and still in 7.5.x)


kaisermann:
Hey folks, following up on [#2063](https://github.com/protobufjs/protobuf.js/issues/2063), we ran into a similar regression and did some bisecting across 15+ versions that might help narrow things down. 

For context, we hit ~7x slower Firestore emulator tests after a _transitive upgrade_ to `7.5.x` (via `firebase-admin`). 

In `7.5.x`, the issue seems to be that [PR #2066](https://github.com/protobufjs/protobuf.js/pull/2066) added `_needsRecursiveFeatureResolution` caching to `_resolveFeaturesRecursive()`, but [`resolveAll()`](https://github.com/protobufjs/protobuf.js/blob/69cced8e00216f1aed69593187ac0c2e34807208/src/namespace.js#L343-L350) still re-traverses the entire tree on every call. The [8.x branch has the fix](https://github.com/protobufjs/protobuf.js/blob/f39bb00221b532d60be487e78fb4fbe8a59b85c0/src/namespace.js#L352-L365), but [PR #2068](https://github.com/protobufjs/protobuf.js/pull/2068) that tried to backport it got reverted due to [#2072](https://github.com/protobufjs/protobuf.js/issues/2072).

Would it be possible to reconsider a backport to `7.5.x`? This regression is pretty tricky to track down when it comes from a transitive dependency. Most people won't know to pin `protobufjs` specifically. 

---

Also, while testing the `8.x` experimental versions, we noticed that while `8.1.2-experimental` seemed to perfom as well as `7.4.0`, but `8.1.6` started to show some performance regression. Not as bad, but still observable.

In case it helps and for a rough idea of the performance differences, we tested a bunch of the 8.x versions against one of our test cases:

| Version | Duration | Notes |
| - | - | - |
| 7.4.0 | **2.5s** | baseline |
| 7.5.x | ~55s | slow |
| 8.0.0 | ~23s | slow |
| 8.0.1-experimental | **2.3s** | ✓ fast |
| 8.0.3-experimental | **2.9s** | ✓ fast |
| 8.0.4-experimental | **3.0s** | ✓ fast |
| 8.1.0-experimental | **4.3s** | ✓ acceptable |
| 8.1.1-experimental | **5.0s** | borderline |
| 8.1.3-experimental | >5s | timeout |
| 8.1.4-experimental | ~4-6s | flaky |
[...]

Issue #2063, opened 2025-04-21
Large performance regression in Root.load & loadSync in 7.5.0


simon-paris:
In 7.5.0, Root.load calls resolveAll. For each proto file that's loaded, it resolves _everything_, resulting in O(n^2) performance where n is the number of input files.

This is a very large performance regression in workflows that have more than one entry point proto file. In my project, initialization is ~100x slower (from 300ms to 30s).

Additionally, the async version calls resolveAll twice, doubling the performance hit. It appears that one call is meant to happen in the sync version and the other in the async version, but the async version actually runs both calls.

The change was made here: https://github.com/protobufjs/protobuf.js/pull/2029/commits/b802c42881af38b21cf39a9e979a351978985f8b#diff-45840cbec0137d0498a3dbcf38eba6a358e2b8bfd9f1eca20abdfc7b386d13b1R208

cc @sofisl

Nushio:
I can confirm. We use Firestore (which uses protobufjs). A commit we made yesterday updated the protobuf library from 7.4.0 to 7.5.0. 
The query time jumped from 0.4ms to 16 seconds.

mkruskal-google:
resolveAll is supposed to cache results and be efficient to call multiple times, this sounds like a bug.  I'll investigate

sofisl:
Are you calling `.resolveAll()` in your code?

sofisl:
Moving 7.4.0 to latest while this gets resolved.

mkruskal-google:
ok so there are a few problems here:

1) resolveAll *doesn't* cache most of the work it does.  It only caches it for the root object, which seems to have been a pre-existing regression we made worse
2) feature resolution is required for correctness, and needs to know about the presence of groups (which naively uses resolved type info)
3) we added a bunch of new resolveAll calls in order for resolved features to always be available

It's possible to fix 2) to use unresolved type information, and short-circuit this regression.  I think resolveAll could still probably be optimized to cache state better, but if we just don't call it in the new locations it won't make things any worse

mkruskal-google:
@simon-paris @Nushio could either of you test against 8.1.0-experimental to confirm this resolves your issues?  There's a remote possibility further optimizations are needed, but I think this will resolve the bulk of the problem

mkruskal-google:
[...]

PR #2157, opened 2026-04-24
fix: Restore lazy type resolution during load

dcodeIO:
Restores pre-editions lazy type resolution for `Root.fromJSON()`, `Root.load()`, and `Root.loadSync()` while still resolving editions features after loading.

The context here is that before editions, we did not eagerly call `resolveAll`. During work on editions, which required feature resolution, this was changed to call `resolveAll` eagerly, in turn introducing performance regressions. It has since been attempted to mitigate some of the fallout, but to me it appears that eagerly calling `resolveAll` is the root cause here. Hence this PR aims at bringing us back to pre-editions baseline by resolving only features eagerly but defer type resolution as usual.

This is technically a breaking change for anyone already relying on eager type resolution after load. As far as I can tell, that behavior was not documented, and callers that need full validation can still call `resolveAll()` explicitly.

Related to #2123.

Follow-up to #2063 and #2066.
Partially reverts the eager load-time type resolution introduced by #2068.

### DIFF
--- a/src/root.js
+++ b/src/root.js
@@ -62,7 +62,7 @@ Root.fromJSON = function fromJSON(json, root) {
         root = new Root();
     if (json.options)
         root.setOptions(json.options);
-    return root.addJSON(json.nested).resolveAll();
+    return root.addJSON(json.nested)._resolveFeaturesRecursive(root._edition);
 };
 
 /**
@@ -118,7 +118,7 @@ Root.prototype.load = function load(filename, options, callback) {
             throw err;
         }
         if (root) {
-            root.resolveAll();
+            root._resolveFeaturesRecursive(root._edition);
         }
         var cb = callback;
         callback = null;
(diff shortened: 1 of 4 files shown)


dcodeIO:
After looking into this again, I think the tradeoff is more nuanced than my original PR description suggests.

The performance argument for this change is weaker than I initially assumed. On current master, the fully resolved path appears roughly unchanged: callers doing `Root.fromJSON(...).resolveAll()` pay about the same total cost with or without this PR. The measurable win is limited to callers that only need the loaded reflection tree with editions features resolved, but do not need full type-reference resolution.

There is also a real ergonomics/correctness argument for the current eager behavior. Eager feature resolution is enough for properties such as `required`, but type-dependent properties still needs full resolution. For example, `delimited` is not fully initialized until fields are resolved. This means the current eager behavior makes `load`/`fromJSON` return a more consistently usable reflection tree.

Given that, I am no longer confident that restoring lazy type resolution is the better default.

Moved this PR back to draft for now. Additional context from the editions work would still be useful, especially around whether future editions work depends on eager type resolution rather than eager feature resolution alone.

Also cc @alexander-fenster, in case you have followed the editions work more closely than I have.

PR #2162, opened 2026-04-24
fix: Forward group end tag in lazy decode

### LINKED ISSUE #2148: decode_setup() doesn't forward the group end-tag argument
protobufjs 7.5.5 and 8.0.1 has a bug where decode_setup(), the lazy-initialization wrapper for .decode() only forwards (reader, length) but not the third e (end-tag) argument. This means on the first decode call for any type, proto2 group end markers (wire type 4) aren't recognized, causing an "invalid wire type 4" error.

There is a workaround, pre-calling setup() on every type avoids hitting decode_setup at runtime.

```
function setupAllTypes(ns: protobufjs.NamespaceBase) {
  if (ns instanceof protobufjs.Type) {
    ns.setup();
  }
  for (const nested of ns.nestedArray) {
    if (nested instanceof protobufjs.Namespace) {
      setupAllTypes(nested);
    }
  }
}
```

dcodeIO:
Makes `decode_setup` forward the internal group end-tag to the decoder.

Fixes #2148


alexander-fenster (approved):

dcodeIO:
Only contains the regression test after merging the current head, which is expected.

PR #2173, opened 2026-04-27
fix: Backport input handling and CLI fixes to 7.x

dcodeIO:
Backports recent input handling and CLI fixes to the 7.x branch. Also updates the generated fixtures and TypeScript definitions.


Tofandel:
I guess it's on this PR, peerDependencies "protobufjs" in package.json should have been bumped to 7.5.6 (the version being released) as there is an api change in utils which does not exist if using the newer version of protobufjs-cli with an older version of protobufjs

PR #1669, opened 2021-12-07
feat: support map key use Long.toString when key is Long  #1203

SzHeJason:
Pull request to resolve  #1203.
protobuf.js version: 6.11.2

PR #2186, opened 2026-04-28
fix: Consistently handle scalar map keys

### LINKED ISSUE #1203: Constructing map<int64, ...> using longHash as keys doesn't yield correct proto on the wire

### LINKED ISSUE #1652: map field with int64 type has broken key in generated encode function

### LINKED ISSUE #2087: Boolean Keys in Protobuf Maps Converted Incorrectly
### Problem description
Hi all, we found an interesting bug. Given a map field which contained a key of type bool， when using a Node.js client to send the map field to a Python server, boolean keys in map fields are incorrectly deserialized in the response. Specifically, False keys are converted to True in the returned message.

### Reproduction steps
The code snippet to help reproduce the issues is provided in [Google drive.](https://drive.google.com/drive/folders/11ktssSjoiKde6fSGXzrTaRV8SXfyTiSF?usp=sharing)
You can reproduce the issue in either of the following two ways:

**Option 1**.  Here is a simplified version of the code to reproduce the issue:
    Proto file.
```
syntax = "proto3";

package map_hotpot;

option java_multiple_files = true;
option java_package = "org.grpctest.java.common.define";
option java_outer_classname = "MapProto";

message MapPot {
 map<bool, double> bool_double_value = 1;
 map<bool, int64> bool_int_value = 2;
 map<bool, bool> bool_bool_value = 3;
 map<bool, string> bool_string_value = 4;
}

service HotpotService {
 // UNARY
 rpc UnaryPot(MapPot) returns (MapPot);
}

```
Nodejs client request message: 

```
const request = {
    bool_double_value: { true: 3.14 },
    bool_int_value: { false: 42 },
    bool_bool_value: { true: true },
    bool_string_value: { **false**: "hello" }
};

```
The response message as received by Python client

```
python server Received request: bool_double_value {
  key: true
  value: 3.14
}
bool_int_value {
  key: true
  value: 42
}
bool_bool_value {
  key: true
  value: true
}
bool_string_value {
  key: true
  value: "hello"
}

```

We can see that the **second** and **fourth** key are switched from **"false" to "true",** which is expected.

**Option 2.** For your convenience, we’ve provided the Docker image. 
`
Load the Docker image: You can access it via this command ``docker pull lishuoiscas/grpc_app:0723`
Prepare the code:
    Create a folder named grpc-auto-test
    Place the relevant code inside this folder
Run the Docker container:
Replace {your_path} with the absolute path to your local grpc-auto-test folder:
`docker run -it --rm -v {your_path}/grpc-auto-test:/grpc-auto-test <image:tag>`
Inside the container, reproduce the issue by running:

`./ls_run.sh`

### Environment
 - MacOs 15.3 and Windows 11
 - Node version 22.17.1
 -Python 3.13 

### Additional context
**Plausible reason**: We confirmed the problem originated from Node.js server by analysing the response sent from Node.js server, as captured by tcpdump. This issue stemmed from the way Node.js convert map values to Javascript objects, where map keys must be a string. We have not found sufficient gRPC-node documentation to determine the exact cause behind this issue.  However, it is very likely that the issue was caused by how the string form of map keys, as stored in Javascript objects, was converted to boolean type during serialization.  In Javascript, the string “false”, when converted to a boolean value, will be boolean true (refer to the [debuggping_process_1.png](https://drive.google.com/drive/folders/11ktssSjoiKde6fSGXzrTaRV8SXfyTiSF?usp=sharing)).   If our assumption is true, then one solution to patch this bug would be to introduce a simple condition during string to boolean conversion as shown in the [debuggping_process_2.png.](https://drive.google.com/drive/folders/11ktssSjoiKde6fSGXzrTaRV8SXfyTiSF?usp=sharing)  
Please note that the suggested change was not based on gRPC-node code, it simply suggested a general idea to fix this issue.

Can you help fix it or double check it? Thank you.

dcodeIO:
Fixes encoding and object conversion of scalar map keys that are stored internally as strings, enabling round-tripping of boolean and 64-bit map keys.

Also fixes the 64-bit map key check to accept the documented internal format.

Fixes #1203
Fixes #1652
Fixes #2087

Supersedes #1653
Supersedes #1669


src/util/minimal.js
```diff
@@ -226,6 +226,27 @@ util.longFromHash = function longFromHash(hash, unsigned) {
...
+        : key;
+};
+
+/**
+ * Converts a boolean key to a boolean value.
+ * @param {string} key Map key
+ * @returns {boolean} Boolean value
+ */
+util.boolFromKey = function boolFromKey(key) {
+    return key === "true" || key === "1";
```
alexander-fenster, line 247: or maybe even true if `Number.isInteger(key)` and it's not zero?

alexander-fenster (approved):

src/util/minimal.js
dcodeIO: Good point. Here, `key`s are always strings, so `Number.isInteger("2")` would be `false` unless we convert first. Since this currently lines up with `util.key2Re = /^true|false|0|1$/;`, I’d lean toward keeping it analogous for now.

dcodeIO:
Note: this will once more need a follow-up protobufjs-cli peer dependency bump after the next protobufjs release due to the new `$util.longFromKey` / `$util.boolFromKey` helpers.

PR #2196, opened 2026-04-28
feat: Add finishInto() for zero-copy finish into existing buffers

jiang1997:
Introduce finishInto(), which allows writing directly into an existing buffer (e.g., WebAssembly memory), thereby avoiding extra copying.

### DIFF
--- a/src/writer.js
+++ b/src/writer.js
@@ -446,7 +446,7 @@ Writer.prototype.ldelim = function ldelim() {
  * @returns {Uint8Array} Finished buffer
  */
 Writer.prototype.finish = function finish() {
-    return this.finishTo(this.constructor.alloc(this.len), 0);
+    return this.finishInto(this.constructor.alloc(this.len), 0);
 };
 
 /**
@@ -458,7 +458,7 @@ Writer.prototype.finish = function finish() {
  * @returns {T} The provided buffer
  * @template T extends Uint8Array
  */
-Writer.prototype.finishTo = function finishTo(buf, offset) {
+Writer.prototype.finishInto = function finishInto(buf, offset) {
     offset >>>= 0;
     var head = this.head.next,
         pos  = offset;
--- a/src/writer.js
+++ b/src/writer.js
@@ -446,15 +446,27 @@ Writer.prototype.ldelim = function ldelim() {
  * @returns {Uint8Array} Finished buffer
  */
 Writer.prototype.finish = function finish() {
-    var head = this.head.next, // skip noop
-        buf  = this.constructor.alloc(this.len),
-        pos  = 0;
+    return this.finishTo(this.constructor.alloc(this.len), 0);
+};
+
+/**
+ * Finishes the write operation, writing into the provided buffer.
+ * The caller must ensure that `buf` has enough space starting at `offset`
+ * to hold {@link Writer#len} bytes.
+ * @param {T} buf Target buffer
+ * @param {number} [offset=0] Offset to start writing at
+ * @returns {T} The provided buffer
+ * @template T extends Uint8Array
+ */
+Writer.prototype.finishTo = function finishTo(buf, offset) {
+    offset >>>= 0;
+    var head = this.head.next,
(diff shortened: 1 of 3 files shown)


dcodeIO:
Do you plan to use `finishTo` somewhere within the library to speed it up? If not, then this should likely not live in protobufjs source itself, but be provided externally.

jiang1997:
> Do you plan to use `finishTo` somewhere within the library to speed it up? If not, then this should likely not live in protobufjs source itself, but be provided externally.

To be honest, my initial motivation was simply to solve an issue in my own project (serializing directly into Wasm memory to avoid extra copies). I haven't really explored using it to speed up the library internally yet.
I understand that users could technically implement this externally by iterating over writer.head. However, I was concerned that this might not be the best approach, as it would require external code to break encapsulation and rely on internal implementation details.

dcodeIO:
One approach to integrate it into protobuf.js itself might be to use `finishTo` to implement `finish`. Then it's not just dead code. Ideally, though, such a change should not require a bunch of additional runtime checks that only exist because of that split.

jiang1997:
> One approach to integrate it into protobuf.js itself might be to use `finishTo` to implement `finish`. Then it's not just dead code. Ideally, though, such a change should not require a bunch of additional runtime checks that only exist because of that split.

Yeah, totally agree. I've already made the changes — finish() now delegates to finishTo(), all runtime checks have been removed, and the documentation has been updated to note that callers are responsible for ensuring sufficient buffer space.

src/writer.js
```diff
@@ -446,15 +446,26 @@ Writer.prototype.ldelim = function ldelim() {
...
+/**
+ * Finishes the write operation, writing into the provided buffer.
+ * The caller must ensure that `buf` has enough space starting at `offset`
+ * to hold {@link Writer#len} bytes.
+ * @param {Uint8Array} buf Target buffer
+ * @param {number} [offset=0] Offset to start writing at
+ * @returns {Uint8Array} The provided buffer
+ */
+Writer.prototype.finishTo = function finishTo(buf, offset) {
+    offset |= 0;
```
dcodeIO: This turns offsets >= 2 GiB negative. Perhaps:

```suggestion
    offset >>>= 0;
```

src/writer.js
```diff
@@ -446,15 +446,26 @@ Writer.prototype.ldelim = function ldelim() {
...
+    return this.finishTo(this.constructor.alloc(this.len), 0);
+};
+
+/**
+ * Finishes the write operation, writing into the provided buffer.
+ * The caller must ensure that `buf` has enough space starting at `offset`
+ * to hold {@link Writer#len} bytes.
+ * @param {Uint8Array} buf Target buffer
+ * @param {number} [offset=0] Offset to start writing at
+ * @returns {Uint8Array} The provided buffer
```
dcodeIO: Since this returns the provided buffer, I think we can make the JSDoc generic here:

```suggestion
 * @param {T} buf Target buffer
 * @param {number} [offset=0] Offset to start writing at
 * @returns {T} The provided buffer
 * @template T extends Uint8Array
```

The definitions generator should then generate `finishTo<T extends Uint8Array>(buf: T, offset?: number): T`, preserving `Buffer` and other `Uint8Array` subclasses.

src/writer.js
jiang1997: Good catch, learned something new! Updated to offset >>>= 0.

jiang1997: The offset >>>= 0 operation reduces the value to a 32-bit number. Node lets TypedArray indices go beyond 2³², so a large offset might end up near the start of the buffer.
I’d suggest making offset required, not optional, to avoid any unexpected behavior.

What are your thoughts on this, @dcodeIO ?

dcodeIO:
Thinking about naming a bit here, I guess a good name could be `finishInto`, then mirroring APIs like `TextEncoder#encodeInto`. Wdyt?

dcodeIO (approved):

Issue #1312, opened 2019-10-24
JSON mismatch between server generated JSON and client side JSON when using wrappers.


Faendrek:
First off, I apologize if this problem has been raised in another issue. I looked over them and couldn't find one that directory talked about a work around for this problem. The closest I found was https://github.com/protobufjs/protobuf.js/issues/677 . But I don't understand how to apply the suggestion of wrappers from OUR code (not protobufJS's source code).

So we use "wrapper" values such as `google.protobuf.StringValue` and `google.protobuf.DoubleValue` to handle distinguishing between null and default. We are facing two problems related to creating and sending JSON from messages that contain these wrapped values.

Given a message with the following structure:

```
message Alarm
StringValue name = 1;
DoubleValue durationInSeconds = 2;
StringValue description = 3;
```

In our server code we transmit this message as JSON on our endpoints in Python. When we create JSON from protobuf instances on the server using MessageToDict or MessageToJSON we end up with the following:

```
{
    name: 'foo',
    durationInSeconds: 34,
    description: 'boiler temperature exceeds limit'
}
```

In the UI we have the same model compiled with ProtobufJS/ProtobufTS.  There are two problems we're having.

The first is that we're not able to take this JSON payload and turn it back into an `Alarm` in the UI since the JSON structure doesn't have wrapped values. Essentially there is a mismatch between the JSON created from a message and the JSON needed to create a message.
At the very least I'd like to be able to use `fromObject` / `toObject` to consume the JSON to make a new Alarm or to produce JSON from an existing Alarm.

This issue describes what I'm talking about and someone has made a PR to allow us to write wrappers for the wrapped values (https://github.com/protobufjs/protobuf.js/issues/677#issuecomment-517997209). What I don't understand where do I include this wrapper code? Basically in our project we have a `.proto` file and we compile it with protobufJS.
Where do I define these wrappers and how can I include them when I compile our `.proto` files?

The second problem is there is no interface available in the UI TS files to describe the JSON structure. The JavaScript interface of an Alarm for instance doesn't need wrapped values. So if we want to use the JSON structure that we get from the server we have to manually create an interface for each of our messages that matches the JSON version of message.
I understand I could just use the Protobuf interface which has wrappers rather than primitives but it means dealing with the extra `.value` everywhere and not having an interface to describe the actual JSON data that will get sent back to the server for updates.

Issue #1304, opened 2019-09-27
Proto3 canonical JSON support


johncsnyder:
There are a number of open issues related to this. According to https://github.com/protobufjs/protobuf.js/issues/351:

> Not yet supported:
>
> "NaN", "Infinity" strings for floats
> Any, Timestamp, Duration, Struct, Wrappers, FieldMask, ListValue, Value, NullValue are handled like messages
> Feel free to reopen!

Any type url fix:
- https://github.com/protobufjs/protobuf.js/issues/850
- PR https://github.com/protobufjs/protobuf.js/pull/1068 seems to fix that issue

Struct, Value:
- https://github.com/protobufjs/protobuf.js/issues/839

Related:
- https://github.com/protobufjs/protobuf.js/issues/916
- https://github.com/protobufjs/protobuf.js/issues/1108

johncsnyder:
I have a working implementation that supports Timestamp and Duration well-known types here https://github.com/johncsnyder/protobuf.js/commit/a6eddb4cb81bf63833ce88121d168228f429ff49

It probably needs a bit of work before making a PR, as it does not support all the cases covered under https://developers.google.com/protocol-buffers/docs/proto3#json  and it changes the default `toJSON` functionality to output Timestamp and Duration as their string representations. Would appreciate some guidance here and how to incorporate this into protobufjs without breaking existing usage.

It creates the following wrappers
```
// Custom wrapper for Timestamp
wrappers[".google.protobuf.Timestamp"] = {
    fromObject: function(object) {
        if (typeof object === "string") {
            const ts = new Date(object);
            const seconds = Math.floor(ts.getTime() / 1000);
            const nanos = ts.getMilliseconds() * 1000000;
            return this.create({
                seconds: seconds,
                nanos: nanos
            });
        } else if (object instanceof Date) {
            const seconds = Math.floor(object.getTime() / 1000);
            const nanos = object.getMilliseconds() * 1000000;
            return this.create({
                seconds: seconds,
                nanos: nanos
            });
        }

        return this.fromObject(object);
    },

    toObject: function(message, options) {
        if (options && options.json) {
            return new Date(message.seconds * 1000 + message.nanos / 1000000);
        }
        return this.toObject(message, options);
    }
};

// Custom wrapper for Duration
wrappers[".google.protobuf.Duration"] = {
    fromObject: function(object) {
        if (typeof object === "string") {
            const unit = function() {
                if (object.slice(-1) == "s") return 1;
[...]

PR #2201, opened 2026-04-29
feat!: Support JSON conversion for scalar wrappers

dcodeIO:
Adds JSON conversion support for `google.protobuf.*Value` scalar wrapper types in converters:

* `google.protobuf.DoubleValue`
* `google.protobuf.FloatValue`
* `google.protobuf.Int64Value`
* `google.protobuf.UInt64Value`
* `google.protobuf.Int32Value`
* `google.protobuf.UInt32Value`
* `google.protobuf.BoolValue`
* `google.protobuf.StringValue`
* `google.protobuf.BytesValue`

Example:

```proto
message SomeMessage {
  google.protobuf.StringValue name = 1;
  google.protobuf.Int64Value count = 2;
  google.protobuf.BoolValue flag = 3;
  google.protobuf.BytesValue data = 4;
}
```

```js
const message = SomeMessage.fromObject({
  name: "abc",
  count: "123",
  flag: false,
  data: "AQI="
});

SomeMessage.toObject(message, {
  json: true,
  longs: String,
  bytes: String
});
// { name: "abc", count: "123", flag: false, data: "AQI=" }
```

Unlike the existing `google.protobuf.Any` wrapper in `wrappers.js`, these value wrappers are simple enough to implement at the shared codegen level, so they also apply to generated static code.

Related: #677, #1312, #1304

### DIFF
--- a/src/converter.js
+++ b/src/converter.js
@@ -118,7 +118,7 @@ converter.fromObject = function fromObject(mtype) {
     ("if(q>util.recursionLimit)")
         ("throw Error(\"max depth exceeded\")");
     if (wrapperValueRe.test(mtype.fullName)) gen
-    ("if(d==null||typeof d!==\"object\"||!Object.prototype.hasOwnProperty.call(d,\"value\"))")
+    ("if(d!=null&&typeof d!==\"object\")")
         ("d={value:d}");
     if (!fields.length) return gen
     ("return new this.ctor");
--- a/src/converter.js
+++ b/src/converter.js
@@ -9,6 +9,8 @@ var Enum  = require("./enum"),
     types = require("./types"),
     util  = require("./util");
 
+var wrapperValueRe = /^\.google\.protobuf\.(?:DoubleValue|FloatValue|Int64Value|UInt64Value|Int32Value|UInt32Value|BoolValue|StringValue|BytesValue)$/;
+
 /**
  * Generates a partial value fromObject conveter.
  * @param {Codegen} gen Codegen instance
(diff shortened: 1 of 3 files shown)


alexander-fenster (approved):
This is great; I'm a little bit concerned that it's a breaking change for those who depend on the existing format. If we are considering major version bump soon, maybe `feat!:` just to be safe?

dcodeIO:
True, this is essentially two pieces:

1. It adds support for scalar inputs where we previously only supported `{ value: ... }`. This is a regular feature, because previously unsupported inputs are now supported.
2. It changes outputs from `{ value: ... }` to plain scalar values when using `{ json: true }` or `.toJSON()`. From a conformance perspective, this is a fix, because the previous behavior was non-standard. From a compatibility perspective, this is a breaking change, because some consumers might already depend on the non-standard behavior.

So yes, as-is, marking this as `feat!:` is the safer choice. Alternatively, the PR could be split into its feature and breaking components, leaving the output change for a later major release. Thoughts?

dcodeIO:
One good argument against splitting is that there's more WKT work perspectively, so locking the mechanism in 8.x already is risky. Likely best to keep the mechanism itself for 9.x then.

alexander-fenster:
Let's just do a single `feat!`. If you are planning more patch / minor releases, this one can probably wait until the good time to go to 9.x.

I'm surely aware of at least one package that would be affected by the breaking change, it's [proto3-json-serializer](https://www.npmjs.com/package/proto3-json-serializer) that I wrote a few years ago for Google. It still depends on v7, but upgrading it to v9 will require someone to go make some (trivial) changes in [this code](https://github.com/googleapis/google-cloud-node/blob/778adb0d5f368448f99bf68dcbcf37c3a7394f5d/core/packages/proto3-json-serializer-nodejs/typescript/src/value.ts#L62) that depends on the internal representation.

Shall we probably make a branch for collecting 9.x features?

Roseymr (approved):

PR #2202, opened 2026-04-29
ci: Track protobuf conformance in CI

dcodeIO:
Adds CI coverage for the official Protocol Buffers conformance suite and uploads the generated logs as artifacts.

The workflow currently runs against upstream v33, but the generator is prepared for v34.1/v35-rc1 where required EditionUnstable cases are present when testing up to Edition 2024. This requires a workaround on our end because the parser intentionally only accepts released editions.

Also updates the README.

[Current results](https://github.com/protobufjs/protobuf.js/actions/runs/25371108781)

### DIFF
--- a/.github/actions/conformance/action.yml
+++ b/.github/actions/conformance/action.yml
@@ -1,0 +1,110 @@
+name: "Run protobuf conformance"
+description: "Build and run the upstream Protocol Buffers conformance suite for protobuf.js."
+inputs:
+  upstream-version:
+    description: "Upstream protocolbuffers/protobuf tag or branch to test against."
+    default: "v33.0"
+  upstream-dir:
+    description: "Directory where upstream protobuf is checked out."
+    default: "tests/conformance/upstream"
+  runner-cache:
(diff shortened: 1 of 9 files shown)


dcodeIO:
Scope note: I intentionally kept the README commitment limited to the current CI surface: state the conformance target, run the official suite, and make the results visible. Broader roadmap language could be premature unless maintainers have time and support to own that work.

For conformance specifically, useful follow-ups include enabling a binary threshold gate, validating newer upstream runners, and settling UTF-8 VERIFY behavior using TextEncoder/TextDecoder in a major release where stricter decoding can be treated as a breaking change.

More broadly, protobuf.js would also benefit from native ESM and BigInt, a platform support bump (minimum Node v12.11.0 for `TextEncoder#encodeInto`, or, even better, alignment with the Node LTS schedule), first-class JSON mapping, first-class support for well-known types, closer CLI alignment with `protoc`, and improved TypeScript output. None of that is required for this PR, but it would be valuable future work if someone wants to pick it up or support me in doing so.

alexander-fenster (approved):

PR #2207, opened 2026-04-29
fix: Accept URL-safe base64 input

dcodeIO:
Adds support for URL-safe base64 input when decoding bytes from plain objects or JSON, using `-` and `_` as aliases for `+` and `/` during decode. Also accepts unpadded base64 input without changing existing base64 output encoding.

### DIFF
--- a/src/util/base64.js
+++ b/src/util/base64.js
@@ -16,10 +16,9 @@ base64.length = function length(string) {
     var p = string.length;
     if (!p)
         return 0;
-    var n = 0;
-    while (--p % 4 > 1 && string.charAt(p) === "=")
-        ++n;
-    return Math.ceil(string.length * 3) / 4 - n;
+    while (p > 0 && string.charAt(p - 1) === "=")
+        --p;
+    return Math.floor(p * 3 / 4);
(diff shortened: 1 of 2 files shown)


src/util/base64.js
```diff
@@ -16,10 +16,9 @@ base64.length = function length(string) {
     var p = string.length;
     if (!p)
         return 0;
-    var n = 0;
-    while (--p % 4 > 1 && string.charAt(p) === "=")
-        ++n;
-    return Math.ceil(string.length * 3) / 4 - n;
+    while (p > 0 && string.charAt(p - 1) === "=")
```
alexander-fenster, line 19: Wondering if a regex replace of `/=+$/` would be faster here.

alexander-fenster: Or maybe not, because we are not copying the string here.

alexander-fenster (approved):

src/util/base64.js
dcodeIO: Agree, the current loop is preferable here as it avoids the temporary allocation.

Issue #296, opened 2015-07-02
Can unknown fields be re-encoded?


vgatto:
Is it possible to decode a serialized message and then re-encode it without losing information if you do not have the same version of the proto file that initially serialized it?  Example:

```
var ProtoBuf = require("protobufjs");
var builder = ProtoBuf.loadProto(
  "  message MessageV1 {" +
  "    optional string name = 1;" +
  "  }" +
  "  message MessageV2 {" +
  "    optional string name = 1;" +
  "    optional string value = 2;" +
  "  }");

var MessageV1 = builder.build("MessageV1");  // Server version
var MessageV2 = builder.build("MessageV2");  // Client version

// Client encodes a message using the newer version.
var message2 = new MessageV2({name: "foo", value: "bar"});
var encodedMessage2 = message2.encode().toBuffer();
console.log(encodedMessage2);
// Prints: <Buffer 0a 03 66 6f 6f 12 03 62 61 72>

// Server decodes the message using the older version.
var decodedMessage2 = MessageV1.decode(encodedMessage2);
console.log(decodedMessage2);
// Prints: { name: 'foo' }

// Server re-encodes to send back to the client:
var reencoded = decodedMessage2.encode().toBuffer();
console.log(reencoded);
// Prints: <Buffer 0a 03 66 6f 6f>
```

In other protobuf implementations, the objects hang on to the unknown tags specifically so they can re-encode the data without losing information. Can ProtoBuf.js do this using some syntax other than the above?

dcodeIO:
I am sorry to disappoint you, but this is not supported currently.

wegylexy:
Any chance of supporting this, essential for a light-weight utility app to quick touch just a few fields in a huge message, or for forward compatibility?

wegylexy:
Never mind, I resorted to use .NET browser-wasm with official `Google.Protobuf`.

PR #2209, opened 2026-04-30
feat: Roundtrip unknown fields

### LINKED ISSUE #775: add unknown field support

### LINKED ISSUE #1619: How to write raw fields directly into writer?
protobuf.js version: 6.11.2

In order to match proto3 semantics, I need to write unknown fields onto the wire. Right now, I capture unknown fields as a Uint8Array slice that includes the key and contents. I want to be able to write the Uint8Array slice directly into the protobuf writer.

```ts
  export const encode = (w: pbjs.Writer, m: TestAllTypesProto3): pbjs.Writer => {
    if (m.unknownFields !== undefined)        { 
      for (const u of m.unknownFields) { 
        w.rawBytes(u); 
      }
    }
  };
```

`w.bytes` is close but it prepends the length which I don't want.

dcodeIO:
Preserves unknown fields encountered during binary decoding and writes them back during encoding, enabling forward-compatible binary roundtrips.

Unknown field data is stored on messages as a non-enumerable, user-accessible `$unknowns` property. Deleting `$unknowns` discards preserved unknown fields explicitly. Also adds `Reader#raw` and `Writer#raw` helpers for reading and writing raw wire bytes directly.

Fixes #775
Fixes #1619 via `Writer#raw`

Supersedes #779
Supersedes #1912

Related: #296

### DIFF
--- a/bench/data/static_pbjs.js
+++ b/bench/data/static_pbjs.js
@@ -32,6 +32,9 @@ $root.Test = (function() {
             $root.Test.Inner.encode(message.inner, writer.uint32(26).fork()).ldelim();
         if (message.float != null && Object.hasOwnProperty.call(message, "float"))
             writer.uint32(37).float(message.float);
+        if (message.$unknowns != null && Object.hasOwnProperty.call(message, "$unknowns"))
+            for (var i = 0; i < message.$unknowns.length; ++i)
+                writer.raw(message.$unknowns[i]);
         return writer;
     };
 
@@ -44,6 +47,7 @@ $root.Test = (function() {
             throw Error("max depth exceeded");
         var end = length === undefined ? reader.len : reader.pos + length, message = _target || new $root.Test();
         while (reader.pos < end) {
+            var start = reader.pos;
             var tag = reader.uint32();
             if (tag === _end) {
                 _end = undefined;
@@ -68,6 +72,8 @@ $root.Test = (function() {
                 }
             default:
                 reader.skipType(tag & 7, _depth, tag >>> 3);
+                $util.makeProp(message, "$unknowns", false);
+                (message.$unknowns || (message.$unknowns = [])).push(reader.raw(start, reader.pos));
                 break;
             }
         }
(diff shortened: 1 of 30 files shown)


alexander-fenster (approved):

dcodeIO:
Note: CI error after merge indicates overlap with #2208:

```
not ok 1201 should omit default string
```

PR #2215, opened 2026-05-01
docs: Modernize README

dcodeIO:
Modernizes the README structure and introductory usage flow, documenting runtime variants, message APIs, common workflows, and advanced features more concisely.

As a follow-up, it would likely be useful to replace or complement the current performance section with automated conformance reporting. For reference, a local run against the upstream v33 conformance runner, excluding TextFormat, passed 4979/5567 cases overall: 100% Binary and 78.7% ProtoJSON, with proto2, proto3, and editions 2023/2024 covered. The remaining differences are in ProtoJSON/WKT behavior and appear to require a few breaking API changes.

Also fixes #2046 by updating the moved example instead

Preview: [README.md](https://github.com/protobufjs/protobuf.js/blob/patch/readme/README.md), [cli/README.md](https://github.com/protobufjs/protobuf.js/blob/patch/readme/cli/README.md)


README.md
```diff
@@ -9,719 +9,354 @@
...
 
-// Continue at "Create a new message" above
-```
+If required fields are missing while decoding proto2 data, `decode` throws `protobuf.util.ProtocolError` with the partially decoded message available as `err.instance`.
 
-### Using services
+## Schemas and code generation
 
-The library also supports consuming services but it doesn't make any assumptions about the actual transport channel. Instead, a user must provide a suitable RPC implementation, which is an asynchronous function that takes the reflected service method, the binary request and a node-style callback as its parameters:
+### Loading JSON descriptors
```
alexander-fenster: Maybe worth mentioning somewhere how to get them (`pbjs -t json`).

dcodeIO: Good point. I decided to go a little further and modernize the CLI README as well. The main README is now the canonical documentation entry point, while the CLI README focuses on command help and links back for broader usage docs. The former “JSON descriptors” wording is now covered as “JSON bundles” in the main README.

alexander-fenster (approved):

PR #2225, opened 2026-05-04
fix: Tighten key2Re regex with non-capturing group

jiang1997:
Add non-capturing group so verify() only matches canonical bool keys: "true", "false", "0", "1".


dcodeIO (approved):

dcodeIO:
Thanks!

PR #2226, opened 2026-05-05
fix: Drop util.inquire and inline optional Node module lookups

BridgeAR:
Inlining `@protobufjs/inquire` into `src/util/inquire.js` (#2146) replaced the eval-based dynamic `require` with a plain `require(moduleName)`. webpack, rspack, and esbuild treat that as a context dependency rooted under `src/util/`, which excludes Node built-ins, so `inquire("fs")` resolves to `null` at runtime. Bundles crash on the first `Root.loadSync` call:

    TypeError: Cannot read properties of null (reading 'readFileSync')
        at fetch (.../protobufjs/index.js)

The four internal call sites (`util.fs`, the `fs` lookup in `util.fetch`, `util.Buffer`, and `util.Long`) now require their target modules with literal strings inside try/catch. `buffer` and `long` resolve as npm packages in browser bundles; `fs` carries a `webpackIgnore` magic comment so webpack/rspack leave the require for the runtime, while other bundlers fall through to the catch.

Removing `util.inquire` is a public-API change. Keeping a helper whose only job is to dodge bundlers is what produced this regression, so the export, its TypeScript declaration, and the tests that exercised it go with it. A back-compat wrapper can be reintroduced if a deprecation cycle is preferred.

Refs: https://github.com/protobufjs/protobuf.js/pull/2146
Refs: https://github.com/protobufjs/protobuf.js/issues/1754


src/util/fetch.js
```diff
@@ -1,10 +1,14 @@
...
 module.exports = fetch;
 
-var asPromise = require("./aspromise"),
-    inquire   = require("./inquire");
+var asPromise = require("./aspromise");
 
-var fs = inquire("fs");
+var fs = null;
+try {
+    fs = require(/* webpackIgnore: true */ "fs");
```
dcodeIO, line 8: Do you know if this annotation is webpack-only, or if other bundlers/runtimes understand it as well? Generally, I wonder if there is a similar but more general mechanism that applies more broadly. One mechanism that comes to mind is the `browser` field in package.json, in case that's applicable here.

BridgeAR: No, I am unsure about that, sorry

PR #2229, opened 2026-05-06
docs: Clarify security policy

dcodeIO:
Clarifies the security policy for transparent disclosure and untrusted schema input, including that malformed or adversarial `.proto` files and JSON descriptors are included in the supported input surface for active release lines.


alexander-fenster (approved):
Yeah :)

PR #2231, opened 2026-05-07
feat(cli): Add optional protoc-gen-pbjs plugin

dcodeIO:
Adds an optional `protoc-gen-pbjs` entrypoint to `protobufjs-cli` for projects that already use `protoc`. The plugin reads `CodeGeneratorRequest`s and reuses the existing `pbjs` logic to emit aggregate JavaScript modules and, when `dts` is specified, matching TypeScript declarations.

Also adds the compiler plugin schema to the CLI package, documents usage, and adds plugin tests.


alexander-fenster (approved):
That's cool! The need to commit the whole JSON representation of `plugin.proto` is unfortunate, but I guess there are no other good options, I did the same thing some time ago in a different project.

Do you think the protos belong to `./cli/google` or maybe to `./google`? That would make it easier for protobuf.js users to write their own `protoc` plugins without having their own copy of `plugin.proto`.

PR #2232, opened 2026-05-07
fix: Improve descriptor extension interoperability

dcodeIO:
Improves `ext/descriptor` interoperability by fixing descriptor input handling for buffers/readers, preserving `proto3_optional` correctly, and reconstructing descriptor map fields as `MapField`s.

Also moves the descriptor extension one level up for better modern ESM/TypeScript consumption with `.js` import specifiers, keeps a backwards-compatible shim folder, and improves its TypeScript declarations. The declarations now properly augment the parent protobuf.js types when the descriptor extension is used.


alexander-fenster (approved):

jtbandes:
Hi @dcodeIO - the change in map field deserialization from `[{k,v}, ...]` to `{k: v, ...}` makes sense, but was somewhat of a breaking change. Would it be possible to include an option to use the old behavior of treating them as repeated key/value pairs instead?

dcodeIO:
Potential workaround:

```js
const protobuf = require("protobufjs");
const descriptor = require("protobufjs/ext/descriptor");

function stripMapEntries(d) {
    if (!d) return;
    if (d.nestedType)
        for (const n of d.nestedType) {
            if (n.options && n.options.mapEntry) n.options.mapEntry = false;
            stripMapEntries(n);
        }
    if (d.messageType) for (const m of d.messageType) stripMapEntries(m);
    if (d.file) for (const f of d.file) stripMapEntries(f);
}

const fileSet = descriptor.FileDescriptorSet.decode(bytes);
stripMapEntries(fileSet);
const root = protobuf.Root.fromDescriptor(fileSet);
```

PR #2235, opened 2026-05-08
fix: Restore first-match namespace lookup

### LINKED ISSUE #2234: 7.5.1 wrong type resolution for same-name nested messages
# [Bug] `_lookupImpl` missing `break` causes wrong type resolution for same-name nested messages

## Environment

- **protobufjs version**: 7.5.1
- **Works in**: 7.5.0
- **Broken in**: 7.5.1

## Description

In protobufjs 7.5.1, the refactored `Namespace.prototype._lookupImpl` function in `src/namespace.js` has a missing `break` statement in the nested namespace search loop. This causes the **last** matching type to be returned instead of the **first**, violating protobuf's name resolution semantics and leading to `invalid wire type` errors at decode time.

## Root Cause

In 7.5.0, the `lookup` function returned immediately upon finding the first match:

```javascript
// 7.5.0 behavior (correct)
if (this._nestedArray[i] instanceof Namespace && (found = this._nestedArray[i].lookup(path)))
    return found;  // returns FIRST match
```

In 7.5.1, the lookup was refactored into `_lookupImpl` with caching, but the loop continues iterating after finding a match:

```javascript
// 7.5.1 behavior (buggy)
for (var i = 0; i < this.nestedArray.length; ++i)
    if (this._nestedArray[i] instanceof Namespace && (found = this._nestedArray[i]._lookupImpl(path)))
        exact = found;  // NO break — loop continues, LAST match wins!
```

The incorrect result is then cached in `_lookupCache`, causing all subsequent lookups for the same name to also return the wrong type.

## Minimal Reproduction

```proto
syntax = "proto2";
package example;

// Message A has a nested Point3D with float fields (4 bytes each)
message PerceptionObstacle {
    message Point3D {
        optional float x = 1;
        optional float y = 2;
        optional float z = 3;
    }
    optional Point3D velocity = 1;  // correctly resolves to nested float version
}

// Top-level Point3D with double fields (8 bytes each)
message Point3D {
    optional double x = 1;
    optional double y = 2;
    optional double z = 3;
}

// Accessory references Point3D — should resolve to top-level (double) version
message Accessory {
    optional Point3D position = 1;
    optional Point3D velocity = 2;
}
```

```javascript

const root = protobuf.loadSync('test.proto');
const accessoryType = root.lookupType('example.Accessory');

// Check which Point3D was resolved for Accessory.position
const positionField = accessoryType.fields.position;
console.log('Accessory.position resolvedType:', positionField.resolvedType.fullName);
// Expected: .example.Point3D (double fields)
// Actual (7.5.1 bug): .example.PerceptionObstacle.Point3D (float fields)
```

**Result in 7.5.1**: `Accessory.position` resolves to `.example.PerceptionObstacle.Point3D` (float, 4 bytes per field) instead of `.example.Point3D` (double, 8 bytes per field).

**Impact**: When decoding binary data, the decoder reads 4 bytes (float) instead of 8 bytes (double), causing the reader position to drift. Eventually the reader encounters a byte whose lower 3 bits form an invalid wire type (e.g., 6), throwing:

```
RangeError: invalid wire type 6 at offset XXXX
```

## Fix

Add `break` after the first match in `_lookupImpl`:

```diff
--- a/src/namespace.js
+++ b/src/namespace.js
@@ -428,8 +428,10 @@ Namespace.prototype._lookupImpl = function lookup(path) {
     // Otherwise try each nested namespace
     } else {
         for (var i = 0; i < this.nestedArray.length; ++i)
-            if (this._nestedArray[i] instanceof Namespace && (found = this._nestedArray[i]._lookupImpl(path)))
+            if (this._nestedArray[i] instanceof Namespace && (found = this._nestedArray[i]._lookupImpl(path))) {
                 exact = found;
+                break;
+            }
     }
```

This restores the "first match wins" behavior from 7.5.0, which is consistent with protobuf's scope resolution rules.

## Additional Context

This regression was introduced in the `_lookupImpl` + caching refactor (PR #2066). The caching logic itself is fine — the only issue is the missing `break` that changes the semantics from "first match" to "last match" when multiple nested namespaces contain a type with the same name.

dcodeIO:
Fixes #2234

### DIFF
--- a/src/namespace.js
+++ b/src/namespace.js
@@ -467,8 +467,10 @@ Namespace.prototype._lookupImpl = function lookup(path, flatPath) {
     // Otherwise try each nested namespace
     } else {
         for (var i = 0; i < this.nestedArray.length; ++i)
-            if (this._nestedArray[i] instanceof Namespace && (found = this._nestedArray[i]._lookupImpl(path, flatPath)))
+            if (this._nestedArray[i] instanceof Namespace && (found = this._nestedArray[i]._lookupImpl(path, flatPath))) {
                 exact = found;
+                break;
+            }
     }
 
     // Set this even when null, so that when we walk up the tree we can quickly bail on repeated checks back down.
--- a/tests/api_namespace.js
+++ b/tests/api_namespace.js
@@ -114,6 +114,27 @@ tape.test("reflected namespaces", function(test) {
     test.equal(specialRoot.lookup("__proto__"), null, "should ignore reserved nested object names");
     test.equal(Object.getPrototypeOf(specialRoot._lookupCache), null, "should isolate lookup cache keys");
 
+    var fallbackRoot = protobuf.Root.fromJSON({
+        nested: {
+            A: {
+                nested: {
+                    Point3D: { fields: {} }
+                }
+            },
+            C: {
+                nested: {
+                    Point3D: { fields: {} }
+                }
+            },
+            B: {
+                fields: {
+                    p: { type: "Point3D", id: 1 }
+                }
+            }
+        }
+    }).resolveAll();
+    test.equal(fallbackRoot.lookupType("B").fields.p.resolvedType.fullName, ".A.Point3D", "should preserve legacy first-match nested fallback lookup");
+
     ns = protobuf.Namespace.fromJSON("My", {
         nested: {
             Message: { fields: {} },


bestycw:
Thanks for the fix；What is the estimated release date of the new version？

dcodeIO:
Probably: For 7.x likely soon after approval, for 8.x alongside #2194 once remaining PRs are collected.

alexander-fenster (approved):

PR #2240, opened 2026-05-09
fix: Harden parser input handling

dcodeIO:
Adds more recursion-limit guards for loading schema definitions from JSON descriptors and `.proto` text.

### DIFF
--- a/src/namespace.js
+++ b/src/namespace.js
@@ -29,11 +29,15 @@ var Type,    // cyclic
  * @function
  * @param {string} name Namespace name
  * @param {Object.<string,*>} json JSON object
+ * @param {number} [depth] Depth of recursion to control nested calls; 0 if omitted
  * @returns {Namespace} Created namespace
  * @throws {TypeError} If arguments are invalid
  */
-Namespace.fromJSON = function fromJSON(name, json) {
-    return new Namespace(name, json.options).addJSON(json.nested);
+Namespace.fromJSON = function fromJSON(name, json, depth) {
+    if (depth === undefined) depth = 0;
+    if (depth > util.recursionLimit)
+        throw Error("max depth exceeded");
+    return new Namespace(name, json.options).addJSON(json.nested, depth);
 };
 
 /**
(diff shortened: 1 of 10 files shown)


src/namespace.js
```diff
@@ -191,9 +195,13 @@ Namespace.prototype.toJSON = function toJSON(toJSONOptions) {
 /**
  * Adds nested objects to this namespace from nested object descriptors.
  * @param {Object.<string,AnyNestedObject>} nestedJson Any nested object descriptors
+ * @param {number} [depth] Depth of recursion to control nested calls; 0 if omitted
  * @returns {Namespace} `this`
  */
-Namespace.prototype.addJSON = function addJSON(nestedJson) {
+Namespace.prototype.addJSON = function addJSON(nestedJson, depth) {
+    if (depth === undefined) depth = 0;
```
alexander-fenster: could probably use `checkDepth` here and above too if you move it to util.

src/parse.js
```diff
@@ -341,14 +349,14 @@ function parse(source, root, options) {
                 }
                 /* eslint-disable no-warning-comments */
                 // TODO: actually enforce visiblity modifiers like protoc does.
-                return parseCommon(parent, token);
+                return parseCommon(parent, token, depth);
```
alexander-fenster, line 345: +1?

src/parse.js
```diff
@@ -341,14 +349,14 @@ function parse(source, root, options) {
...
 
             case "service":
-                parseService(parent, token);
+                parseService(parent, token, depth + 1);
                 return true;
 
             case "extend":
-                parseExtension(parent, token);
+                parseExtension(parent, token, depth);
```
alexander-fenster, line 352: +1? or I might be missing a piece of logic here

alexander-fenster (approved):

src/parse.js
dcodeIO: Yeah, here, `depth` is our logical nesting level of the input structure, not necessarily the number of stack frames. Not incrementing here gives us some ~200 stack frames at 100 levels of nesting, which should still be fine but makes depth easier to reason about than just calls.

Edit: Or perhaps not, not sure what's actually easier to reason about. Preferences?

alexander-fenster: OK, makes sense!

src/namespace.js
dcodeIO: Consolidated these as per your suggestion, as `util.checkDepth`, but decided not to move it all the way down to the minimal runtime. Is now included from light upwards, which leaves Reader with one manual check, and codegen with what should likely remain inline checks.

PR #2249, opened 2026-05-12
fix: Preserve explicit enum zero if not the default

### LINKED ISSUE #2247: Regression in 8.2.0: explicit enum value 0 on the wire is decoded as typeDefault when typeDefault is non-zero
**Summary**

In `protobufjs@8.2.0`, decoding an enum field whose wire payload carries the explicit value `0` returns the enum's `typeDefault` instead, when that default is non-zero (e.g. a proto2 enum whose first declared value is negative).

This is a regression from the entire `8.0.x` line. Other explicit enum values decode correctly — only `0` is silently rewritten.

**Minimal reproduction**

```js
const protobufjs = require("protobufjs");

const root = protobufjs.Root.fromJSON({
  nested: {
    Op: { values: { UNKNOWN: -1, INSERT: 0, UPDATE: 1, DELETE: 2 } },
    M:  { fields: { op: { type: "Op", id: 4 } } }
  }
});
const M = root.lookupType("M");

// Wire: field 4 varint, value 0  →  explicit "op = 0"
const buf = Buffer.from([0x20, 0x00]);
console.log(M.decode(buf).op);
```

**Observed**

| Version | `M.decode([0x20, 0x00]).op` |
|---|---|
| 8.0.0 | `0` ✅ |
| 8.0.1 | `0` ✅ |
| 8.0.2 | `0` ✅ |
| 8.0.3 | `0` ✅ |
| 8.1.x-experimental (all) | `0` ✅ |
| **8.2.0** | **`-1`** ❌ |

The other explicit values (`1`, `2`, …) decode correctly in every version. Only the wire value `0` is dropped in 8.2.0:

```
=== protobufjs 8.2.0 ===
  op=0 explicit  hex=[20 00] → op = -1   ← BUG
  op=1 explicit  hex=[20 01] → op = 1
  op=2 explicit  hex=[20 02] → op = 2
  no op field    hex=[]      → op = -1   (typeDefault, fine)
```

If the first declared enum value is `0` (so `typeDefault === 0`), the bug does not surface. It only triggers when `typeDefault !== 0` and the wire carries `0`.

**Expected**

The wire explicitly encodes `0`, so the decoded value should be `0` regardless of what `typeDefault` is. This matched 8.0.x behavior and matches every other protobuf decoder I checked.

**Suspected cause**

Between 8.0.3 and 8.2.0, the decoder gained an "implicit presence" code path (around #2197 / #2208). The new path appears to compare the post-decode value against `0` (or against `typeDefault`) and skip the assignment — which is fine when `typeDefault === 0`, but discards a legitimate explicit `0` when `typeDefault` is something else.

`#2208` is documented as an encoder-only change, so the behavior change observed here on the decode path looks unintentional.

**Notes**

- `8.2.0` is a minor bump from `8.0.x` and the CHANGELOG marks every relevant change as a "Bug Fix"; no breaking change or migration guidance is documented for this behavior.
- No `MIGRATION.md` / `UPGRADING.md` exists in the repo, and the `README` does not mention any v8.x decode behavior change.
- `8.1.x` was officially skipped per CHANGELOG; the regression appears in the first stable release after `8.0.3`.

**Environment**

- protobufjs: 8.2.0 (regression vs. all 8.0.x)
- node: v22.22.0

dcodeIO:
Preserves explicit wire value `0` for implicit enum fields whose schema default is non-zero. This restores 8.x behavior for currently accepted non-conformant schemas, adding a note to revisit in a major release.

Fixes #2247

### DIFF
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -120,7 +120,6 @@ module.exports = [
             "no-useless-return": 1,
             "no-useless-assignment": 0,
             "no-void": 1,
-            "no-warning-comments": 1,
             "no-with": 1,
             "radix": 1,
             "vars-on-top": 0,                   // makes code harder to read, not faster
--- a/src/decoder.js
+++ b/src/decoder.js
@@ -156,7 +156,11 @@ function decoder(mtype) {
             ("case %i:{", field.id)
                 ("if(u!==%i)", types.basic[type])
                     ("break");
-            if (type === "string" || type === "bytes") gen
+            if (field.resolvedType instanceof Enum && field.typeDefault !== 0) gen
+                // TODO: Protoc rejects open enums whose first value is not zero.
+                // We should do the same, but for v8 this would be a regression.
+                ("if((v=r.%s())!==%j)", type, field.typeDefault);
+            else if (type === "string" || type === "bytes") gen
                 ("if((v=r.%s()).length)", type);
             else if (types.long[type] !== undefined) gen
                 ("if(typeof(v=r.%s())===\"object\"?v.low||v.high:v!==0)", type);
--- a/tests/api_type.js
+++ b/tests/api_type.js
@@ -264,6 +264,28 @@ tape.test("decode known fields by wire type", function(test) {
     test.end();
 });
 
+// TODO: Protoc rejects open enums whose first value is not zero.
+// We should do the same, but for v8 this would be a regression.
+// Remove this test once we enforce this restriction.
+tape.test("decode implicit enum zero with non-zero default", function(test) {
+    var Message = protobuf.Root.fromJSON({
+        nested: {
+            Op: { values: { UNKNOWN: -1, INSERT: 0 } },
+            Message: {
+                fields: {
+                    op: { type: "Op", id: 4 }
+                }
+            }
+        }
+    }).lookupType("Message");
+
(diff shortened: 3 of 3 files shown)


alexander-fenster (approved):

zbinlin:
Have same issue in integer:


```javascript
const root = protobufjs.Root.fromJSON({
  nested: {
    Error: {
      fields: {
        code: {
          type: "int32",
          id: 1
        },
      }
    },

    SubRsp: {
      fields: {
        error: {
          type: "Error",
          id: 1
        }
      }
    },
  },
});
const T = root.lookupType("SubRsp");
const ob = T.encode(T.create({
  error: {
    code: 0,
  },
})).finish();
console.log(ob.buffer.slice(ob.byteOffset, ob.byteOffset + ob.byteLength));
// <0a 02 08 00>

const buf = Buffer.from([0x0a, 0x02, 0x08, 0x00]);
console.log(T.toObject(T.decode(buf)));
/**
 * actual: { error: {} }
 * expected: { error: { code: 0 } }
 */
```

dcodeIO:
This seems to be missing `toObject(theMessage, { defaults: true })`. Is less clear to me, need to check.

dcodeIO:
Related: https://github.com/protobufjs/protobuf.js/issues/2250

PR #2261, opened 2026-05-18
fix: Expose Field rule declaration

dcodeIO:
Supersedes #1606

### DIFF
--- a/src/field.js
+++ b/src/field.js
@@ -11,7 +11,7 @@ var Enum  = require("./enum"),
 
 var Type; // cyclic
 
-var ruleRe = /^(required|optional|repeated)$/;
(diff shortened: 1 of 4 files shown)


src/field.js
```diff
@@ -11,7 +11,7 @@ var Enum  = require("./enum"),
 
 var Type; // cyclic
 
-var ruleRe = /^required|optional|repeated$/;
+var ruleRe = /^(required|optional|repeated)$/;
```
alexander-fenster: nit: `(?:   )` since we don't use grouping – but that was a fun bug in the original regex!

alexander-fenster (approved):

PR #2257, opened 2026-05-17
feat: Support BigInt conversions

dcodeIO:
With Long.js nowadays supporting BigInt construction and conversion, this upgrades Long.js and wires `{ longs: BigInt }` into conversion options. Note that for 8.x this affects `fromObject` and `toObject` conversions, whereas switching the runtime representation to BigInt is scheduled for 9.x.

Supersedes #1908
Related: #1151, grpc/grpc-node#3047

PR #2265, opened 2026-05-18
fix: Treat fixed64 as unsigned in converters

dcodeIO:
See https://github.com/protobufjs/protobuf.js/pull/2257#issuecomment-4483222410

Fixes #1344

### DIFF
--- a/cli/targets/static.js
+++ b/cli/targets/static.js
@@ -634,7 +634,7 @@ function buildType(ref, type) {
                     + JSON.stringify(field.typeDefault.low) + ","
                     + JSON.stringify(field.typeDefault.high) + ","
                     + JSON.stringify(field.typeDefault.unsigned)
-                + ") : " + field.typeDefault.toNumber(field.type.charAt(0) === "u") + ";");
+                + ") : " + field.typeDefault.toNumber(field.type === "uint64" || field.type === "fixed64") + ";");
         else if (field.bytes) {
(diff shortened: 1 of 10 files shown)


jtbandes:
I was trying to do this as well, and ran into an issue where [this mutation](https://github.com/protobufjs/protobuf.js/blob/6447a4325f49452df652023cae9dc6209a76cad4/src/converter.js#L72) (setting `unsigned=true` after construction) results in a long object that doesn't match my expected one. I'm not familiar enough with long.js to know if this is a real issue, but it made my naive test fail, so wanted to call it to your attention:

```ts
var msgFromNumber = Message.fromObject({
    fixed64Val: 11000000000000000000,
    sfixed64Val: -9000000000000000000,
});
test.same(msgFromNumber.fixed64Val, protobuf.util.Long.fromString("11000000000000000000", true), "longs from numbers");
```

```
# converters - Message.fromObject - longs
not ok 80 longs from numbers
  ---
    operator: deepEqual
    expected: |-
      Long { low: 827064320, high: -1733830216, unsigned: true }
    actual: |-
      Long { low: -1, high: 2147483647, unsigned: true }
```

alexander-fenster (approved):

dcodeIO:
That was indeed a bug. The code there shouldn't manually set `.unsigned` for two reasons: first, it's too late because conversion already took the signed route, and second, it's bad practice because doing so can modify a cached instance like `Long.ZERO`.

PR #2294, opened 2026-05-30
fix: Backport consistency and correctness fixes (7.x)

dcodeIO:
Backports #2287, #2293 and #2275 to 7.x.

### DIFF
--- a/bench/data/static_pbjs.js
+++ b/bench/data/static_pbjs.js
@@ -5,7 +5,7 @@ var $protobuf = require("../../minimal");
 
 var $Reader = $protobuf.Reader, $Writer = $protobuf.Writer, $util = $protobuf.util;
 
-var $root = $protobuf.roots.test_bench || ($protobuf.roots.test_bench = {});
+var $root = $protobuf.roots["test_bench"] || ($protobuf.roots["test_bench"] = {});
 
 $root.Test = (function() {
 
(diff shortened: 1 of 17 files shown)


alexander-fenster (approved):

UnderKoen:
Hi @dcodeIO , I think backporting https://github.com/protobufjs/protobuf.js/pull/2287 is incorrect as failing for `Message.fromObject(null)` is a change in contract and should not happen in a minor version change. 

This causes errors for us in nestjs where the null value cannot be converted into `google.protobuf.Empty`. 

I feel like returning `null` or `{}` is approperiate especially for `google.protobuf.Empty`

dcodeIO:
Did a quick check, and you are right that this used to work for fieldless messages. That behavior was accidental, though: `null` is not an empty message; `{}` is.

If needed, a narrow compatibility fix here could be to allow `null` precisely for fieldless messages<s>, but I am not yet convinced that preserving this accidental behavior of accepting an invalid value is the right tradeoff?</s> Turns out this is a relatively simple fix.

UnderKoen:
For my use case it doesn't really matter what is return, just as long as it does not throw an error.

It may be that @grpc/proto-loader should handle the null case better. 

We ran into this problem because of our nestjs implementation where we implement function which return `google.protobuf.Empty` as void an there for it returns undefined into the flow and throws :/

PR #2297, opened 2026-06-02
feat: Add spec-compliant protojson extension

### LINKED ISSUE #677: Automatically wrapping & unwrapping wrappers.proto types?

### LINKED ISSUE #839: Well-known types support (Struct, Value)

### LINKED ISSUE #916: Reading from the proto3 json standard
Is there any support for or are there any plans to support reading from the [proto3 json format](https://developers.google.com/protocol-buffers/docs/proto3#json)? #351 seems to be about writing to this format. I am storing protos generated from another language in this format and would like to be able to read it in.

### LINKED ISSUE #1108: proto3 JSON serialization
Hello, I was wondering if this library can generate a JS/TS client that uses JSON serialization instead of binary format? the goal is to have very lightweight client, and make it compatible with twirp. thanks

### LINKED ISSUE #1304: Proto3 canonical JSON support

### LINKED ISSUE #1312: JSON mismatch between server generated JSON and client side JSON when using wrappers.

### LINKED ISSUE #1922: Parsing camelCase and snake_case JSON objects into protobuf.js messages
Google's JSON parsing functions, such as [C++'s `JsonStringToMessage()`](https://protobuf.dev/reference/cpp/api-docs/google.protobuf.util.json_util/), accept [JSON keys as camelCase](https://protobuf.dev/programming-guides/proto3/#json) even if the backing field is snake_case.

By contrast, protobuf.js (AFAICT) only supports one format or the other, depending on how the `.proto` is initially [parsed](https://github.com/protobufjs/protobuf.js/blob/42e5a9ca85044800b16e193020e1d4d2e6b4010c/index.d.ts#L1069). This is WAI as discussed in issues like [this one](https://github.com/protobufjs/protobuf.js/issues/895).

I am moving an operation from C++ to TypeScript, and therefore needing to parse a JSON object that _may or may not_ contain snake_case fields using protobuf.js. This was previously not an issue (we can all agree it's not ideal) because C++'s parser supports both. However I cannot see a reasonable way to do the same with protobuf.js, because `encode()` and `fromObject()` only look for keys with the configured casing.

I am considering using [camelcase-keys](https://www.npmjs.com/package/camelcase-keys) to force my JSON objects to camelCase before passing them to protobuf.js, but this is brittle (notably, it mangles keys that are actually [map entries](https://protobuf.dev/programming-guides/proto3/#maps), which could only be detected by a protobuf-aware parser).

I'm wondering what my options are here. Is there a both-cases-compatible `fromJSON()` operation I've missed? Do I need to implement my own code that traverses the message structure and corrects appropriate keys? Or is this something that could be added to this library?

dcodeIO:
Adds a spec-compliant ProtoJSON extension derived from [proto3-json-serializer](https://www.npmjs.com/package/proto3-json-serializer). Cc @alexander-fenster as the original author.

ProtoJSON is exposed via `ext/protojson` with explicit `fromJson`, `fromJsonString`, `toJson`, and `toJsonString` APIs, plus an optional `install()` helper for reflected `Type` convenience methods.

Example usage:

```js
import protobuf from "protobufjs";
import protojson from "protobufjs/ext/protojson.js";

const root = ...;
const MyType = root.lookupType("MyType");
const message = protojson.fromJson(MyType, { value: 1 });
const json = protojson.toJson(MyType, message);
```

Also updates conformance reporting and README documentation to show binary, ProtoJSON, and Text Format coverage.

A future major release may move some of this functionality into the main library, but keeping it as an extension is the cleanest way to support ProtoJSON for now.

Fixes #677
Fixes #839
Fixes #916
Fixes #1108
Fixes #1304
Fixes #1312
Fixes #1922

Supersedes #1265

Also covers the reflection/runtime ProtoJSON parts of #2201. Static-module integration may become relevant again if we decide to move ProtoJSON support closer to the main library.

### DIFF
--- a/ext/protojson.js
+++ b/ext/protojson.js
@@ -456,7 +456,7 @@ function toJsonValue(type, message, options, depth) {
                 arr[j] = writeSingular(field, value[j], options, depth);
             setOwn(out, key, arr);
         } else {
-            if (!hasOwn(message, field.name))
+            if (!hasOwn(message, field.name) || isImplicitDefault(field, value))
                 continue;
             setOwn(out, key, writeSingular(field, value, options, depth));
         }
--- a/ext/protojson.js
+++ b/ext/protojson.js
@@ -1,0 +1,903 @@
+// Copyright 2021 Google LLC
+// Copyright 2026 The protobuf.js Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Derived from proto3-json-serializer v3.0.4 and modified by The protobuf.js Authors.
+
+"use strict";
+var protobuf = require("../light");
+
+/* global BigInt */
+
+var Type = protobuf.Type,
+    Enum = protobuf.Enum,
+    util = protobuf.util;
+
+var protojson = protobuf.protojson = module.exports = {};
+
+function isExtension(field) {
+    return Boolean(field.declaringField) || field.name.charAt(0) === ".";
+}
+
+function extensionName(field) {
+    var df = field.declaringField || field,
+        full = df.fullName.charAt(0) === "." ? df.fullName.slice(1) : df.fullName;
+    return "[" + (df.protoName ? full.replace(/[^.]+$/, df.protoName) : full) + "]";
+}
+
+function indexField(index, key, field, type) {
+    var existing = index[key];
+    if (existing !== undefined && existing !== field)
+        throw Error(type.fullName + ": duplicate ProtoJSON field name " + JSON.stringify(key));
+    index[key] = field;
+}
+
+function fieldsByJsonName(type) {
+    if (type._fieldsByJsonName)
+        return type._fieldsByJsonName;
+    var index = Object.create(null),
+        fields = type.fieldsArray,
+        i = 0;
+    for (; i < fields.length; ++i) {
+        var field = fields[i].resolve();
+        if (isExtension(field))
+            indexField(index, extensionName(field), field, type);
+        else {
+            indexField(index, field.name, field, type);
+            indexField(index, field.jsonName, field, type);
+            indexField(index, field.protoName, field, type);
+        }
+    }
+    type._fieldsByJsonName = index;
+    return index;
+}
+
+// --- reading scalars ---
+
+var INT_RANGE = {
+    int32:    ["-2147483648", "2147483647"],
+    sint32:   ["-2147483648", "2147483647"],
+    sfixed32: ["-2147483648", "2147483647"],
+    uint32:   ["0", "4294967295"],
+    fixed32:  ["0", "4294967295"],
+    int64:    ["-9223372036854775808", "9223372036854775807"],
+    sint64:   ["-9223372036854775808", "9223372036854775807"],
+    sfixed64: ["-9223372036854775808", "9223372036854775807"],
+    uint64:   ["0", "18446744073709551615"],
+    fixed64:  ["0", "18446744073709551615"]
+};
+
+var LONG_TYPE = { int64: 1, uint64: 1, sint64: 1, fixed64: 1, sfixed64: 1 };
+
+var MAX_FLOAT = 3.4028234663852886e38;
+var hasBigInt = typeof BigInt !== "undefined";
+
+var NUMERIC_RE = /^[+-]?(?:[0-9]*\.[0-9]+|[0-9]+\.?)(?:[eE][+-]?[0-9]+)?$/;
+
+var SKIP = {};
+
+function invalid(name, value, what) {
+    return Error(name + ": " + what + ": " + JSON.stringify(value));
+}
+
+function parseIntegerString(value, type, name) {
+    var str;
+    if (typeof value === "number") {
+        if (!isFinite(value) || Math.floor(value) !== value)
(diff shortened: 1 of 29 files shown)


ext/protojson.js
```diff
@@ -0,0 +1,903 @@
...
+    var result = (seconds < 0 || nanos < 0 ? "-" : "") + Math.abs(seconds);
+    if (nanos)
+        result += "." + nanosToString(nanos);
+    return result + "s";
+};
+
+WKT_FROM[".google.protobuf.Timestamp"] = function (type, value) {
+    if (typeof value !== "string")
+        throw invalid(type.fullName, value, "expected timestamp string");
+    var match = /^(\d{4})-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01])T([01]\d|2[0-3]):([0-5]\d):([0-5]\d)(?:\.(\d{1,9}))?(Z|[+-](?:[01]\d|2[0-3]):[0-5]\d)$/.exec(value);
```
alexander-fenster, line 538: Not sure what the `:[0-5]\d` means at the end of the regex; I think it blocks valid timestamps like `'2026-06-03T09:19:00.000-0700'` from being parsed properly (returning "invalid timestamp").

alexander-fenster:
That's great!

One thing I noticed in my tests that if an protobuf object is created with `.create()`, the JSON serializer will include default values. E.g. for the proto message

```proto
message Test {
  int32 int_value = 1;
}
```

the code

```js
const Test = root.lookup('Test');
console.log(protojson.toJson(Test, Test.create({intValue: 0})));
```

will emit `{ intValue: 0 }` as opposed to an empty object as expected.

Otherwise, passes my tests :)

alexander-fenster (approved):

ext/protojson.js
dcodeIO: Here `:[0-5]\d` is the minutes portion of a timezone offset, like in `-07:59`, whereas `-0700` is not valid RFC 3339 iiuc, and hence not ProtoJSON that protoc accepts. Is asserted by the conformance harness in `TimestampWithMissingColonInOffset`.

PR #2302, opened 2026-06-04
fix: Avoid name collisions in generated code

### LINKED ISSUE #1058: Unable to create static module for proto containing message called 'Object'
protobuf.js version: 6.8.6

Steps to reproduce:
- Create ```object.proto```
```proto
syntax = "proto3";
message Object {}
```
- Create `bundle.js` as a static-module:
```
npx pbjs -t static-module object.proto -o bundle.js
```
- Create `main.js`:
```js
require('./bundle.js').Object.create({});
```
- Run `main.js`
```
node main.js
```
Witness error
```
/path/to/project/bundle.js:37
                for (var keys = Object.keys(properties), i = 0; i < keys.length; ++i)
                                       ^

TypeError: Object.keys is not a function
    at new Object (/path/to/project/bundle.js:37:40)
    at Function.create (/path/to/project/bundle.js:51:20)
    at Object.<anonymous> (/path/to/project/main.js:1:93)
    at Module._compile (internal/modules/cjs/loader.js:678:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:689:10)
    at Module.load (internal/modules/cjs/loader.js:589:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:528:12)
    at Function.Module._load (internal/modules/cjs/loader.js:520:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:719:10)
    at startup (internal/bootstrap/node.js:228:19)
```

Expected result is no error.

---

Clearly, the message `Object` causes a name conflict with the javascript `Object`.

It also fails if I create `bundle.js` as a json-module (`npx pbjs -t json-module object.proto -o bundle.js`) 

In the static-module case, it can be temporarily fixed in `bundle.js` by assigning javascript's `Object` to a variable `JSObject`, say, and using `JSObject.keys(properties)`.   However, this requires such a change each time the code is generated.

A better work-around is to use reflection.  For example, there is no error if `main.js` is:
```javascript
let protobuf = require('protobufjs');
let root = protobuf.loadSync('object.proto')
root.Object.create()
```

In my particular case, I don't have control of the proto file to change the message name.  

Using reflection is ok, but it's a shame.  Particularly because I'm using Typescript, for which I import the types from a generated `bundle.d.ts`, but then I need a `bundle.js` around at runtime (could just be empty).  It's a shame to have this unused file knocking around... I'd like it to be useful containing a static module.

I raise it as an issue here because:
  * The same proto works fine in other languages.
  * The inconsistency between using reflection vs. static code.

Thanks for any consideration.

### LINKED ISSUE #1113: Error: duplicate name 'constructor' in Type
protobuf.js version: 6.8.8

I'm implementing a Node.js client to a service that communicates via protobuf messages. The message contents hold code-related information. For instance, there is a protobuf model similar to the following that represents the context of a function call:

```
// models.proto
package my.models;

message FunctionCallContext {
    string method_name = 1;
    repeated string arguments = 2;
    bool constructor = 3;
    // etc
}
```
It appears that the `constructor` field is causing conflict when I load the models from its `.proto` file. From a shallow investigation, it looks like the validation is failing simply because the library provides each Message with an identically-named property: https://github.com/dcodeIO/protobuf.js/blob/master/src/type.js#L162 

So does this mean the field name is off limits should I choose to use this library? Are there others? Should this be remedied or just documented? Maybe this is an edge case scenario but it seems unfortunate. Also, I don't manage the models, so a "fix" involving an edit to the `.proto` isn't available to me.

**Reproduction:**
With the model example above, either
```js
require('protobufjs').load(models.proto).then(/* */);
```
or
```sh
pbjs -t json models.proto > bundle.json
```
will result in
```
Error: duplicate name 'constructor' in Type FunctionCallContext
    at Type.add (/Users/matthewhenry/workspace/cs/contrast-node-protect/core/node_modules/protobufjs/src/type.js:331:15)
    at parseField (/Users/matthewhenry/workspace/cs/contrast-node-protect/core/node_modules/protobufjs/src/parse.js:378:16)
    at parseType_block (/Users/matthewhenry/workspace/cs/contrast-node-protect/core/node_modules/protobufjs/src/parse.js:338:21)
    at ifBlock (/Users/matthewhenry/workspace/cs/contrast-node-protect/core/node_modules/protobufjs/src/parse.js:286:17)
    at parseType (/Users/matthewhenry/workspace/cs/contrast-node-protect/core/node_modules/protobufjs/src/parse.js:304:9)
    at parseCommon (/Users/matthewhenry/workspace/cs/contrast-node-protect/core/node_modules/protobufjs/src/parse.js:259:17)
    at parse (/Users/matthewhenry/workspace/cs/contrast-node-protect/core/node_modules/protobufjs/src/parse.js:728:21)
    at process (/Users/matthewhenry/workspace/cs/contrast-node-protect/core/node_modules/protobufjs/src/root.js:107:30)
    at fetch (/Users/matthewhenry/workspace/cs/contrast-node-protect/core/node_modules/protobufjs/src/root.js:166:13)
    at Root.load (/Users/matthewhenry/workspace/cs/contrast-node-protect/core/node_modules/protobufjs/src/root.js:194:13)
```
Commenting out the `constructor` field of the model will allow the commands to execute without error.

dcodeIO:
Updates generated code to avoid collisions between schema-derived names and common JavaScript names.

* Generates anonymous functions so function names cannot collide with scoped names
* Aliases global references like `Object` as `$Object = $util.global.Object` when used

Also removes and updates stale CLI dependencies.

Fixes #1058
Fixes #1113

Supersedes #1064


cli/targets/static.js
```diff
@@ -249,9 +249,133 @@ var renameVars = {
...
+    "globalThis",
+    "isFinite",
+    "isNaN",
+    "parseFloat",
+    "parseInt",
+    "queueMicrotask",
+    "setInterval",
+    "setTimeout",
+    "structuredClone",
+    "undefined",
```
alexander-fenster, line 339: `"null"`?

alexander-fenster (approved):

cli/targets/static.js
dcodeIO: One would think so, but `null`, `true`, `false` etc. are keywords one can't shadow, whereas `undefined` is not :)

Issue #2278, opened 2026-05-21
inquire() triggers webpack warning


ElmarFrerichs:
TL;DR: Please remove the imports of the deprecated and unused inquire() function as it thows warnings in webpack

Hello,

I dug through quite a few issues and code snippets now, I hope I didn't miss something.

Manifestation:
* The build pipeline of one of my projects failed
* That build pipeline was set to "CI=true", so it fails on warnings
* webpack threw the following warning: `WARNING in ./node_modules/@protobufjs/inquire/index.js 16:14-33\n
Critical dependency: the request of a dependency is an expression`. For the curious: https://github.com/webpack/webpack/blob/294197c06f45809751cde07e1860711138b3f41d/lib/dependencies/ContextDependencyHelpers.js#L275
* This is https://github.com/protobufjs/protobuf.js/blob/d28ec57e9eae7e1a67985ba76676c70534c42cab/src/util/inquire.js#L16 triggering this message
* As always in the JS universe, this is multiple levels deep in dependencies
* This now sets us up for the choice of ignoring warnings in our pipeline or living with constant "your build failed" messages, which essentially boils down to the first choice

This line was last changed in #1941 to replace a line with `eval()` - rightfully so, thank you for that :-) .

But why the warning from webpack? Because that module name might contain user input, in which case you offer an attacker the possibility to load modules of their choice, maybe even from other user accesible storage...

The `inquire` function already has a `@deprecated` mark on it since at least 5 years, and as far as I can tell, it's not used in protobuf. It is still imported though:
* for the tests: https://github.com/protobufjs/protobuf.js/blob/d28ec57e9eae7e1a67985ba76676c70534c42cab/tests/util_inquire.js#L3
* as part of utils/minimal.js, as a `require()`, thus propagating to the main index.d.ts file via transitive imports: https://github.com/protobufjs/protobuf.js/blob/d28ec57e9eae7e1a67985ba76676c70534c42cab/src/util/minimal.js#L17
* as part of the package-lock.json in the cli as `@protobufjs/inquire` still (and with an old version at that): https://github.com/protobufjs/protobuf.js/blob/d28ec57e9eae7e1a67985ba76676c70534c42cab/cli/package-lock.json#L375

[...]

PR #2305, opened 2026-06-05
fix: Remove inquire submodule

dcodeIO:
Removes the obsolete inquire submodule because keeping it around is still causing issues with bundlers. This is technically a small breaking API change, but I think it is safe to assume that the `protobuf.util.inquire` API is not used given its inherent issues. If it is used, which I doubt, the straightforward fix would be to add `@protobufjs/inquire` as an explicit dependency.

See https://github.com/protobufjs/protobuf.js/issues/2278#issuecomment-4628603975


alexander-fenster (approved):

deiga:
You'll be "happy" to hear, that on 7.6.3 we are still seeing the same issue 😬 
```
WARNING in ../../node_modules/@protobufjs/inquire/index.js 16:15-33
  ⚠ Critical dependency: the request of a dependency is an expression
    ╭─[16:22]
 14 │       return null;
 15 │     }
 16 │     var mod = require(moduleName);
    ·                       ──────────
 17 │     if (mod && (mod.length || Object.keys(mod).length)) return mod;
 18 │     return null;
    ╰────
```

These are all places I could find where `@protobufjs/inquire` are mentioned:
- `node_modules/protobufjs/src/util/minimal.js:util.inquire = require("@protobufjs/inquire");`
- `node_modules/protobufjs/package.json:    "@protobufjs/inquire": "^1.1.2",`
- `node_modules/protobufjs/dist/light/protobuf.js.map:...`
- `node_modules/protobufjs/dist/light/protobuf.min.js.map:...`
- `node_modules/protobufjs/dist/protobuf.js.map:...`
- `node_modules/protobufjs/dist/minimal/protobuf.js.map:...`
- `node_modules/protobufjs/dist/minimal/protobuf.min.js.map:...`
- `node_modules/protobufjs/dist/protobuf.min.js.map:...`
- `node_modules/@protobufjs/inquire/package.json:  "name": "@protobufjs/inquire",`
- `node_modules/@protobufjs/inquire/README.md:...`

dcodeIO:
That's in so far expected that the inquire submodule has been pulled from main/8.x in v8.6.1, but not from 7.x. Can possibly be backported, but I'm also curious what is blocking an upgrade to 8.x in your case?

deiga:
> That's in so far expected that the inquire submodule has been pulled from main/8.x in v8.6.1, but not from 7.x. Can possibly be backported, but I'm also curious what is blocking an upgrade to 8.x in your case?

It's not a direct dependency in our project, so upgrading means forcing resolution or nudging our deps to upgrade. Neither of which is particularly easy 😬

dcodeIO:
Now backported to 7.x, with the inquire submodule removed there as well.

deiga:
Lovely, thanks! 😍

Will test out ASAP!

PR #2306, opened 2026-06-05
fix(cli): Consistently wait for pbts output before JSDoc exit

dcodeIO:
Makes sure that JSDoc waits for generated `pbts` output to flush before exiting. Turns out that otherwise, piped output can produce incomplete declarations in some sandboxed environments.

### DIFF
--- a/cli/lib/tsd-jsdoc/publish.js
+++ b/cli/lib/tsd-jsdoc/publish.js
@@ -2,7 +2,7 @@
 
 var fs = require("fs");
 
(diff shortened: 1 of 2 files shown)


dcodeIO:
Seems there's still something off in that particular environment. Need to look more closely.

dcodeIO:
Strange stuff these sandboxes inside sandboxes :)

alexander-fenster (approved):

PR #2308, opened 2026-06-06
fix: Preserve descriptor metadata needed by protoc-gen-pbjs

dcodeIO:
Preserves file-level edition features, proto and JSON field names, and group metadata when importing descriptors by accepting a descriptor context object in place of the prior syntax/edition string.

With this change, `protoc-gen-pbjs` has enough metadata to match `pbjs`. in turn passing conformance.


alexander-fenster (approved):

jtbandes:
I believe this was actually a breaking change, since `keepCase: true` used to be effectively the default. I had some code using 

```ts
const descriptorSet = descriptor.FileDescriptorSet.decode(data);
const root = protobufjs.Root.fromDescriptor(descriptorSet);
root.resolveAll();
const rootType = root.lookupType(name);
rootType.toObject(rootType.decode(...), ...);
```

which now has to change to:
```ts
...
const root = protobufjs.Root.fromDescriptor(descriptorSet, { keepCase: true });
...
```
to maintain the existing behavior.

We would appreciate if changes like this would be called out in the release notes, and preferably avoid putting these in "patch" releases for semantic versioning. 🙏

jtbandes:
Another point – the comment on [`json_name`](https://github.com/protocolbuffers/protobuf/blob/f8e30e545e10cc9c3dc01c5535c5b746d2878bf6/src/google/protobuf/descriptor.proto#L301-L305) indicates that if absent, it should be deduced by converting the field name to camelCase. However I've observed in practice that this new behavior of using the json_name seems to only happen when the json_name was actually set in the binary FDS. I had some other data lying around with no json_name set, and that data behaved as expected (producing snake_case field names when using toObject) even without `keepCase: true`, which seems contrary to the intention of the descriptor.proto comments.

PR #2313, opened 2026-06-09
fix: support hardened prototype properties

turadg:
## Summary

- replace prototype `constructor` and `toString` shadowing assignments with descriptor-based definitions
- update static service generation to emit descriptor-based constructor setup
- add `globalThis` as a final portable global fallback while preserving existing `global` / `window` / `self` behavior
- add a dependency-free regression test for JavaScript's override mistake
- move source guards for unsafe assignment patterns to ESLint

## Context

Direct assignment to an own property can fail in strict mode when the same property is inherited as non-writable. This can happen in hardened JavaScript environments and affects patterns such as assigning `prototype.constructor` or `toString`.

## Validation

- `npm.cmd run lint:sources`
- `npm.cmd run test:sources`
- `npm.cmd run build:bundle`
- GitHub Actions `Test` workflow, including Node 12

### DIFF
--- a/cli/targets/static.js
+++ b/cli/targets/static.js
@@ -1010,7 +1010,7 @@ function buildService(ref, service) {
     --indent;
     push("};");
     push("");
-    push("(" + escapeName(service.name) + ".prototype = " + globalRef("Object") + ".create($protobuf.rpc.Service.prototype)).constructor = " + escapeName(service.name) + ";");
+    push(globalRef("Object") + ".defineProperty(" + escapeName(service.name) + ".prototype = " + globalRef("Object") + ".create($protobuf.rpc.Service.prototype), \"constructor\", { value: " + escapeName(service.name) + ", writable: true, enumerable: false, configurable: true });");
 
     if (config.create) {
         push("");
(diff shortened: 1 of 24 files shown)


dcodeIO (commented):
Looking at the diff, I am hesitant to add `@endo/init` just for a regression test here. I'd prefer to test the behavior directly without pulling in another dependency graph.

I also wonder how far we should describe this as SES/HardenedJS compatibility while protobuf.js reflection still relies on `new Function(...)` for runtime code generation. That seems like an important detail, but I am not familiar with whether such `eval`-like patterns are already a blocker for HardenedJS.

That said, the `defineProperty` changes look otherwise fine to me if we can keep this scoped to a dependency-free test that freezes or defines the relevant inherited properties directly, and ideally delegate the source guards to ESLint.

turadg:
Thanks for the fast review!

The regression test now eschews Endo and models the TC39-described “override mistake” directly:
https://github.com/tc39/how-we-work/blob/main/terminology.md#override-mistake

Thanks for thinking comprehensively about SES and raising the `new Function(...)` behavior.  I checked with SES expert @mhofman and SES should not interfere with protobuf.js’s use of `new Function()`.

dcodeIO (approved):

dcodeIO:
Thanks!

turadg:
Good idea to set up the lint rules!

PR #2315, opened 2026-06-09
Fix importing from ESM by revising export patterns

### LINKED ISSUE #2314: no export named 'load'
### Expected behavior

#1949 was closed due to age and (I assume) due to [recent ESM improvements](https://github.com/protobufjs/protobuf.js/issues/2047#issuecomment-4660466462), but I still get errors in an ESM project when I try to do `import { load } from 'protobufjs'`.

### Actual behavior

```
import { load } from 'protobufjs';
         ^^^^
SyntaxError: Named export 'load' not found. The requested module 'protobufjs' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'protobufjs';
const { load } = pkg;

    at #asyncInstantiate (node:internal/modules/esm/module_job:327:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:431:5)
    at async node:internal/modules/esm/loader:633:26
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)
```

### Minimal reproduction

A one-line file, `test.mjs`:

```js
import { load } from 'protobufjs';
```

### protobufjs version

8.6.1

### protobufjs-cli version, if applicable

_No response_

### Environment, if applicable

Node.js 24

joshkel:
Node.js determines valid identifiers to import from CJS to ESM by statically analyzing CJS code. For performance, it only supports a subset of code. This is documented at

https://github.com/nodejs/cjs-module-lexer/blob/main/README.md

Assigning to `module.exports` instead of `protobuf = module.exports` is the simplest way I found to make the static analyzer happy.

Fixes #2314


dcodeIO:
Generally looks good to me, just wondering if we can shorten the `module.exports` accesses to `exports` despite the `module.exports` reassignment. Apart from that, we need to make sure that not assigning to `protobuf` doesn't somehow confuse JSDoc, which produces our typings. Can be checked by rerunning `npm run build:types`.

joshkel:
The generated types appear to be identical, so no problems there. Thanks for pointing that out.

Using `exports` has limitations. Building off of [this Stack Overflow explanation](https://stackoverflow.com/a/26451885/25507) (which says it's as if each CJS module starts with `var module = new Module(...); var exports = module.exports;`), and based on local testing:

```js
var protobuf = exports = require('./index-minimal');
```

This approach of reassigning `exports` doesn't work, because it doesn't touch the `module.exports` field that actually controls the behavior.

```js
var protobuf = module.exports = require('./index-minimal')
exports.encoder = require("./encoder");
```

This approach of reassigning `module.exports` then assigning within `exports` doesn't work, because `exports` refers to the old, now inactive empty `module.exports` object, not the reassigned one.

```js
var protobuf = module.exports = exports = require('./index-minimal');
exports.encoder = require("./encoder");
```

This **does** work. I personally find the extra assignments strange, but I'm happy to go that route if you prefer.

dcodeIO:
Perhaps `exports = module.exports = require('./index-minimal');` is enough, where `exports` takes the place of `protobuf`?

dcodeIO (approved):

dcodeIO:
Made the suggested change and verified locally that named imports now work. Thanks!

joshkel:
Thanks, @dcodeIO!

PR #2324, opened 2026-06-10
fix: Remove renamed reflection objects by identity

### LINKED ISSUE #2016: Calling `ns.remove(node)` on renamed nodes does not work.
protobuf.js version: 7.3.2 and newer.

Steps to reproduce:

- Find a `Type` object in the tree: `const myType = ns.lookupType('.MyType')`
- Change its name: `myType.name = "NewName"`
- Remove the type from the parent: `myType.parent.remove(myType)`

Expected:

- The type is removed from the parent, and no longer in the `nestedArray` property

Actual:

- The type is still a child of the parent

I've added a test in a fork: https://github.com/runeh/protobuf.js/commit/a04e6ea56228100f17dc6636026c604ef980c4a3

I assume it's due to this line: https://github.com/protobufjs/protobuf.js/blob/59569c12c85c1c7b783ace9a71775b1d05a08e9c/src/namespace.js#L261

The `remove` code assumes that the name will not have changed between the node being added and removed.

dcodeIO:
Adds a helper to remove reflection objects by identity when their current `name` no longer matches the key used when they were added. Applies to namespace children, fields, and oneofs.

Fixes #2016

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -59,6 +59,28 @@ util.toObject = function toObject(array) {
 };
 
 /**
+ * Removes a value from an object.
+ * @param {Object.<string,*>|undefined} object Object to remove from
+ * @param {*} value Value to remove
+ * @param {string} [key] Optional key for fast path removal
+ * @returns {boolean} `true` if removed, otherwise `false`
+ */
+util.remove = function remove(object, value, key) {
+    if (!object)
+        return false;
+    if (key !== undefined && Object.prototype.hasOwnProperty.call(object, key) && object[key] === value) {
+        delete object[key];
+        return true;
+    }
+    for (var names = Object.keys(object), i = 0; i < names.length; ++i)
+        if (object[names[i]] === value) {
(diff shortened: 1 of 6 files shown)


src/util.js
```diff
@@ -58,6 +58,28 @@ util.toObject = function toObject(array) {
...
+    if (!object)
+        return false;
+    if (key !== undefined && Object.prototype.hasOwnProperty.call(object, key) && object[key] === value) {
+        delete object[key];
+        return true;
+    }
+    for (var names = Object.keys(object), i = 0; i < names.length; ++i)
+        if (object[names[i]] === value) {
+            delete object[names[i]];
+            return true;
```
alexander-fenster, line 78: I guess this assumes that there's only one key with this value

alexander-fenster (approved):

src/util.js
dcodeIO: Right, updated the doc comment accordingly

PR #2337, opened 2026-06-23
fix: Implement open/closed enum semantics

dcodeIO:
Implements OPEN and CLOSED enum semantics across codegen and the ProtoJSON and Text Format extensions, covering the relevant singular, repeated unpacked, repeated packed, and map-entry cases.

Background: Open enum fields accept any signed int32 value, while closed enum fields reject or drop unknown values. During decode, unknown closed enum values are preserved in `$unknowns` when unknown-field preservation is enabled.


alexander-fenster (approved):
`

PR #2345, opened 2026-06-23
fix: omit proto3 implicit-presence defaults when encoding from parse().root

spokodev:
A proto3 singular scalar set to its **default** value is incorrectly serialized onto the wire when the type is built via `parse(proto).root` or `new Root().addJSON(json)` and encoded without first calling `resolveAll()`.

```js
const M = protobuf.parse('syntax="proto3"; message M { int32 i32=1; string s=2; bool b=3; }')
  .root.lookupType("M");

M.encode({ i32: 0 }).finish()    // [8, 0]   — spec requires []
M.encode({ s: "" }).finish()     // [18, 0]  — spec requires []
M.encode({ b: false }).finish()  // [24, 0]  — spec requires []
M.encode({ i32: 5 }).finish()    // [8, 5]   — correct
```

Calling `root.resolveAll()` first (or using `load()`/`loadSync()`/`fromJSON()`, which resolve internally) produces the correct `[]`, so only the lazy `parse().root` encode path is affected.

### Cause
`Type#setup()` — the lazy initializer that builds the encoder on first `encode()` — resolves field **types** (`field.resolve()`) but never resolves **feature defaults**. So at codegen time `field._features.field_presence` is still `undefined`, and `Field#hasPresence` returns `undefined !== "IMPLICIT"` → `true`. The generated encoder then takes the presence-tracking branch (no default-value check) for an implicit-presence field and emits its default.

### Impact
Non-canonical proto3 output from the most common inline-proto path: larger payloads, and byte-mismatches against peers that omit defaults (Go/C++/Python, and protobuf.js's own `load()`), which breaks deterministic serialization, signing/hashing of serialized bytes, and golden-file/conformance comparisons. Decoding is unaffected.

### Fix
Resolve feature defaults in `setup()` before the codecs are generated — the same resolution `load()`/`loadSync()`/`fromJSON()` already perform — so `parse().root` produces the same wire form. Verified across syntaxes: proto3 implicit fields now omit defaults (`[]`); proto2, proto3 `optional`, and editions explicit-presence fields still emit them (`[8,0]`); `-0.0` is still preserved.

### Tests
Added a `comp_optional` case mirroring the existing "implicit scalar conversion defaults" test but **without** the `resolveAll()` call. It fails on the current code (`[8,0]`…) and passes with the fix. Full `test:sources` green (2723 passing, 0 failing); no regressions.

### DIFF
--- a/src/type.js
+++ b/src/type.js
@@ -494,6 +494,13 @@ Type.prototype.setup = function setup() {
     // Sets up everything at once so that the prototype chain does not have to be re-evaluated
     // multiple times (V8, soft-deopt prototype-check).
 
+    // Resolve feature defaults (e.g. proto3 implicit field presence) before the
+    // codecs are generated, so encoders built from a parsed-but-not-resolveAll'd
+    // root omit implicit-presence default values per the proto3/editions spec.
+    var root = this.root;
+    if (root && root._needsRecursiveFeatureResolution && (root._edition || this._edition))
+        root._resolveFeaturesRecursive(root._edition);
+
     var fullName = this.fullName,
         types    = [];
     for (var i = 0; i < /* initializes */ this.fieldsArray.length; ++i)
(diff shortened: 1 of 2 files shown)


dcodeIO (approved):

dcodeIO:
Thanks!

PR #2352, opened 2026-07-02
fix: handle EOF during options parsing

alexander-fenster:
Incomplete option in the proto file might cause an infinite loop when parsing. Follow up for #1923.

### DIFF
--- a/tests/comp_options-parse.js
+++ b/tests/comp_options-parse.js
@@ -1,6 +1,5 @@
 var tape = require("tape");
 var protobuf = require("..");
-var childProcess = require("child_process");
 
 tape.test("Options", function (test) {
     var root = protobuf.loadSync("tests/data/options_test.proto");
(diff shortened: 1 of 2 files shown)


dcodeIO (approved):

tests/comp_options-parse.js
```diff
@@ -164,6 +165,25 @@ tape.test("Options", function (test) {
         test.end();
     });
 
+    test.test(test.name + " - unterminated option name", function (test) {
+        var result = childProcess.spawnSync(process.execPath, ["-e",
```
dcodeIO: Seems this can be tested without a child process by calling `protobuf.parse` with this source directly. Would make the test a bit more compact.

alexander-fenster: Sure, that makes it much simpler.

Issue #2020, opened 2024-08-21
utf8 read function frequent string concatenation generate a large number of ConsString objects


LukeChenyk:
protobuf.js version: 7.2.6

[,](https://github.com/protobufjs/protobuf.js/commit/75172cd11be137bbabd2fba7a02b15067695ebad) this commit has some problem.
On v8, frequent string concatenation can generate a large number of ConsString objects, leading to increased memory usage.

```js

// lib/utf8/index.js
utf8.read = function utf8_read(buffer, start, end) {
    if (end - start < 1) {
        return "";
    }

    var str = "";
    for (var i = start; i < end;) {
        var t = buffer[i++];
        if (t <= 0x7F) {
            str += String.fromCharCode(t);
        } else if (t >= 0xC0 && t < 0xE0) {
            str += String.fromCharCode((t & 0x1F) << 6 | buffer[i++] & 0x3F);
        } else if (t >= 0xE0 && t < 0xF0) {
            str += String.fromCharCode((t & 0xF) << 12 | (buffer[i++] & 0x3F) << 6 | buffer[i++] & 0x3F);
        } else if (t >= 0xF0) {
            var t2 = ((t & 7) << 18 | (buffer[i++] & 0x3F) << 12 | (buffer[i++] & 0x3F) << 6 | buffer[i++] & 0x3F) - 0x10000;
            str += String.fromCharCode(0xD800 + (t2 >> 10));
            str += String.fromCharCode(0xDC00 + (t2 & 0x3FF));
        }
    }

    return str;
};
```

chrome inspect:
![image](https://github.com/user-attachments/assets/fede059b-47a8-4b88-b892-5b0966f617da)

dcodeIO:
Addressed by https://github.com/protobufjs/protobuf.js/pull/2355, in [@protobufjs/utf8 1.1.2](https://github.com/protobufjs/protobuf.js/releases/tag/utf8-v1.1.2)

PR #2353, opened 2026-07-02
perf: batch String.fromCharCode in utf8_read_js fallback

dsolinger:
## Summary

`utf8.read` has a fast ASCII path, but as soon as a string contains **any** non-ASCII byte the whole payload is handed to `utf8_read_js`, which builds the result with per-character `str += String.fromCharCode(...)`. For a large non-ASCII field this makes V8 construct a deep tree of [`ConsString`](https://gist.github.com/mraleph/3397008) rope nodes, so decode time becomes super-linear and it produces a large amount of short-lived garbage (heavy GC pressure).

This PR changes `utf8_read_js` to accumulate code units into a chunk array and flush them via `String.fromCharCode.apply` in 8192-unit batches — the same strategy the ASCII fast path and older releases already use. Behaviour is unchanged, including the CVE-2026-44288 overlong / out-of-range replacement-character handling.

## Benchmark

2 MB all-2-byte string (U+03BB `λ`), 20 iterations, Node 26:

| | ms/read |
|---|---|
| before (per-char) | 79.68 |
| after (batched) | 7.18 |

**~11× faster**, and the allocation/GC profile drops from O(n) rope nodes to ~ceil(n/8192) intermediate strings.

## Behaviour / correctness

No observable behaviour change. Added two tests to `tests/util_utf8.js`:
- a large multibyte string (mix of 2-byte, 3-byte and surrogate-pair code points) that spans many 8192-unit flush boundaries
- an ASCII-prefixed string, exercising the handoff from the ASCII fast path into `utf8_read_js` with a non-empty `str` prefix

All existing utf8 tests (empty buffer, long buffer, chunk-boundary, surrogate-pair-over-chunk, overlong-sequence replacement) continue to pass.

## Context

The ASCII fast path already avoids this for pure-ASCII input, but real-world payloads with any non-ASCII content (names, paths, i18n text, symbol tables) still hit the slow fallback. This is the same per-character-concatenation issue previously reported in #2020 for the pre-fast-path implementation.

### DIFF
--- a/src/util/utf8.js
+++ b/src/util/utf8.js
@@ -40,28 +40,42 @@ utf8.length = function utf8_length(string) {
 };
 
 function utf8_read_js(buffer, start, end, str) {
-    for (var i = start; i < end;) {
-        var t = buffer[i++];
+    // Batch code units and flush via String.fromCharCode.apply in 8192-unit
+    // chunks to avoid the per-character ConsString buildup of `str += ...`.
+    var parts = null,
+        chunk = [],
+        i = 0; // chunk write index
+    for (var pos = start; pos < end;) {
+        var t = buffer[pos++];
         if (t <= 0x7F) {
-            str += String.fromCharCode(t);
+            chunk[i++] = t;
         } else if (t >= 0xC0 && t < 0xE0) {
-            var c2 = (t & 0x1F) << 6 | buffer[i++] & 0x3F;
-            str += c2 >= 0x80 ? String.fromCharCode(c2) : replacementChar;
+            var c2 = (t & 0x1F) << 6 | buffer[pos++] & 0x3F;
+            chunk[i++] = c2 >= 0x80 ? c2 : 0xFFFD;
         } else if (t >= 0xE0 && t < 0xF0) {
-            var c3 = (t & 0xF) << 12 | (buffer[i++] & 0x3F) << 6 | buffer[i++] & 0x3F;
-            str += c3 >= 0x800 ? String.fromCharCode(c3) : replacementChar;
+            var c3 = (t & 0xF) << 12 | (buffer[pos++] & 0x3F) << 6 | buffer[pos++] & 0x3F;
+            chunk[i++] = c3 >= 0x800 ? c3 : 0xFFFD;
         } else if (t >= 0xF0) {
-            var t2 = (t & 7) << 18 | (buffer[i++] & 0x3F) << 12 | (buffer[i++] & 0x3F) << 6 | buffer[i++] & 0x3F;
+            var t2 = (t & 7) << 18 | (buffer[pos++] & 0x3F) << 12 | (buffer[pos++] & 0x3F) << 6 | buffer[pos++] & 0x3F;
             if (t2 < 0x10000 || t2 > 0x10FFFF)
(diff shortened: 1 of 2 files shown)


dsolinger:
**Note on the standalone `@protobufjs/utf8` package**

This PR fixes the fallback in the inlined `src/util/utf8.js`, which reaches consumers on protobufjs 8.x (and any 7.x line that later inlines utf8).

It does **not** help consumers still on protobufjs ≤7.x, which depend on the standalone [`@protobufjs/utf8`](https://www.npmjs.com/package/@protobufjs/utf8) package. That package's latest release (`1.1.1`, published 2026-04-27 as the CVE-2026-44288 fix) shipped the per-character `str += String.fromCharCode(...)` implementation and carries the same super-linear/GC regression on large non-ASCII strings.

Since the `lib/utf8/` subpackage source was removed from this repo in #2146, there doesn't appear to be a source path to regenerate it. If a maintainer is willing to cut a `@protobufjs/utf8@1.1.2` with the same batched accumulation (validation unchanged), it would fix every protobufjs 7.x user without a major-version bump. Happy to prepare that if it's useful — just not sure where the canonical source for the standalone package lives now.

dsolinger:
**Small-string overhead check**

Since batching adds a bit of fixed machinery (array + `.apply` + final `.slice`), I checked whether it regresses small non-ASCII strings (which all take the `utf8_read_js` path). It doesn't — the difference is within measurement noise across the whole range.

Non-ASCII strings (U+03BB `λ`), Node 26, ns/read:

| chars | batched (this PR) | per-char (before) |
|------:|------------------:|------------------:|
| 1 | 32 | 30 |
| 4 | 40 | 41 |
| 16 | 72 | 70 |
| 64 | 248 | 257 |
| 256 | 1046 | 1071 |
| 1024 | 4012 | 4175 |
| 8192 | 31659 | 31477 |
| 65536 | 196229 | 200069 |

The per-character ConsString buildup only becomes expensive well past the 8192-unit flush threshold, so below that the two are equivalent, and above it the batched version pulls ahead (≈11× at 2 MB). Net: no small-string cost, large-string win.

dcodeIO:
Would be interesting to tesk against this upcoming branch: https://github.com/protobufjs/protobuf.js/tree/patch/encoder-arch. That might also be the right place to make related changes given the potential for conflict.

dsolinger:
Thanks. I tested against `patch/encoder-arch`, and that branch already resolves this much more completely than this PR does. Routing inputs ≥ `TEXT_DECODER_MIN_LENGTH` (64) to native `TextDecoder` means large strings never reach `utf8_read_js` at all, so the ConsString buildup is gone regardless of ASCII/non-ASCII.

Measured on the same 2 MB all-`λ` (non-ASCII) string, Node 26:

|                                  | ms/read |
|----------------------------------|---------|
| master (per-char fallback)       | ~80     |
| this PR (batched fallback)       | ~7      |
| patch/encoder-arch (TextDecoder) | ~1.7    |

On `patch/encoder-arch`, `utf8_read_js` only ever handles sub-64-byte inputs, where batching makes no measurable difference — so this PR's change is essentially moot there. I'm happy to close this PR in favor of `patch/encoder-arch`.

Issue #2354, opened 2026-07-02
@protobufjs/utf8@1.1.1 (7.x) still has the per-character decode regression — consider a 1.1.2 republish


dsolinger:
## Summary

The fix for CVE-2026-44288 (`@protobufjs/utf8@1.1.1`, and the corresponding rewrite of `src/util/utf8.js`) replaced the batched `String.fromCharCode.apply` accumulation with per-character `str += String.fromCharCode(...)`. On V8 this builds a deep ConsString rope for long strings, making decode super-linear with heavy GC pressure.

## Impact

- **protobufjs 8.4.0+** has an ASCII fast path, so pure-ASCII input is unaffected — but the non-ASCII fallback (`utf8_read_js`) still concatenates per-character. Addressed in #2353.
- **protobufjs ≤7.x** depends on the standalone [`@protobufjs/utf8`](https://www.npmjs.com/package/@protobufjs/utf8) package, whose latest release (`1.1.1`) is pure per-character concatenation with no fast path. Every 7.x consumer hits the regression on any reasonably large string.

## Benchmark

2 MB non-ASCII string (U+03BB `λ`), 20 iterations, Node 26:

| | ms/read |
|---|---|
| before (per-char) | 79.68 |
| after (batched) | 7.18 |

**~11× faster**, allocation drops from O(n) ConsString nodes to ~ceil(n/8192) intermediate strings. The overlong / out-of-range → U+FFFD validation is fully preserved.

## Request

The standalone `@protobufjs/utf8` package's source appears to have been removed from this repo in #2146, so there's no obvious PR path to regenerate it. Would a maintainer be willing to cut a `@protobufjs/utf8@1.1.2` that restores the batched accumulation (validation unchanged)? That would fix all protobufjs 7.x consumers without requiring a major-version bump to 8.x.

Happy to prepare the implementation if pointed at where the canonical source for the standalone package now lives.

Related: #2353, #2020, CVE-2026-44288 (GHSA-q6x5-8v7m-xcrf)

dcodeIO:
For context, the path for updating and publishing the standalone module is the protobufjs-v7.x branch, which still contains it. For 8.x, I left a comment on the PR.

PR #2355, opened 2026-07-02
perf: batch String.fromCharCode in @protobufjs/utf8 read (7.x)

dsolinger:
Follow-up to #2354, targeting the standalone `@protobufjs/utf8` package on `protobufjs-v7.x` (as suggested there).

## Problem

The standalone `@protobufjs/utf8` package — depended on by all protobufjs 7.x consumers — builds the decoded string with per-character `str += String.fromCharCode(...)`. For a large string this makes V8 construct a deep tree of [`ConsString`](https://gist.github.com/mraleph/3397008) rope nodes, so decode becomes super-linear and churns a large amount of short-lived garbage (heavy GC pressure). This is the same shape as #2020, and it regressed in `1.1.1` relative to the batched `1.1.0` implementation.

Unlike the 8.x line, the 7.x standalone package has no ASCII fast path, so **every** string over a few KB hits this — ASCII or not.

## Change

Accumulate code units into a chunk array and flush via `String.fromCharCode.apply` in 8192-unit batches — the same strategy `1.1.0` used. Validation is unchanged: CVE-2026-44288 overlong / out-of-range sequences still decode to U+FFFD.

The intermediate flush slices to the write index (`chunk.slice(0, i)`), so a surrogate pair straddling the 8192-unit boundary (which writes 2 units and can push the index 8191 → 8193) cannot leave a stale trailing unit behind. The existing `surrogate_pair_bug.txt` test covers exactly this case.

## Benchmark

2 MB all-`λ` (U+03BB) string, 20 iterations, Node 26:

| | ms/read |
|---|---|
| before (per-char) | ~80 |
| after (batched) | ~6.8 |

**~11× faster**; allocation drops from O(n) rope nodes to ~ceil(n/8192) intermediate strings.

## Tests

`node lib/utf8/tests/index.js` → 16/16 pass, including the existing surrogate-pair-over-chunk case and a new large-multibyte string spanning many flush boundaries.

Related: #2354, #2353, #2020, CVE-2026-44288 (GHSA-q6x5-8v7m-xcrf)

### DIFF
--- a/lib/utf8/index.js
+++ b/lib/utf8/index.js
@@ -39,34 +39,45 @@ utf8.length = function utf8_length(string) {
  * @returns {string} String read
  */
 utf8.read = function utf8_read(buffer, start, end) {
-    if (end - start < 1) {
+    if (end - start < 1)
         return "";
-    }
-
-    var str = "";
-    for (var i = start; i < end;) {
-        var t = buffer[i++];
+    // Batch code units and flush via String.fromCharCode.apply in 8192-unit
+    // chunks to avoid the per-character ConsString buildup of `str += ...`.
+    var parts = null,
+        chunk = [],
+        i = 0, // chunk write index
+        t, t2, c2, c3;
+    while (start < end) {
+        t = buffer[start++];
         if (t <= 0x7F) {
-            str += String.fromCharCode(t);
+            chunk[i++] = t;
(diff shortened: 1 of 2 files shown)


dcodeIO (approved):

dcodeIO:
Thanks!

PR #2358, opened 2026-07-04
fix: reject truncated declarations without TypeError

morgan-coded:
Truncated `.proto` input could hit parser null reads and throw a raw `TypeError` instead of the parser's normal `illegal ...` error. Several guards were written `if (!nameRe.test(token = next()))`, which a null end-of-input token slips past because `nameRe.test(null)` is `true` (`null` stringifies to `"null"`) — so the null reached a downstream deref and crashed.

This keeps those EOF cases on the same catchable parse-error path as the options guard from #2352. It covers the field paths (type, name, id) plus the remaining statements that had the same gap: `oneof`, `group`, map fields, top-level message/enum/service names, `extend` reference, and `package`. Valid input is unchanged.

I checked each truncated input against the old `TypeError` path, added parse tests for them, then ran `npm run lint:sources`, `npm run test:sources`, and `npm test`.

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -220,6 +220,9 @@ function parse(source, root, options) {
     }
 
     function parseId(token, acceptNegative, max) {
+        if (token === null) {
+            throw illegal(token, "end of input");
+        }
         switch (token) {
             case "max": case "MAX": case "Max":
                 return max || maxFieldId;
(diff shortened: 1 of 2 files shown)


dcodeIO (approved):

dcodeIO:
Thanks!

PR #2359, opened 2026-07-05
feat: Rework encoder architecture and add packed reader fast paths

dcodeIO:
Introduces a new encoder architecture that replaces the ops list approach with direct writes. Is ~30% faster.

Also makes the following adjacent changes:

- Adds new primitives to write and read packed arrays
- Uses a cached DataView for fixed-width types when reaching a threshold where the allocation is amortized
- Uses TextDecoder in more cases, starting at 64 bytes
- Reduces plain array support surface by instead converting upon reader creation
- Changes `encodeDelimited` to fork unconditionally, fixing an issue with forked empty writers
- Removes Node 4-era internals that are no longer needed/supported
- Restructures the benchmark into multiple cases, adding two, updates results, adds charts
- Regenerates affected fixtures


vini123:
虚幻之物对应着冥冥之路！：《羽》

alexander-fenster (approved):

PR #2363, opened 2026-07-06
fix: Preserve parsed json_name in field options

### LINKED ISSUE #2350: [Bug/Breaking Change] Keep original json_name value in field options since v8.6.0
### Expected behavior

In v8.5.x and earlier, field.options.json_name was always available after parsing, following standard Protobuf parsing behavior.
The new Field.jsonName shortcut is helpful, but json_name Field.options should remain in  for backward compatibility and spec consistency. json_name is a standard Protobuf field option, and removing it breaks metadata traversal, code generation, and custom tooling that relies on complete option data.

### Actual behavior

Protobufjs v8.6.0+ moves the json_name field option from Field.options to the standalone Field.jsonName property and deletes it from the options object. This unlisted breaking change breaks existing code that reads json_name directly from field options.

### Minimal reproduction

1. Define a proto field with explicit json_name:
syntax = "proto3";
message Test {
  string user_name = 1 [json_name = "userName"];
}
2. Parse with protobufjs ≥8.6.0
3. field.jsonName works correctly, but field.options.json_name is undefined

### protobufjs version

8.6.0

### protobufjs-cli version, if applicable

_No response_

### Environment, if applicable

Node 22.22.1

dcodeIO:
Preserves declared `[json_name = "..."]` field options while still lifting the value onto `Field#jsonName`.

Fixes #2350

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -867,15 +867,12 @@ function parse(source, root, options) {
         // lift json_name onto Field
         if (name === "json_name" && parent instanceof Field) {
             parent.jsonName = value;
-            return;
         }
         if (parent.setOption)
             parent.setOption(name, value);
     }
 
     function setParsedOption(parent, name, value, propName) {
-        if (name === "json_name" && parent instanceof Field)
-            return; // lifted onto Field#jsonName above
         if (parent.setParsedOption)
             parent.setParsedOption(name, value, propName);
     }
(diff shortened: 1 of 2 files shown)


alexander-fenster (approved):

deepak929200:
> Preserves declared `[json_name = "..."]` field options while still lifting the value onto `Field#jsonName`.
> 
> Fixes #2350

deepak929200:
> > Preserves declared `[json_name = "..."]` field options while still lifting the value onto `Field#jsonName`.
> > 
> > Fixes #2350
> 
>

PR #2364, opened 2026-07-06
fix: merge repeated message-valued fields within map entries

Ram-blip:
## Summary

Message-valued map entries currently decode each repeated `value` field occurrence into a new message. When a single synthetic map entry contains the message-valued `value` field more than once, fields from earlier occurrences can be discarded.

This change passes the previously decoded map value as the merge target, matching the existing behavior for ordinary singular embedded-message fields.

## Behavior

For a valid wire payload where one map entry contains two message-valued `value` fields:

```txt
value #1: { key: "merged" }
value #2: { values: ["x"] }
```

Before this change, the first partial message was overwritten:

```js
{ value: { a: { values: ["x"] } } }
```

After this change, the message values are merged:

```js
{ value: { a: { key: "merged", values: ["x"] } } }
```

The merge target is reset for each synthetic map entry, so separate map entries and duplicate map keys keep their existing behavior.

## Tests

Added regression coverage for both decode paths affected by this generator change:

- `tests/comp_maps.js` covers the runtime/reflection decoder.
- `tests/comp_static_maps.js` covers the generated static decoder.

Both tests construct a valid wire payload where a single synthetic map entry contains the message-valued `value` field more than once. The first occurrence contains one field, and the second occurrence contains another field. The expected behavior is that those embedded message values are merged, not overwritten.

Regenerated the affected static fixture with:

```sh
npm run build:tests
```

This updates `tests/data/test.js` so the generated static decoder passes the existing map value as the merge target.

### DIFF
--- a/src/decoder.js
+++ b/src/decoder.js
@@ -102,7 +102,7 @@ function decoder(mtype) {
                                 ("break");
 
             if (types.basic[type] === undefined) gen
-                            ("v=types[%i].decode(r,r.uint32(),undefined,q+1)", i); // can't be groups
+                            ("v=types[%i].decode(r,r.uint32(),undefined,q+1,v)", i); // can't be groups
             else gen
                             ("v=r.%s()", type === "string" ? stringMethod(field) : type);
 
--- a/tests/comp_maps.js
+++ b/tests/comp_maps.js
@@ -105,6 +105,27 @@ tape.test("maps", function(test) {
         test.end();
     });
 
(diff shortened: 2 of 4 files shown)


dcodeIO (approved):

dcodeIO:
Nice find, thanks!

Ram-blip:
Thanks for the quick review and merge! Glad I could help.

PR #2375, opened 2026-07-08
fix: base64-encode bytes field defaults in toObject

spokodev:
`Type#toObject` with `{ bytes: String }` is documented to return base64 strings, and the path for explicitly set values already does. The defaults path built the string with `String.fromCharCode`, which produces a raw latin1 string. A non-empty bytes default (proto2 or editions) came out as binary text that did not match a set value and could not be read back by `fromObject`. `tests/data/test.proto` documents the expected base64 for its `moo` default (`bW9v`).

This encodes the default with `util.base64.encode`, matching the set-value branch. Adds a converter test covering the default, the match with a set value, and the `fromObject` round trip. The empty-default case is unchanged.

### DIFF
--- a/src/converter.js
+++ b/src/converter.js
@@ -298,7 +298,7 @@ converter.toObject = function toObject(mtype) {
             else if (field.bytes) {
                 var arrayDefault = Array.prototype.slice.call(field.typeDefault);
                 gen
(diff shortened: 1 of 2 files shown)


dcodeIO (approved):

dcodeIO:
Thanks!

PR #2389, opened 2026-07-12
fix: preserve first enum alias from JSON descriptors

Ram-blip:
## Summary

JSON descriptor roots currently overwrite the reverse enum lookup for aliased enum values, causing ProtoJSON serialization to emit the last alias instead of the first alias.

ProtoJSON serializers should emit the first listed enum alias for a numeric value. Parsed `.proto` roots already preserve this behavior, but roots created through `Root.fromJSON(root.toJSON())` could emit a later alias because `Enum` construction overwrote `valuesById[id]`.

This change preserves the first reverse mapping while keeping all forward name-to-id mappings.

## Behavior

Given:

```proto
syntax = "proto3";

enum AliasChoice {
  option allow_alias = true;
  UNKNOWN = 0;
  OLD = 1;
  NEW = 1;
}

message AliasMessage {
  AliasChoice choice = 1;
}
```

Before this change:

```js
protojson.toJson(jsonRoot.lookupType("AliasMessage"), { choice: 1 });
```

could emit:

```js
{ choice: "NEW" }
```

After this change, JSON descriptor roots match parsed `.proto` roots and emit:

```js
{ choice: "OLD" }
```

Forward mappings remain unchanged, so both `OLD -> 1` and `NEW -> 1` are preserved.

## Tests

Added a focused ProtoJSON regression test in `tests/api_protojson.js`.

The test builds both:

- a parsed `.proto` root
- a JSON descriptor root using `Root.fromJSON(parsedRoot.toJSON())`

It verifies that both roots serialize the aliased enum value using the first listed alias.

## Commands run

```sh
npx tape tests/api_protojson.js
npm run lint:sources
npm test
```

### DIFF
--- a/src/enum.js
+++ b/src/enum.js
@@ -82,8 +82,11 @@ function Enum(name, values, options, comment, comments, valuesOptions) {
 
     if (values)
         for (var keys = Object.keys(values), i = 0; i < keys.length; ++i)
-            if (keys[i] !== "__proto__" && typeof values[keys[i]] === "number") // use forward entries only
-                this.valuesById[ this.values[keys[i]] = values[keys[i]] ] = keys[i];
+            if (keys[i] !== "__proto__" && typeof values[keys[i]] === "number") { // use forward entries only
+                this.values[keys[i]] = values[keys[i]];
+                if (this.valuesById[values[keys[i]]] === undefined)
+                    this.valuesById[values[keys[i]]] = keys[i];
+            }
 }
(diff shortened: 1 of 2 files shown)


alexander-fenster (approved):

alexander-fenster:
Just for the reference, here's the related section of the official docs: https://protobuf.dev/programming-guides/json/#serialization-behavior

PR #2393, opened 2026-07-17
fix: preserve non-finite defaults in toObject

Ram-blip:
## Summary

Preserve non-finite proto2 floating-point defaults when converting messages to plain objects.

When `defaults: true` is enabled:

* `Infinity` remains `Infinity`
* `-Infinity` remains `-Infinity`
* `NaN` remains `NaN`

When `json: true` is also enabled, these values are converted to their documented string representations:

* `"Infinity"`
* `"-Infinity"`
* `"NaN"`

## Root cause

The generated default-initialization path used codegen's `%j` formatter for scalar defaults.

Because `%j` uses `JSON.stringify`, valid non-finite floating-point defaults were embedded into the generated converter as `null`.

This made defaulted float and double fields behave differently from explicitly owned fields, whose non-finite values were already converted correctly.

## Fix

Handle actual numeric non-finite `float` and `double` defaults separately.

The converter now emits the numeric JavaScript value for ordinary object conversion and the corresponding quoted string when JSON-compatible conversion is requested.

The branch is guarded with `typeof field.typeDefault === "number"` so unexpected descriptor values cannot be coerced into generated numeric source.

## Tests

Added regression coverage for:

* parsing `inf`, `-inf`, and `nan` proto2 defaults;
* message prototype default access;
* plain-object conversion with `defaults: true`;
* JSON-compatible conversion with `defaults: true` and `json: true`;
* explicitly owned non-finite values;
* omission of unset fields when defaults are not requested.

### DIFF
--- a/cli/targets/static.js
+++ b/cli/targets/static.js
@@ -345,6 +345,18 @@ function globalRef(name) {
     return "$" + name;
 }
 
+function floatDefaultLiteral(value) {
+    if (Object.is(value, -0))
+        return "-0";
+    if (value === Infinity)
+        return globalRef("Infinity");
+    if (value === -Infinity)
+        return "-" + globalRef("Infinity");
+    if (Number.isNaN(value))
+        return globalRef("NaN");
+    return JSON.stringify(value);
+}
+
 function isIdentifierReference(node, parent) {
     if (!parent)
         return true;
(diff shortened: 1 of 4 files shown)


dcodeIO:
During review, I found an adjacent issue with negative zero that also goes through `%j` and is therefore emitted as `0`. The PRs linked above contain related parser and codegen fixes, hopefully enabling this PR to test `inf`, `-inf`, `nan` and `-0` all the way through from parsing to reflection and static codegen. Wdyt?

Ram-blip:
> During review, I found an adjacent issue with negative zero that also goes through `%j` and is therefore emitted as `0`. The PRs linked above contain related parser and codegen fixes, hopefully enabling this PR to test `inf`, `-inf`, `nan` and `-0` all the way through from parsing to reflection and static codegen. Wdyt?

Good catch, that makes sense. Thanks for preparing #2402 and #2403 for this. 

Once they land, I’ll rebase and extend coverage for `inf`, `-inf`, `nan`, and `-0` from parsing through reflection/default access, `toObject`, and static codegen.

For `json: true`, I’ll preserve `-0` as a finite numeric value while only stringifying the three non-finite values. I’ll also add static-target coverage for prototype defaults and generated `toObject`, and address any remaining `%j`/`JSON.stringify` default-emission paths that the new coverage exposes.

dcodeIO (approved):

dcodeIO:
Thanks!

PR #2401, opened 2026-07-26
feat(parse): enforce edition 2024 local visibility across files

### LINKED ISSUE #2400: Edition 2024 export/local visibility parses but is not enforced during cross-file resolution
src/parse.js still carries the TODO from #2060 — export/local visibility modifiers parse for edition >= 2024, but nothing enforces them at cross-file symbol resolution (TODO at src/parse.js:352 as of 04db6a5a). Would you want an outside PR for a first pass scoped to resolver metadata and focused imported-symbol fixtures, not broader Edition 2024 support?

morgan-coded:
Closes #2400.

Rebased onto master and reduced: #2415 landed the parsing and preservation half, so this is now only the resolution check, 21 lines in `src/field.js` plus tests.

After a successful lookup, `Field#resolve` rejects a resolved type or enum that is `local` to a different file:

```
'.a.Hidden' is local to 'a.proto' and cannot be referenced from 'c.proto'
```

If either side has no filename, which covers programmatic construction, `fromJSON`, and bundled common types, the reference cannot be proven to cross files and is allowed.

Tests sit in `tests/parse_editions.js` beside the existing export/local parsing cases: same-file references still resolve, cross-file `local` throws, `export` and unmodified symbols are unaffected, and pre-2024 editions plus proto2/proto3 behave exactly as before. That file goes 29 to 39 tests; the full suite 3201 to 3211, all green, with lint and type checks clean.

Two things left out on purpose, in case you would rather they were in:

- **Service RPC request and response types are not covered.** `Method#resolve` resolves them through `lookupType` and methods do carry a filename, so the same check would be live there. Sharing it across modules felt like a bigger structural decision than this change should make on its own; it is about 10 lines if you want it here.
- **`tryHandleExtension` in `root.js` is deliberately untouched.** It is a boolean try inside the deferred-retry loop, so throwing there would break deferred extensions and make errors depend on file load order. Extension fields are still checked through the sister field's own `resolve`.

The earlier note about `visibility` not surviving `toJSON` / `fromJSON` no longer applies, since #2415 carries it through the descriptor surface.

### DIFF
--- a/src/field.js
+++ b/src/field.js
@@ -334,6 +334,26 @@ Field.prototype.toJSON = function toJSON(toJSONOptions) {
 };
 
 /**
+ * Verifies that a resolved type or enum may be referenced by the given field.
+ * A symbol declared `local` (edition 2024) is only visible within its own file.
+ * If either file name is unknown - programmatic construction, `fromJSON`, bundled
+ * common types - the reference cannot be proven to cross files and is allowed.
+ * @param {Field} field Referencing field
+ * @param {Type|Enum} resolved Resolved type or enum
+ * @returns {undefined}
+ * @throws {Error} If the referenced symbol is local to another file
+ * @inner
+ */
+function checkVisibility(field, resolved) {
+    if (resolved.visibility !== "local")
+        return;
+    var from = (field.declaringField || field).filename;
+    if (!from || !resolved.filename || from === resolved.filename)
+        return;
+    throw Error("'" + resolved.fullName + "' is local to '" + resolved.filename + "' and cannot be referenced from '" + from + "'");
+}
+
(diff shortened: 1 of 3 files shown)


morgan-coded:
Rebased this onto master and it shrank a lot. #2415 landed the parsing and preservation since I opened this, so the parse.js half is gone and what's left is just the resolution check: 21 lines in `src/field.js` plus tests.

`Field#resolve` throws when a resolved type or enum is `local` to a different file, and stays quiet when either side has no filename (programmatic construction, `fromJSON`, bundled types). Suite goes 3201 to 3211, lint and type checks clean. Updated the description with the details.

PR #2418, opened 2026-08-24
fix: Harden ProtoJSON and Text Format parsing

dcodeIO:
Bounds integer literals before passing them to the `BigInt` constructor in the ProtoJSON and Text Format extensions, and replaces `push.apply` with a loop that reliably handles large Text Format string literals.

### DIFF
--- a/ext/protojson.js
+++ b/ext/protojson.js
@@ -91,6 +91,15 @@ function invalid(name, value, what) {
     return Error(name + ": " + what + ": " + JSON.stringify(value));
 }
 
+function normalizeIntegerString(str, name, what) {
+    str = str.replace(/^\+/, "").replace(/^(-?)0+(?=\d)/, "$1");
+    if (str === "-0")
+        str = "0";
+    if (str.length > 20) // max uint64 and min int64
+        throw Error(name + ": out of range for " + what);
+    return str;
+}
+
 function parseIntegerString(value, type, name) {
     var str;
     if (typeof value === "number") {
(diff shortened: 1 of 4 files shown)


ext/protojson.js
```diff
@@ -91,6 +91,15 @@ function invalid(name, value, what) {
     return Error(name + ": " + what + ": " + JSON.stringify(value));
 }
 
+function normalizeIntegerString(str, name, what) {
+    str = str.replace(/^\+/, "").replace(/^(-?)0+(?=\d)/, "$1");
+    if (str === "-0")
```
alexander-fenster, line 96: the previous regex will still allow `-00000` if I'm reading correctly?

alexander-fenster (approved):

ext/protojson.js
dcodeIO: Generally, leading zeros remain accepted but are trimmed before parsing. Here specifically, the `0+` match and digit lookahead remove all leading zeros while always leaving one digit, so standalone `0` and `-0` are unchanged by the regex.

**Edit:** Relatedly, checking the ProtoJSON spec, I do not see language requiring quoted integer strings with leading zeros to be rejected, nor is there a conformance test for that case. The upstream C++ parser passes quoted integers to `absl::SimpleAtoi`, which accepts leading zeros, so this errs on the side of doing the same.

alexander-fenster: Yeah my bad, I did not realize that the `\d` in the lookahead might as well be a zero too. Np, this is all good!

PR #2426, opened 2026-09-02
fix: Parse character, octal, and hex string escapes

jiang1997:
Support additional protobuf character escapes, including escaped quotes, plus bounded octal and hexadecimal escapes. Add tests for parsed defaults and numeric escape boundaries.

Unicode escapes and validation of malformed or unknown escapes remain unsupported.

### DIFF
--- a/src/tokenize.js
+++ b/src/tokenize.js
@@ -9,13 +9,19 @@ var setCommentRe = /^ *[*/]+ */,
     setCommentAltRe = /^\s*\*?\/*/,
     setCommentSplitRe = /\n/g,
     whitespaceRe = /\s/,
-    unescapeRe = /\\(.?)/g;
+    unescapeRe = /\\([0-7]{1,3}|[xX][0-9a-fA-F]{1,2}|.?)/g;
 
 var unescapeMap = {
-    "0": "\0",
+    "a": "\x07",
+    "b": "\b",
+    "f": "\f",
     "r": "\r",
     "n": "\n",
-    "t": "\t"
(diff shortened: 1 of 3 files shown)


alexander-fenster:
@jiang1997 Do you know what's the behavior of the `protoc` parser for these cases? I can probably go figure out but if you have a pointer to their code or docs that would've been handy :)

jiang1997:
@alexander-fenster Yes. I used the protobuf string literal grammar and protoc's tokenizer as references:

- Spec: https://protobuf.dev/reference/protobuf/proto3-spec/#string-literals
- Implementation: https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/io/tokenizer.cc#L1163-L1235

protoc consumes up to three octal digits and two hexadecimal digits. I also verified that the parsed defaults match protoc's output.

jiang1997:
Separately, Unicode escapes (`\uXXXX` and `\UXXXXXXXX`) are still missing. I think supporting them would be cleaner with a small character-by-character scanner instead of extending the current regex.

Would that make sense for a follow-up PR?

alexander-fenster (approved):

alexander-fenster:
Hi @jiang1997, thank you! I approved and merged this PR; yes, the proper Unicode support should probably go in a separate PR.

I assume the current code won't parse non-ASCII codes like "\xff" correctly because it will go through `String.fromCharCode(255)` resulting in some weird character rather than the intended byte 255, we can probably handle this separately.

PR #2434, opened 2026-09-13
fix(cli): pbjs fails with a clear error when a glob matches no files

### LINKED ISSUE #2432: pbjs: glob matching zero files silently generates empty output with exit 0
`pbjs -t json 'typo*.proto'` writes an empty `{}` bundle and exits 0 when the glob matches nothing, so a typo'd glob in a build script produces an empty codegen artifact with a green build.

### Repro

```
$ pbjs -t json 'nonexistent*.proto' > out.json; echo $?
0                # out.json is "{}"
```

### Cause

After glob expansion in `cli/pbjs.js` (`glob.sync` → splice), a zero-match pattern leaves `files = []`; `root.loadSync([])` succeeds and the target emits an empty bundle. There is no post-expansion check.

### Proposed fix

Track whether any glob pattern expanded to zero matches and fail with a clear stderr message (`no files matched pattern '…'`) and exit 1, consistent with the existing usage-error path.

<details><summary>AI transparency</summary>
Found during an agent-run improvement audit of a downstream consumer's dependency: exploration/verification by GPT-5.6 Luna and Grok 4.6 (OpenCode CLI); planning and implementation by GLM (ZCode). Repro executed locally against master (0ad9a28a).
</details>

JavaGT:
Fixes #2432.

## What changed

`pbjs` now errors with exit 1 and a stderr message when a glob pattern expands to zero files, instead of silently generating an empty bundle:

```
$ pbjs -t json 'typo*.proto'
pbjs: no files matched pattern 'typo*.proto'    # exit 1 (previously: wrote "{}", exit 0)
```

## Why

A typo'd glob in a build script produced an empty codegen artifact with a green build — the failure surfaced only later, when consumers of the generated file misbehaved. Glob expansion in `main()` replaced the pattern with its (possibly empty) match list and nothing re-checked it.

## Verification

- Zero-match pattern → stderr message, exit 1, empty stdout.
- Working glob (`dir/*.proto`) and plain file arguments behave exactly as before (output byte-identical, exit 0).
- `npm test` passes on this branch (the CLI package itself has no test suite; the cases above were run manually against `cli/bin/pbjs`).

## AI transparency

Found during an agent-run improvement audit of a downstream consumer's dependency. Exploration/verification: GPT-5.6 Luna and Grok 4.6 via OpenCode CLI (both independently confirmed the behavior). Planning and implementation: GLM (ZCode agent). Verified against master at 0ad9a28a.


JavaGT:
Post-implementation review caught two refinements, now pushed: programmatic `pbjs.main(args, callback)` callers now receive the error through the callback (previously only the CLI path was handled, which would have hung a callback-only consumer), and the message no longer carries a `pbjs: ` prefix, matching how the rest of this file reports errors.

alexander-fenster:
It seems like the decoder fix does not belong to this PR?

